PR #15, opened 2020-06-25
Meta tags, misc bugfixes and changes

itinerare:
New:
- Added configurable meta tags-- used for link previews on social media, in Discord, etc.
- These display the page title of the linked page (or the login page if it requires auth), a message (configurable in config/lorekeeper/settings.php), and an image (a copy of the Lorekeeper "Inventory" image by default, adjustable via the Site Images admin panel)
- When a character is linked, rather than displaying the site's meta tag image, it will display the thumbnail image for the character
- Add sorting options for the user list (alphabetical and reverse alphabetical for both names and aliases respectively, rank (same as default behavior), and oldest/newest first). Default behavior is unchanged. Also adds sorting for admin user list.
- Add toggle to news post edit form to "bump" post (alert users) if desired
- Add an "ask first" option to the gift art option on character profiles

Bugs fixed:
- Editing character description or notes and then trying to open a modal results in an interminable dimmed page, no modal (thanks again to [Speedy](https://github.com/SpeedyD)!)
- Subtypes assigned to MYO slots aren't properly applied to characters on design approval
- Approved MYO submissions go to approved design submissions panel rather than approved MYO submissions. Not retroactive. (This is done by adding a new column for design update type.) However, design updates at varying parts of the submission process should have their type set correctly, as it's set at submission and approval (in addition to at creation) as an alternate form of backwards-compatibility. Note that existing submitted MYOs will be placed in the design update queue as a consequence of this, but, again, the type should be set properly to MYO when they are processed
- Editing traits on an image with more than one copy of a single trait consumes duplicate copies (thanks to [Draginraptor](https://github.com/Draginraptor) for discovering the fix itself!)
- Attempting to approve a design update when the image would go into a new directory errors
- Rejecting a design update sends a notification to the processing user rather than the user who opened the request
- Individual raffles roll duplicate winners
- Hidden character images invisible to everyone including permitted staff
- MYO slot creation, design approvals show in approving staff member's ownership log
- Reuploading a character image with a set/non-cropped thumbnail throws an "invalid path" error

Misc:
- Tweak the mobile CSS of the timestamp box on the dashboard slightly
- Removed extra spaces from page titles
- Disable HTMLPurifier cache. This was previously on a branch when the definitions were updated, with the expectation that given time it would account for the changes; this does not seem to have been the case, so pending a tidier solution, this seems the most appropriate course of action.
- Add subtype display to design update requests. Displays subtype on character image for MYOs, if the image has one(/the subtype is pre-set)
- Set shop image max width (for the sake of mobile/smaller screens)
- Order news posts in admin and user-facing lists by time of last update
- Adjust design update log to use set design update type (as discussed above) when being created
- Set raffle group visibility setting to cascade to any raffles in the group on editing the group (setting the group to visible will cause all raffles within it to be visible and vice versa). Individual raffles can still be toggled at will

Requires migration!


Draginraptor (changes requested):
- app/Helpers/Helpers.php: apparently there's a TODO to remove
- UserController/BrowseController: for the sorting of users, is it necessary to check for isset? Maybe use the default option on switch?
- Using Auth within the CharacterImage model feels off - shouldn't users be authenticated in the Controller instead? (Following the principle of having different 'jobs' for these file types)
- Should ExtensionService be there?

itinerare:
> * app/Helpers/Helpers.php: apparently there's a TODO to remove
It's ostensibly a to-do but HTMLPurifier's caching is a mystery so it remains a to-do (that is, it wasn't expiring). I folded it into main since the fact that it _wasn't_ expiring was causing issues.
> * UserController/BrowseController: for the sorting of users, is it necessary to check for isset? Maybe use the default option on switch?
Hmm, yeah, I'll look into that
> * Using Auth within the CharacterImage model feels off - shouldn't users be authenticated in the Controller instead?
> (Following the principle of having different 'jobs' for these file types)
I've been meaning to move that sort of thing.... but haven't gotten around to it to this point |D less busy now though so
> * Should ExtensionService be there?
It was intended to be there! But it may be more prudent to move it to dedicated [modified] version since plans have chanted

itinerare:
UserController/BrowseController: turns out the isset Is necessary, but it's tidier now nonetheless
Moved Auth out of the CharacterImage model and into controller(s) (Browse, Character)

Draginraptor (approved):
Alright, looks good now!

preimpression (approved):

PR #19, opened 2020-07-26
Add newest/oldest sorting for submissions

Draginraptor:
Adds the ability to sort submissions by oldest or newest
Oldest is the default

Also fixes a bug where searching all categories would result in no prompts showing


preimpression (approved):
Looks good

PR #30, opened 2020-10-19
Adjustments to Admin Panel + Bugfix

preimpression:
Moved several category links out of the sidebar and into their target indexes (aka item categories is now a link in items) to make the sidebar a little less Tall, as well as reorganized some of the Data sidebar links by purpose (shops and currencies together, prompts and loot tables together, etc)


itinerare (approved):

config/lorekeeper/admin_sidebar.php
```diff
@@ -75,9 +75,42 @@
...
+                'name' => 'MYO Approvals',
+                'url' => 'admin/myo-approvals/pending'
+            ],
+        ]
+    ],
     'Data' => [
         'power' => 'edit_data',
         'links' => [
+            [
+                'name' => 'Char. Categories',
```
Draginraptor, line 111: Not really a fan of this being the sole abbreviation of character in the sidebar

Draginraptor (commented):
Just thinking about 'Char. Categories' being the only instance of 'Character' that's abbreviated xD

config/lorekeeper/admin_sidebar.php
preimpression: It was the longest thing on the sidebar after removing Categories, it just makes the sidebar so wide haha. I can remove it if the consensus is back to normal though

Draginraptor: Personally in favour of maintaining consistency with "Character", but maybe @itinerare has an opinion? :p

itinerare: I don't personally see an issue w it but I also understand the feeling toward consistency.... I don't have a strong opinion either way though |D

preimpression: Made this change @Draginraptor !

itinerare (approved):

Draginraptor (approved):
Apologies for the late reply! But yes, my desire for consistency has be satisfied :p

PR #32, opened 2020-10-24
Fixing error 'Undefined variable: $stock' 

Ne-wt:
defines $stock before it is called instead of in the return view area.


Draginraptor (approved):
Simple and solid

itinerare (approved):

PR #35, opened 2020-11-02
Add open/closed toggle, delayed comments, and bumping to sales posts

itinerare:
- Adds a toggle (open/closed) to sales posts; mostly just impacts display (the result is displayed in the post title)
- Adds a 'comments open at' time/date field; if this is set (and in the future), comments won't be displayed for the post until that point. (A message is instead displayed beneath it on the post's page itself saying that comments aren't open yet, and will open at (x ) time; uses the pretty_date helper).
- If a post is set to 'open' and there is a set future 'comments open at' time, the title will display 'preview' instead of 'open'
- Staff with permissions to edit sales posts are still able to see the comments interface, so that if any set-up needs to be done before a sale opens, it can be done without any need to scramble
- Add the ability to 'bump' extant sales posts

Requires migration : Tested locally and live

### DIFF
--- a/app/Http/Controllers/Admin/SalesController.php
+++ b/app/Http/Controllers/Admin/SalesController.php
@@ -64,7 +64,7 @@ class SalesController extends Controller
     {
         $id ? $request->validate(Sales::$updateRules) : $request->validate(Sales::$createRules);
         $data = $request->only([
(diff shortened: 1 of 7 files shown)


Ne-wt (approved):
Solid and wanted changes, definitely necessary long-term!

PR #36, opened 2020-11-05
Merge Reports into modified main

Ne-wt:
merge reports addon into modified main - will allow reportable compatibility with comments, galleries etc in future


itinerare (approved):
I've already gone through and tested + done a fixes pass, plus given everything another once-over here

Draginraptor (approved):
Looks good to me

PR #37, opened 2020-11-08
Switch art/design credits to use on-site username/ID instead of alias

itinerare:
- Add css to make images in parsed text not stretch the page, especially on mobile
- Switch artist and designer credits to using on-site usernames/IDs rather than aliases; user selection now a dropdown using selectize.
- Add a function akin to character ownership updating which, on verification or viewing of a user's profile, assigns any existing character image creators with the same alias as the user to the user's ID instead
- This function can also extract alias from dA URLs provided as credit to perform this operation
- Add a command to perform the same function for existing character image creators (`php artisan assign-art-credits-to-ids`) where possible (that is, where a user exists on the site with a given alias); aliases without associated accounts are converted to urls (assumes a dA account). The "alias" column can be safely dropped after this operation, if desired.
- Update masterlist/MYO masterlist search to use on-site usernames instead
- Add a helper function which formats user profile urls from various sites in the fashion of the old (username)@ dA links, and changed art/design credits to use it.
- Add a config file for sites including regex to facilitate the above/easy addition of sites

Tested locally - Requires migration

### DIFF
--- a/app/Http/Controllers/BrowseController.php
+++ b/app/Http/Controllers/BrowseController.php
@@ -133,29 +133,6 @@ class BrowseController extends Controller
         if($request->get('sale_value_max')) $query->where('sale_value', '<=', $request->get('sale_value_max'));
 
         if($request->get('is_trading')) $query->where('is_trading', 1);
-        if($request->get('is_gift_art_allowed')) switch($request->get('is_gift_art_allowed')) {
-            case 1:
-                $query->where('is_gift_art_allowed', 1);
-            break;
-            case 2:
-                $query->where('is_gift_art_allowed', 2);
-            break;
-            case 3:
-                $query->where('is_gift_art_allowed', 1)->orWhere('is_gift_art_allowed', 2);
-            break;
-        }
-        if($request->get('is_gift_writing_allowed')) switch($request->get('is_gift_writing_allowed')) {
-            case 1:
-                $query->where('is_gift_writing_allowed', 1);
-            break;
-            case 2:
-                $query->where('is_gift_writing_allowed', 2);
-            break;
-            case 3:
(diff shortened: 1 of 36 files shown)


app/Http/Controllers/BrowseController.php
```diff
@@ -283,16 +289,22 @@ public function getMyos(Request $request)
 
         // Searching on image properties
         if($request->get('species_id')) $imageQuery->where('species_id', $request->get('species_id'));
-        if($request->get('artists')) {
+        if($request->get('artist')) {
```
preimpression, line 292: Inconsistency in request for 'artist' vs 'artists'

app/Http/Controllers/BrowseController.php
```diff
@@ -283,16 +289,22 @@ public function getMyos(Request $request)
...
-                $query->where('alias', $artistName);
+            // Usernames are prevented from containing spaces, but this is to deal with previously made accounts with spaces in names
+            $artistName = str_replace('%20', ' ', $artistName);
+            $artists = User::where('name', 'LIKE', '%' . $artistName . '%')->pluck('id')->toArray();
+            $imageQuery->whereHas('artists', function($query) use ($artists) {
+                $query->whereIn('user_id', $artists);
             });
         }
-        if($request->get('designers')) {
+        if($request->get('designer')) {
```
preimpression, line 301: Inconsistency in request for 'designer' vs 'designers'

preimpression (commented):
Overall, looks good! Just the couple of weird inconsistencies we talked abt on Discord.

preimpression (approved):

Draginraptor (approved):
Sorry that it took a while, but I think the code looks good!

PR #44, opened 2020-11-17
Feature/auth rework

itinerare:
- Rework alias storage; move to new table
- Change verification checking and alias setting to fit new storage method
- Rework user alias checking/display for new storage method
- Add additional information to sites config for construction of links, etc
- Add auth and primary_alias toggles to sites config-- controlling whether it should be offered as an auth provider and whether aliases from the site can be a user's primary alias
- Build out a single alias info migration command; has an additional option, `--drop-columns`, which drops the alias columns when it is finished moving information either to user_id or url columns. It's recommended to run it without once and ensure data has been moved properly before running it again with the option.
- Update character ownership to use URL instead of alias
- Update character owner display to use prettyProfileLink function/handle a variety of sites
- Update character logs to use URL
- Add helper function to check if a URL is from a site used for auth, and if so, if it corresponds to an on-site user
- Add helper function to format 'name'@'link' string without link
- Tidy up art/design credit entry
- Remove search by alias from masterlist/MYO/sublist search
- Update item artist credits to new system/to use user ID
- Update user alias display in admin panel to support multiple aliases per user, specify site the alias is associated with
- Update clearing user aliases to support multiple aliases per user
- Fix MYO slots not being transferrable to alias/URL/offsite users #49


app/Helpers/Helpers.php
```diff
@@ -188,6 +188,35 @@ function randomString($characters)
...
+    foreach(Config::get('lorekeeper.sites') as $key=>$site) if(isset($site['auth']) && $site['auth']) {
+        preg_match_all($site['regex'], $url, $matches); 
+        if($matches != []) {$urlSite = $key; break;}
+    }
+    if($matches[0] == [] && $failOnError) throw new \Exception('This URL is from an invalid site. Please provide a URL for a user profile from a site used for authentication.');
+
+    // and 2. if it contains an alias associated with a user on-site.
+    if($matches[1] != [] && isset($matches[1][0])) {
+        $alias = App\Models\User\UserAlias::where('site', $urlSite)->where('alias', $matches[1][0])->first();
+        if($alias) $recipient = App\Models\User\User::find($alias->user_id);
```
corowne, line 213: You can use `$recipient = $alias->user;`, though I don't think it makes a major difference

corowne (changes requested):
Looks good, just a few things:
- Admin panel (editing a user) - in the clear alias section, the alias textbox still uses the old alias as its value so the box is blank
- Admin\Users\UserController.php - related to the above; postUserAlias (to clear the old alias) needs to be updated to use the new alias table too
- BrowseController.php - masterlist character/MYO/sublist search by username still searches on owner_alias, user search by alias still searches on alias column on users

(Also not sure if within scope of this update - items still use artist_alias)

app/Helpers/Helpers.php
itinerare: Ahh yes
I don't think it does make much of a diff, but it feels tidier, so I like it, haha

itinerare:
All requested changes made-- I updated item artists to the new system as well. Going to take a look at addressing #49 before requesting review again though.

app/Http/Controllers/Admin/Users/UserController.php
```diff
@@ -110,19 +111,23 @@ public function postUserBasicInfo(Request $request, $name)
...
         }
+        if(!$alias) {
+            flash('Invalid alias.')->error();
+        }
         else if (!Auth::user()->canEditRank($user->rank)) {
             flash('You cannot edit the information of a user that has a higher rank than yourself.')->error();
         }
-        else if($user->alias && $user->update(['alias' => null])) {
+        else if($alias && $alias->delete()) {
             UserUpdateLog::create(['staff_id' => Auth::user()->id, 'user_id' => $user->id, 'data' => json_encode($logData), 'type' => 'Clear Alias']);
```
corowne, line 131: I think this'll need a check to see if the alias is the primary alias/the only alias the user has and edit has_alias if the user has no other valid identification

corowne (changes requested):
Looks good I think, just one last thing!

app/Http/Controllers/Admin/Users/UserController.php
itinerare: Oh yes! Thanks, I'll add that

corowne (approved):
Good to go I think!

PR #45, opened 2020-11-18
Feature/Toggle prompt submission visibility

itinerare:
- Adds a 'hide prompt submissions [never/until prompt ends/forever]' option to prompt creation/editing
- Alters the viewable() submission scope to hide submissions (viewable neither in users' submission logs nor via link) associated with a prompt which either hides submissions forever or (if a prompt has a set future end + is set to hide submissions until end) hides them until the prompt concludes. Staff with the 'manage submissions' permission can, as before, always view all submissions.
- Adds a line to prompt entries in the 'more details' collapse which notes that submissions to the prompt are hidden/hidden until the prompt's end.

Tested locally - Requires migrate

### DIFF
--- a/app/Models/User/User.php
+++ b/app/Models/User/User.php
@@ -450,9 +450,9 @@ class User extends Authenticatable implements MustVerifyEmail
      *
      * @return \Illuminate\Pagination\LengthAwarePaginator
      */
-    public function getSubmissions()
+    public function getSubmissions($user = null)
     {
-        return Submission::with('user')->with('prompt')->where('status', 'Approved')->where('user_id', $this->id)->orderBy('id', 'DESC')->paginate(30);
+        return Submission::with('user')->with('prompt')->viewable($user ? $user : null)->where('user_id', $this->id)->orderBy('id', 'DESC')->paginate(30);
     }
 
     /**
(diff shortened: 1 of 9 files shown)


preimpression (approved):

app/Models/User/User.php
```diff
@@ -472,9 +472,9 @@ public function updateArtDesignCredits()
      *
      * @return \Illuminate\Pagination\LengthAwarePaginator
      */
-    public function getSubmissions()
+    public function getSubmissions($user = null)
```
Draginraptor, line 475: Is taking in $user necessary if getSubmissions() is already called on the user itself? Maybe $this could be used instead.

itinerare: Ahhh hmm true, I'll take a look at that

itinerare: AH I remembered why it's there
it's taking in the user _viewing_, rather than the user whose submissions are being viewed, so on the controller etc end it's doing an auth check and piping that in

Draginraptor: Ooh I see, that clears that up then 👌

Draginraptor (approved):

PR #53, opened 2020-11-28
Misc small bugfixes

itinerare:
- Fix design update index display showing vote counts erroneously
- Fix admin user index link not directing to user edit page
- Fix user profile inventory display showing items with a count of 0

Fixing misc small issues as reported by folks/the bug report channel. No specific actions needed.

### DIFF
--- a/resources/views/admin/designs/index.blade.php
+++ b/resources/views/admin/designs/index.blade.php
@@ -27,26 +27,32 @@
       <div class="col-md-3 font-weight-bold">{{ $isMyo ? 'MYO Slot' : 'Character' }}</div>
       <div class="col-3 col-md-3 font-weight-bold">User</div>
       <div class="col-2 col-md-2 font-weight-bold">Submitted</div>
-      <div class="col-2 col-md-2 font-weight-bold">Votes</div>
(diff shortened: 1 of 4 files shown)


resources/views/admin/designs/index.blade.php
```diff
@@ -27,26 +27,32 @@
...
-            if($vote == 2) $approveSum += 1;
-        }
-    ?>
+    @if(Config::get('lorekeeper.extensions.design_update_voting'))
+        <?php
+            $rejectSum = 0;
+            $approveSum = 0;
+            foreach($r->voteData as $voter=>$vote) {
+                if($vote == 1) $rejectSum += 1;
+                if($vote == 2) $approveSum += 1;
```
Draginraptor, line 42: You could use increment (++$var or $var++) for these :>

Draginraptor (approved):
Did comment bout the increment thing but I still think it's good for merging

resources/views/admin/designs/index.blade.php
itinerare: Could, but given it works and it's just fixing the fact that it was being displayed when it shouldn't have been, I'm (shrugs) about it
Comment still appreciated though!

PR #61, opened 2020-12-19
Update Laravel version to 8.0

corowne:
**Laravel version update**
- Packages updated
- Exceptions in Handler.php changed to Throwable
- String and array helpers updated to use Str and Arr in laravel/helpers
- Email verification resending link updated to use post method
- Changes to AppServiceProvider to continue using Bootstrap

**Bug fixes**
- Gallery currency queue page errors if no currencies exist
- Missing code in CharacterManager.php


composer.json
```diff
@@ -8,32 +8,34 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.1.3",
+        "php": "^7.3",
         "boxconnect/deviant-php": "dev-master",
         "doctrine/dbal": "2.10.4",
```
itinerare, line 13: This can I think safely be allowed to upgrade at this point, given that the issue was newer versions requiring php 7.2

itinerare (approved):
Looks good-- only thing is possibly allowing doctrine/dbal to update, but I leave that at your discretion given the particulars of getting packages to cooperate with each other

preimpression:
Will the new Throwable need to be used for every instance of Exception currently used, across Extensions etc?

Issue #38, opened 2020-11-09
Users with long usernames cause the sidebar to partially disappear


SpeedyD:
### Bug as reported:

**URL of page:** All user pages
**Description of bug:** Sidebar on the left gets hidden behind site content
**How was the bug caused (exact steps to reproduce):**
Make your browser thinner on the userpage of a user with a long name.

### Further details

Users with long names causes the links in the sidebar on the left to get hidden behind site content, if the browser isn't wide enough. Considering not everyone uses full HD resolutions on their screen, their browser might not ever get wide enough.

It's not a major bug as the mobile size picks up the pace in general, but it's still a problem that happens in core Lorekeeper.

SpeedyD:
Realization was made that this also occurs on _character pages_ when the _character_ has a long name.

SpeedyD:
Minor suggestion for fix?

https://stackoverflow.com/questions/15012712/truncate-string-in-laravel-blade-templates/24902688#24902688

Truncating names. It's not important to have the full name of characters or users in the sidebar anyway, especially since they're displayed on the pages themselves, too.

itinerare:
Ah, yeah, it'd be pretty easy to truncate especially long names/that'd be a decent solution... probably just needs some fiddling as to precisely how long it's allowed to be before it truncates, esp w regards to figuring out something that's a generally applicable value.

SpeedyD:
On core Lorekeeper, the longest term on both character pages and userpages is 'OWNERSHIP HISTORY'. That's 17 characters.
With a long name, when the browser is smallest, it seems about 9-11 characters are visible..

Considering that the menu items get new-lined if they're squished, the issue is solely the breaking of the column when the name is long..

If there is a way to see browser width in Laravel, I'd suggest either variable widths, but at the purest, most basic form, I'd suggest using around 10 as a maximum.

itinerare:
I was actually looking into determining viewport width for something else.... if I recall correctly, that is a Hassle, so 10 just a hard cap is probably the idea.

SpeedyD:
For characters, (`resources\views\character\_sidebar.blade.php`) I think it's easier to just replace the `{{ $character->fullName }}` on line 2 with `{{ $character->slug }}`. The slug scrunches up automatically because of the hyphens, so it really just depends on the given name whether it might mess things up or not. (Not showing the name doesn't matter for the sidebar anyway..)

[...]

PR #68, opened 2021-01-19
Develop Issues fix PR

Ne-wt:
Fixed majority of issues (did not fix / attempt to fix issue #38)

All fixes where tested locally, logged in and out for those that could be affected by auth

### DIFF
--- a/app/Http/Controllers/Users/UserController.php
+++ b/app/Http/Controllers/Users/UserController.php
@@ -116,7 +116,7 @@ class UserController extends Controller
      */
     public function getUserSublist($name, $key)
     {
-        $query = Character::myo(0)->visible()->where('user_id', $this->user->id);
+        $query = Character::myo(0)->where('user_id', $this->user->id);
         $imageQuery = CharacterImage::images(Auth::check() ? Auth::user() : null)->with('features')->with('rarity')->with('species')->with('features');
 
         $sublist = Sublist::where('key', $key)->first();
@@ -128,6 +128,8 @@ class UserController extends Controller
         if($subSpecies) $imageQuery->whereIn('species_id', $subSpecies);
 
         $query->whereIn('id', $imageQuery->pluck('character_id'));
+
+        if(!Auth::check() || !(Auth::check() && Auth::user()->hasPower('manage_characters'))) $query->visible();
 
         return view('user.sublist', [
             'user' => $this->user,
@@ -145,9 +147,12 @@ class UserController extends Controller
      */
     public function getUserMyoSlots($name)
     {
+        $myo = $this->user->myoSlots()->get();
+        if(!Auth::check() || !(Auth::check() && Auth::user()->hasPower('manage_characters'))) $myo->visible();
+
         return view('user.myo_slots', [
             'user' => $this->user,
-            'myos' => $this->user->myoSlots()->visible()->get(),
+            'myos' => $myo,
             'sublists' => Sublist::orderBy('sort', 'DESC')->get()
         ]);
     }
--- a/app/Http/Controllers/Users/UserController.php
+++ b/app/Http/Controllers/Users/UserController.php
@@ -78,7 +78,7 @@ class UserController extends Controller
      */
     public function getUserCharacters($name)
     {
-        $query = Character::myo(0)->visible()->where('user_id', $this->user->id);
+        $query = Character::myo(0)->where('user_id', $this->user->id);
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Users/UserController.php
```diff
@@ -94,10 +94,12 @@ public function getUserCharacters($name)
         $imageQuery->whereNotIn('species_id', $subSpecies);
 
         $query->whereIn('id', $imageQuery->pluck('character_id'));
+        $admin_characters = $query->orderBy('sort', 'DESC')->get();
```
preimpression, line 98: Might I suggest instead of an entire new variable, doing something like `if(!Auth::check() || !(Auth::check() && Auth::user()->hasPower('manage_characters')) $query->visible();` ?

composer.json
```diff
@@ -10,7 +10,7 @@
     "require": {
         "php": "^7.4",
         "boxconnect/deviant-php": "dev-master",
-        "doctrine/dbal": "^3.0",
+        "doctrine/dbal": "2.2",
```
preimpression, line 13: October 15 commit by Merc suggests `"doctrine/dbal": "2.10.4",` and I believe it worked well on that for awhile.

resources/views/user/characters.blade.php
```diff
@@ -9,22 +9,40 @@
...
-@if($characters->count())
-    <div class="row">
-        @foreach($characters as $character)
-            <div class="col-md-3 col-6 text-center mb-2">
-                <div>
-                    <a href="{{ $character->url }}"><img src="{{ $character->image->thumbnailUrl }}" class="img-thumbnail" /></a>
-                </div>
-                <div class="mt-1 h5">
-                    {!! $character->displayName !!}
+@if(Auth::check() && Auth::user()->hasPower('manage_characters'))
```
preimpression, line 12: If you take my suggestion, don't forget to revert this back or let it work with the singular $characters variable!

resources/views/user/profile.blade.php
```diff
@@ -86,20 +86,37 @@
         @endforeach
     @endif
```
preimpression, line 87: The $user->character() etc code might work as a Controller-set variable with the addition of my query suggestion from above, to reduce redundancy?

app/Services/ShopService.php
```diff
@@ -111,7 +111,7 @@ public function updateShopStock($shop, $data, $user)
             if(isset($data['item_id'])) {
                 foreach($data['item_id'] as $key => $itemId)
                 {
-                    if(!$data['cost'][$key]) throw new \Exception("One or more of the items is missing a cost.");
+                    if($data['cost'][$key] < 0) throw new \Exception("One or more of the items is missing a cost.");
```
preimpression, line 114: Tested, throws a SQL error. Try: ` == null` instead of `< 0`

preimpression (changes requested):
A handful of suggestions. I only tested one thing, but I've seen the query->visible() bit used previously in something I believe Mercury did

preimpression:
Looking good so far! I'd suggest also handling Sublists and MYOs the same way!
Also, the masterlist might need the same functionality added

composer.json
itinerare: This was due to v2.11+ of the package requiring php 7.3 when at the time 7.2 was recommended for LK
As for the specific issue with migrations... what's been happening? I've not been having issues with it and I've been doing a ton of stuff with migrations w/ the existing version settings

itinerare (commented):
Most of this looks good! Mostly curious about doctrine/dbal and if you plan to follow Uri's suggestion re ML/sublists/MYOs?

composer.json
preimpression: This was reported in the discord and reverting fixed the issue:
[Discord Link](https://discord.com/channels/678909186683437056/678909804940623893/794637552320774185)

I linked this stackoverflow related to their error: https://stackoverflow.com/questions/33817983/artisan-migration-error-class-doctrine-dbal-driver-pdomysql-driver-not-fo

I also recall having to do the same since I got the same error they were facing at one point and a revert fixed it.

Draginraptor: To add on, I usually noticed it erroring when '->change()' is used (for editing existing columns)

itinerare: Ahh, ok, that makes sense-- I only half-remembered the issue as reported, so ty! I also have not in fact been editing existing columns in that way so that tracks

Ne-wt:
added to myos/sublists and also added a missing use Settings to gallery collaborator model that caused a 500 error when trying to comment

itinerare (approved):

Draginraptor (approved):

PR #71, opened 2021-02-06
adding missing $asset['quantity'] in the assethelper

Ne-wt:
adding $asset['quantity'] to item assethelper line 263


itinerare:
I checked and I've got a site running w/ this absent and no issues stemming from that? So I'm not sure what to make of it. What issues have you been encountering?

Ne-wt:
neither have I really, but seems odd to leave it out when every other service in the switch does have it :0

itinerare:
Unless it's been explicitly linked to an issue, and tested + found to be a solution to said issue, I don't think it's practical to make this kind of change, esp solely on the basis of what its neighbors are up to...

Ne-wt:
very fair!

PR #73, opened 2021-02-10
Character Items Bugfix

preimpression:
- Consolidates  character limit into `populateCategoryData` as it was duplicated in two different functions that use `populateCategoryData`
- Fixes inability to disable category character ownership (use case: accidental enabling)
--- Thoughts: Should this handle currently character-owned items?

Tbh I am not sure on the usage of the tertiary and the standalone variable? But it was how it was before so I left it and used the same method for character ownership.

### DIFF
--- a/app/Services/ItemService.php
+++ b/app/Services/ItemService.php
@@ -41,8 +41,6 @@ class ItemService extends Service
 
             $data = $this->populateCategoryData($data);
 
-            isset($data['character_limit']) && $data['character_limit'] ? $data['character_limit'] : $data['character_limit'] = 0;
-
             $image = null;
             if(isset($data['image']) && $data['image']) {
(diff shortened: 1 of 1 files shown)


itinerare (approved):

itinerare:
Thinking on handling currently character-owned items.... I can see a situation potentially happening where things go real weird and it being automatic is practical. But on the other hand I can also see an accidental toggle-off going _really_ sideways with that sort of automation in place.

preimpression:
Perhaps a standalone button + function or even a console command to remove + pass the items back to their players? One that gives plenty of warning like "This many characters have an item in this category. Are you sure? Are you really sure?"

itinerare:
Yeah I think if anything a standalone solution would be wisest

PR #74, opened 2021-02-14
Bug cleanup

### LINKED ISSUE #38: Users with long usernames cause the sidebar to partially disappear

### LINKED ISSUE #72: Sublists sorting different by default
### Bug as reported:

**URL of page:** `https://www.worldoflingua.com/sublist/<sublistkey>`
**Description of bug:** Sublists are, by default, sorted by Newest First, rather than Number Descending.
**How was the bug caused (exact steps to reproduce):**
Create a Sub Masterlist. Then use the new link in the sidebar once on the masterlist,
The display is, for some reason, sorted by Newest First by default. While the normal default is Number Descending.

### Further details

Considering the masterlist default is Number Descending, it is a bit strange that the Sublists are Newest First by default. It is quite confusing, especially if you're used to a different default.
I have no idea where in the code this happens, as far as I can personally tell, it's all on the same system, so it should be Number Descending.

itinerare:
Misc:
- Add a toggle to gallery editing to show the prompt selection box during gallery submission creation

Bugs Fixed:
- Message on prompt entry regarding submissions being hidden temporarily/indefinitely does not display properly
- More graceful handling for several errors when the currency used for an item's resale info is deleted
- Missing break in InventoryController
- Comment replies do not inherit the the parent comment's type (if it has one); add command to fix existing comments
- Sublists do not sort by number descending, consistent with masterlist, closes #72
- Users and characters with long usernames cause sidebar to partially disappear, closes #38 (Long usernames are truncated, characters have been switched to slugs only-- credit to Speedy for the fix as discussed in that issue)
- Fix empty rarity display when a trait has none
- Character sale value cannot contain decimals (credit to Speedy again)

Requires `php artisan migrate`, `php artisan fix-child-comment-types`


preimpression (approved):
Seems good to me

Draginraptor (approved):

PR #95, opened 2021-04-24
Add the ability to attach characters to sales posts

itinerare:
As on the tin. Opted to integrate it directly since if not used it has no more impact than the addition of the "add character" button when creating/editing sales posts.

- Type of sale can be specified per character, with relevant fields displayed for each (price for flatsale, SB/MI/AB for auction, and so on.)
- Short notes and/or a link to a place to claim/bid on/etc. the character can also be added.
- Once a character is attached, they can also be set open/closed semi-independently of the sales post (if the sales post is closed, all attached characters within are displayed closed as well).
- Sales posts with attached characters are divided into separate cards (header, followed by character(s), followed by the standard body of the sales post)-- each attached character is displayed in its own card, with sale type and info alongside info about the character, such as species and rarity. The character's first available/visible image is used in the event that a character receives design update(s) after being sold, for the sake of archival. 
- If there is only one character associated with a sales post, the full image rather than thumbnail is displayed as well as the character's traits.
- Very small but moves sales model to its own directory/namespace given the addition of an accompanying model; I have done my best to be thorough but worth mentioning just in case.

Includes a setting at the end of config/lorekeeper/settings to set preferred currency symbol for prices (default is $).

Tested locally/live, requires migrate, update-sales-comments. As usual esp for new features I'll not turn away more testing, of course.


database/migrations/2021_04_23_205230_add_link_to_sale_characters.php
```diff
@@ -0,0 +1,34 @@
...
+class AddLinkToSaleCharacters extends Migration
+{
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('sales_characters', function (Blueprint $table) {
```
preimpression, line 16: Very short migration - consider merging into the previous one for the PR?

database/migrations/2021_04_23_210656_add_is_open_to_sales_characters.php
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Run the migrations.
+     *
+     * @return void
+     */
+    public function up()
+    {
+        Schema::table('sales_characters', function (Blueprint $table) {
+            //
+            $table->boolean('is_open')->default(1);
```
preimpression, line 18: Ditto my previous comment - merge into other migration?

preimpression (changes requested):
Approved except the migration merging might be good for saving space + whatnot since they are really only an additional couple of lines!

itinerare:
Fair enough! I can compress em

preimpression (approved):

PR #98, opened 2021-05-03
Edits to new Birthday functionality + User settings forms

preimpression:
- Add admin side logging for birth date changes
- Change birthday column to a Carbon instance
- Fix bug where selectMonth would not show the actual birth month
- Change columns of profile to be two columns instead of birthday all on its own - and adding Clear style to divs after it so they don't get harassed by the avatar's float.
- Encase all user account forms in cards, user and admin side. This is because with more and more user forms in the settings, it is becoming very easy on non-customized LK to differentiate between the different forms.


preimpression:
Git believes the forms look a lot different than the actual edits, which were indentation and encasing in a div - that's pretty much it.

itinerare (approved):

PR #100, opened 2021-05-03
Feature/expand account authentication

corowne:
Authentication of various other social media sites
- Ability to choose the initial site linked from the primary ones available in the sites config
- Aliases management in user settings
- Change primary alias, hide non-primary aliases, remove non-primary aliases
- By default it's possible to link multiple accounts on the same site (but not the same account multiple times)
- List of aliases linked in sidebar on user profile page - non-paginated for now as it seems unlikely to become a very long page

Misc
- Changed fullName in sites config to full_name
- Slightly adjusted user admin panel page visually to handle page length better

.env must be updated to contain:

```
DEVIANTART_CLIENT_ID=
DEVIANTART_CLIENT_SECRET=

TWITTER_CLIENT_ID=
TWITTER_CLIENT_SECRET=

INSTAGRAM_CLIENT_ID=
INSTAGRAM_CLIENT_SECRET=

TUMBLR_CLIENT_ID=
TUMBLR_CLIENT_SECRET=

IMGUR_CLIENT_ID=
IMGUR_CLIENT_SECRET=

TWITCH_CLIENT_ID=
TWITCH_CLIENT_SECRET=
```

as required. It should be safe to leave out providers that will not be used for authentication.
Currently putting the redirect URIs into the services.php file - if this causes any issues (not seeing any on local testing), they can be added into .env as an additional {PROVIDER}_REDIRECT_URI field, like so:
```DEVIANTART_REDIRECT_URI=/auth/callback/deviantart```

### CODE REVIEW

**Review by itinerare — APPROVED** [review]

File: app/Models/User/UserAlias.php [hunk 1]
```diff
@@ -52,7 +69,7 @@ public function user()
     public function getUrlAttribute()
     {
         if($this->site == 'tumblr') return 'https://'.$this->alias.Config::get('lorekeeper.sites.tumblr.link');
-        else return 'https://'.Config::get('lorekeeper.sites.'.$this->site.'.link').'/'.$this->alias;
+        else return 'https://'.Config::get('lorekeeper.sites.'.$this->siteDisplayName.'.link').'/'.$this->alias;
```
itinerare, line 72: Needs to be changed back to $this->site I believe

itinerare (changes requested):
Adjusted review based on bug report-- looks good to me besides for this, however

Draginraptor (approved):
Besides the $this->site issue mentioned, the rest of the code looks fine

itinerare (approved):

itinerare (approved):

PR #104, opened 2021-05-05
fix $character->image refs into $character->character->image

JuniJwi:
Change "$character->image" references to "$character->character->image" references in order to use the correct object. The original method pulls the FIRST image linked to the character, while the second references the CURRENT/ACTIVE image of a character.

### DIFF
--- a/resources/views/sales/_character.blade.php
+++ b/resources/views/sales/_character.blade.php
@@ -2,14 +2,14 @@
     <div class="m-1">
         <div class="row">
             <div class="col-md-6 text-center align-self-center">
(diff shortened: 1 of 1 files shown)


itinerare:
Oh no this is... a feature, and not a bug! See #95

JuniJwi:
OH THATS HUGE BRAINED FHOAWIGH THANK YOU FOR CLARIFYING!

itinerare:
IT'S OK!! I'm happy to clarify! Esp given there have been many updates in the span of this week already but it's not yet time for the weekly roundup... alt. "I don't expect people to read every PR/that I haven't even told them about yet"

Issue #40, opened 2020-11-15
v2.0.0 - Milestone


corowne:
. | Legend
------------ | -------------
❓ | Unassigned
💬 | Needs discussion
👀 | Under review

**New features/Major updates**
- [ ] Upgrade to Laravel 8 - @corowne
- [ ] Update aliases to accommodate other social media sites - branch `feature/auth-rework` @itinerare
- [ ] Expand range of social media sites usable for aliases - @corowne

**Minor changes**
Pending

**Bug fixes**
- [ ]  Users with long usernames cause the sidebar to partially disappear #38
- [ ]  Attempting to edit stock of a store with "free" (cost 0) items errors #26
- [ ]  Users with especially long usernames cannot purchase any/some item(s) from shops #25
- [ ]  MYO submission open/close site setting has no effect, uses design update site setting instead #23

itinerare:
At least on DH, folk that need to upgrade php should give plenty time for that operation to complete before updating (and may want to do it in stages with plenty of time between)

PR #111, opened 2021-05-17
Clean up commands going toward v2

itinerare:
Trying to clean things up and make them more user-friendly going toward release.

- Create unified 2.0.0 update command, to be run after composer update (runs the list of commands currently living in #40-- migrate, clears caches, migrates aliases, etc. Just more efficient and hopefully easier on the user.)
- Update setup-admin-user to prompt, on local, user if they would like to mark their email verified or if they would like to set an alias (in lieu of doing these manually/via DB edits. Also for ease of use for folk not familiar with DB edits).
- Condense existing commands where viable (to curb command bloat)

Tested locally; update command is `php artisan update-lorekeeper-v2`. Can be safely run however many times.


app/Console/Commands/SetupAdminUser.php
```diff
@@ -80,9 +83,25 @@ public function handle()
             $this->line("Email: ".$email);
             $confirm = $this->confirm("Proceed to create account with this information?");
 
+            // If env variables indicate a local instance, double-check
```
preimpression, line 86: BLESS.

preimpression (approved):

Ne-wt (approved):

PR #127, opened 2021-06-15
Fix bug with galleries being (incorrectly) visible

itinerare:
Adds a visible() scope to galleries and correctly checks against it.
Tested live, no further actions required

### DIFF
--- a/resources/views/galleries/index.blade.php
+++ b/resources/views/galleries/index.blade.php
@@ -30,8 +30,8 @@
                         {{ $gallery->children->count() && (isset($gallery->start_at) || isset($gallery->end_at)) ? ' ・ ' : '' }}
                         @if($gallery->children->count())
                             Sub-galleries:
-                            @foreach($gallery->children as $count=>$child)
-                                {!! $child->displayName !!}{{ $count < $gallery->children->count() - 1 ? ', ' : '' }}
+                            @foreach($gallery->children()->visible()->get() as $count=>$child)
+                                {!! $child->displayName !!}{{ !$loop->final ? ', ' : '' }}
                             @endforeach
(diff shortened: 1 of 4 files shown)


resources/views/galleries/index.blade.php
```diff
@@ -30,8 +30,8 @@
                         {{ $gallery->children->count() && (isset($gallery->start_at) || isset($gallery->end_at)) ? ' ・ ' : '' }}
                         @if($gallery->children->count())
                             Sub-galleries:
-                            @foreach($gallery->children as $count=>$child)
-                                {!! $child->displayName !!}{{ $count < $gallery->children->count() - 1 ? ', ' : '' }}
+                            @foreach($gallery->children()->visible()->get() as $count=>$child)
+                                {!! $child->displayName !!}{{ !$loop->final ? ', ' : '' }}
```
Draginraptor, line 34: What's $loop here?

itinerare: Oh you really need to check out the loop variable ([Laravel docs on it](https://laravel.com/docs/8.x/blade#the-loop-variable)), it's exceptionally useful

Draginraptor: Ooh, I'll probably be using that the next time I write a loop. Thanks!
And I'll approve this PR as well.

Draginraptor (approved):

resources/views/galleries/index.blade.php
itinerare: Yeah Uri pointed it out to me a while back and I've been having a grand ol time since... here it also results in more consistent behavior as well so it's nice
In any case, thank you!

PR #132, opened 2021-07-10
Final -> Last

preimpression:
$loop->final should be $loop->last or it errors


itinerare (approved):
Confirmed that ->last is correct/working live

PR #134, opened 2021-07-17
fix: will now log to the correct user

preimpression:
fix: submission approval will now log to the correct user instead of the approving admin when including currency from user's bank

This issue likely arose because $user was only the submitter for a short while before it was switched back into $staff within the above block.

patch 2 includes removing the duplicate birthday name for the web route which was previously blocking `php artisan route:cache` from running.

### DIFF
--- a/app/Services/SubmissionManager.php
+++ b/app/Services/SubmissionManager.php
@@ -376,7 +376,7 @@ class SubmissionManager extends Service
             {
                 foreach($addonData['currencies'] as $currencyId=>$quantity) {
                     $currency = Currency::find($currencyId);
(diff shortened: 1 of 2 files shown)


itinerare (approved):
I've had success with $submission->user_id but this should work as well, so

preimpression:
You're right - that's actually better in the long run since it's once less table to be hopping through.

itinerare:
👍

itinerare (approved):

Ne-wt:
seems ok, the $submission->user_id will prevent anything accidental even if the $user works fine (better safe than sorry)

Ne-wt (approved):

PR #179, opened 2021-09-26
feat(tests): set up automatic testing via github actions

itinerare:
- includes moving header before HTML output to address an error

Runs very basic tests/sees if the app boots on a push or PR to any branch, as well as on publishing of releases; we can build out more/more structured tests down the line if we want, but for the moment it might be enough just to have it ping to make sure basic functionality is there/it can attempt to load... It may be necessary to add some minor checks to views to accommodate this but we can deal with that as needed/appropriate.

From my understanding/testing this works best when implemented across relevant branches? So PRing into master.
I've been working on this for a diff project, so a comparable setup can be viewed in action [there](https://github.com/itinerare/Mundialis/actions/workflows/laravel.yml).


Draginraptor (dismissed):

itinerare:
This is mostly just clogging up the PR list at this point; closing.

PR #183, opened 2021-09-29
Add flex-wrap to pagination to keep it from breaking mobile view.

JuniJwi:
Currently, if pagination gets too long it breaks the width of the mobile view. Adding flex-wrap to the pagination property in css fixes that!


itinerare:
This has been bugging me forever, thank you !!!

Ne-wt (approved):

itinerare (approved):

PR #192, opened 2021-10-08
refactor: move loot_select info fetching to views

itinerare:
- this is in the name of compatibility between different exts
- and to reduce support time demanded by issues with that

Going forward, ext authors can merely add support for a new entity to the `_loot_select` and `_loot _select_row` views... It's a "messy" fix but I think it's worthwhile.

Tested locally, no further action required.


Ne-wt (approved):
While it may not be "best practice", in common interest of maintainers and the community I feel this is a perfect solution and definitely will be a welcome one as the project expands.

PR #194, opened 2021-10-09
Fix for inventory select visible button

### LINKED ISSUE #185: Another inventory_select invalid quantity error
URL of page: New/edit trade, etc; places where inventory select occurs
Description of bug: Checkbox to select all visible items in inventory_select does not seem to be properly setting quantity, results in "invalid quantity" error. 
How was the bug caused (exact steps to reproduce): In this case,
- Attempt to create or edit a trade,
- Use checkbox in table header to select all visible items

I glanced at the JS in question and it looks like this should *work*? So I'm a little at a loss. Any ideas @draginraptor ?

Draginraptor:
Fixes #185 

I changed the way stack_quantity was accessed, but didn't update the js for the button to use this new method, resulting in the bug.

Tested locally.

### DIFF
--- a/resources/views/widgets/_inventory_select_js.blade.php
+++ b/resources/views/widgets/_inventory_select_js.blade.php
@@ -53,15 +53,14 @@
         }
         function selectVisible() {
             var $target = $('.user-item:not(.hide)');
-            $target.addClass('category-selected');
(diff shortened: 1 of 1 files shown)


itinerare (approved):
Makes sense! Thanks for taking a look at it.

PR #243, opened 2022-01-19
Implement PHP CS fixer

itinerare:
In essence, this combs over the project's PHP files and adjusts it to suit its configuration... I've adjusted the config to hit some compromise between standards/consistency and the existing style. I think the project could stand to benefit from this kind of thing to automatically keep an eye on/enforce consistent style, given that at this point there's a good few folk that have contributed (which is great!)

Needless to say, it shouldn't actually alter any functionality. While I'm including a github action to automatically apply fixes on push/PR (so that anyone contributing doesn't necessarily need to worry overmuch, it'll just be handled), it can also be run locally with `composer lint` or `composer sniff` for a dry run.


Draginraptor (approved):
Honestly, I ponder this sometimes, and absolutely love this

Ne-wt (approved):
much needed for an ever growing project and may even help newer members of the community to contribute 👍

itinerare:
> Honestly, I ponder this sometimes, and absolutely love this

Thank you! I'm actually quite pleased with it having played around with the config... I think it'll be of help going forward too since it means there's less need to worry about getting styling Just So/can focus on content instead. ... At least for the PHP. there are solutions for other things but I haven't found one I esp like for views

itinerare:
> much needed for an ever growing project and may even help newer members of the community to contribute +1

Would be ideal tbh, love to lower the barrier to that

PR #300, opened 2022-04-17
Add Toyhouse OAuth2 authentication

corowne:
It's here!! Big thank you to TH admin for getting OAuth2 up and working 🙏💖💕

Go to [https://toyhou.se/~developer](https://toyhou.se/~developer) to set up a client and grab your keys.

This setup was a bit more finicky than the others (code-wise), and there are a few things to note while using it:

- This can't be tested from localhost (there might be workarounds but it might still be annoying, I set up an instance on my server to test it)
- When creating a new app, the redirect URL has to be very specifically `http://your-site.com/auth/callback/toyhouse` as it doesn't seem to use wildcard matching. You may have to enter both non-www and www variants if you use them, etc.

Aside from these, no database/composer updates required. Please add to .env:
```
TOYHOUSE_CLIENT_ID=
TOYHOUSE_CLIENT_SECRET=
```

A couple of notes:
- I wrote a custom Socialite provider, but an official one MIGHT be added eventually so I'll change it then.
- Unlike the others, this one uses the Socialite driver stateless - received this recommendation for a stubborn error. It seems to work fine but I'm not entirely sure about the implications of this yet.
- Since TH is widely used within the adoptables community, and with dA sinking like the Titanic, I've opted to make this a primary alias type by default.

On an unrelated note I used the admin account setup to verify/enter an initial deviantART alias, but it still acted like I didn't have an alias attached (redirected me to the link page, and the new alias I authorised got set as a primary alias, so I ended up with 2 primary aliases) so there's probably a bug to look at there.

### DIFF
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -81,7 +81,9 @@ class HomeController extends Controller
             return redirect()->to(Auth::user()->has_alias ? 'account/aliases' : 'link');
         }
 
-        $result = Socialite::driver($provider)->stateless()->user();
+        // Toyhouse runs on Laravel Passport for OAuth2 and this has some issues with state exceptions,
+        // admin suggested the easy fix (to use stateless)
+        $result = $provider == 'toyhouse' ? Socialite::driver($provider)->stateless()->user() : Socialite::driver($provider)->user();
         if ($service->saveProvider($provider, $result, Auth::user())) {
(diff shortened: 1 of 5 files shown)


itinerare (approved):
Tested and works like a charm!
Noting here that I used all variations of redirect URL (w/ and w/o www, http vs https) in case it makes a difference....

Draginraptor (approved):

preimpression (approved):

PR #301, opened 2022-04-17
feat: add retroactive watermarking command

Ne-wt:


itinerare:
Hmm, it might be worth taking a look at the function in CharacterManager? iirc? that processes the image instead of just applying the watermark directly since there's a lot of considerations that get taken into account there

Ne-wt:
> Hmm, it might be worth taking a look at the function in CharacterManager? iirc? that processes the image instead of just applying the watermark directly since there's a lot of considerations that get taken into account there Also, thinking about thumbnails maybe

good point! added in next commit

itinerare (approved):

Draginraptor (approved):

preimpression (approved):

PR #309, opened 2022-05-02
Fixes Character Selection being unclear

AW0005:
I was playing around with making prompts, and ended up getting very confused when I kept getting the "Character does not exist" error on submission.

Turns out I kept thinking I had to click the "Add Character" button to get the character's details to populate in after entering the code, when I really should have just been clicking outside of the input.

For me that really doesn't feel intuitive, and so this is a tiny bug fix / feature? that changes the character selection input to be "onInput" rather than "onChange" so it updates as a user is typing. It does mean that you get an inline error about the character not existing until you've finished typing the code, but I find that to be more helpful.

I made this change to `_character_select_js.blade copy.php` to be thorough, although I'll be honest I don't know what the copy is used for.

screencast 2022-05-02 14-41-51

### DIFF
--- a/resources/views/admin/sales/_character_select_js.blade.php
+++ b/resources/views/admin/sales/_character_select_js.blade.php
@@ -18,7 +18,7 @@
         });
 
         function attachListeners(node) {
-            node.find('.character-code').on('change', function(e) {
+            node.find('.character-code').on('input', function(e) {
                 var $parent = $(this).parent().parent().parent().parent();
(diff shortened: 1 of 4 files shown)


itinerare:
A fair change! Might be worth updating the sales post and gallery submission versions of that setup as well...

AW0005:
@itinerare Good call! I just pushed an update to those two files as well. I think I'll leave the change I made in the copied file since it's there already, haha.

itinerare:
Sounds good, and fair enough!

Ne-wt (approved):

preimpression (approved):

Draginraptor (approved):

PR #310, opened 2022-05-02
Add the ability to generate the year in char codes

AW0005:
#### This adds the ability to use `{year}` in character code config to substitute in the current year.

This was small enough feature I figured I'd pitch it to core, but if you all would rather I set this up as an extension, I'd be happy to figure out doing that.

screencast 2022-05-02 16-25-11


itinerare:
Yeah, given it's small and readily configurable/entirely optional that's a fine addition as far as I'm concerned.

itinerare (approved):

Ne-wt (approved):

Draginraptor (approved):

PR #312, opened 2022-05-15
Update to Laravel 9/PHP 8.1

itinerare:
Updates to Laravel 9 and thus necessarily to require PHP 8. Since PHP 8.1 is the current version, I've just set it there for the moment. No particular impetus for this, but I figure it's easier to keep on top of these things as they come, and I was already updating my other projects/working out what needed adjusting, haha.

Also includes updating to the ignition error page package, which was adopted after Laravel 5 (hence being absent to this point) but has been standard for a while.

Tested locally; requires `composer install`.


Draginraptor (approved):

corowne:
Sorry for taking a while getting back on this, the required update to PHP 8 set off a chain of...unfortunate events on my local setup. Looks good I think!

PR #314, opened 2022-05-22
Bug:: Transfer log has errors if Character was Deleted

AW0005:
This fixes an issue where the Transfer Log will no longer load up and gives an error
"Trying to get property 'url' of non-object" if a character that had been in a transfer was later deleted (and therefore no longer exists to pull information from).

I hide the thumbnail entirely if the character doesn't exist and swap out the `h3` content with "Character Deleted".

### DIFF
--- a/resources/views/home/_transfer.blade.php
+++ b/resources/views/home/_transfer.blade.php
@@ -1,9 +1,13 @@
 <div class="transfer-row mb-2">
-    <div class="transfer-thumbnail"><a href="{{ $transfer->character->url }}"><img src="{{ $transfer->character->image->thumbnailUrl }}" class="img-thumbnail" alt="Thumbnail for {{ $transfer->character->fullName }}" /></a></div>
+    @if(isset($transfer->character)) <div class="transfer-thumbnail"><a href="{{ $transfer->character->url }}"><img src="{{ $transfer->character->image->thumbnailUrl }}" class="img-thumbnail" alt="Thumbnail for {{ $transfer->character->fullName }}" /></a></div> @endif
     <div class="transfer-info card ml-2">
         <div class="card-body">
             <div class="transfer-info-content">
-                <h3 class="mb-0 transfer-info-header"><a href="{{ $transfer->character->url }}">{{ $transfer->character->fullName }}</a></h3>
+                @if(isset($transfer->character))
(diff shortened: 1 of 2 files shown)


resources/views/home/_transfer.blade.php
```diff
@@ -1,9 +1,13 @@
 <div class="transfer-row mb-2">
-    <div class="transfer-thumbnail"><a href="{{ $transfer->character->url }}"><img src="{{ $transfer->character->image->thumbnailUrl }}" class="img-thumbnail" alt="Thumbnail for {{ $transfer->character->fullName }}" /></a></div>
+    @if(isset($transfer->character)) <div class="transfer-thumbnail"><a href="{{ $transfer->character->url }}"><img src="{{ $transfer->character->image->thumbnailUrl }}" class="img-thumbnail" alt="Thumbnail for {{ $transfer->character->fullName }}" /></a></div> @endif
```
itinerare, line 2: This should just be able to be `@if($transfer->character)` iirc

resources/views/home/_transfer.blade.php
```diff
@@ -1,9 +1,13 @@
...
     <div class="transfer-info card ml-2">
         <div class="card-body">
             <div class="transfer-info-content">
-                <h3 class="mb-0 transfer-info-header"><a href="{{ $transfer->character->url }}">{{ $transfer->character->fullName }}</a></h3>
+                @if(isset($transfer->character))
```
itinerare, line 6: Likewise here

itinerare (commented):
This broadly looks good! Might be worth tidying up just slightly, iirc the isset isn't strictly required...? since it should simply fail if there's no object at the other end even without it.

AW0005:
@itinerare 
You're right! I've had php complain at me in certain cases when I didn't use isset, so I've ended up opting to use it more often than not as a safeguard.

I cleaned it up, and made sure to cover the other `_transfer` template since there's two of them, and they could both run into the error as it turns out.

itinerare:
Oh yeah definitely you need it a good amount of times-- just not here!
Sounds good, definitely like to cover all bases.

itinerare (approved):

Sabriel-Koh (approved):

Draginraptor (approved):

PR #330, opened 2022-06-19
Bug:: Issue fix for ItemSlots not being usable

AW0005:
Fixes the following situation:

1. Non-Staff user has an item with the slot tag
2. Non-Staff user attempts to use item with the slot tag to get a slot
3. SlotService attempts to call `createCharacter` in the CharacterManager to create the slot
4. User passed to `createCharacter` is not staff, so when it tries to log it as an adminAction, it fails.

### DIFF
--- a/app/Services/CharacterManager.php
+++ b/app/Services/CharacterManager.php
@@ -170,7 +170,9 @@ class CharacterManager extends Service
                 ));
             }
 
-            $this->logAdminAction($user, 'Created Character', 'Created '.$character->displayName);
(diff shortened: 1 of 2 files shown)


itinerare:
Hmm, the check to see if logging succeeds or fails is still relevant so I'd rather it not be removed. I think a better solution might be to have the logging function return true if the user is not staff...

AW0005:
I can do that! I honestly had the same thought but didn't immediately have a solution.

itinerare:
Fair enough!

itinerare (approved):

preimpression (approved):

PR #336, opened 2022-07-01
User Profile Character/Sublist pages edited

SpeedyD:
- Swapped Encoding to UTF-8 from ANSI
- Made it look consistent with Masterlist

(PR is also possible for main branch, but I figure develop is the better call.)


itinerare (approved):
Could potentially be a matter of opinion, but I like the change. Folk can adjust as they see fit regardless, so!

preimpression (approved):

PR #341, opened 2022-07-13
Prompt permalinks

AW0005:
I've got this as an extension right now:
http://wiki.lorekeeper.me/index.php?title=Extensions:Prompt_Permalink

But it was mentioned in discord that this would be a good candidate for including in core.

### DIFF
--- a/app/Http/Controllers/PromptsController.php
+++ b/app/Http/Controllers/PromptsController.php
@@ -112,6 +112,9 @@ class PromptsController extends Controller
     public function getPrompt(Request $request, $id)
     {
         $prompt = Prompt::where('id', $id)->get()->first();
+
+        if(!$prompt) abort(404);
+
         return view('prompts.prompt', [
             'prompt' => $prompt,
         ]);
--- a/app/Http/Controllers/PromptsController.php
+++ b/app/Http/Controllers/PromptsController.php
@@ -102,4 +102,18 @@ class PromptsController extends Controller
             'categories' => ['none' => 'Any Category'] + PromptCategory::orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
         ]);
     }
+
+    /**
+     * Shows an individual prompt.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Contracts\Support\Renderable
+     */
+    public function getPrompt(Request $request, $id)
+    {
+        $prompt = Prompt::where('id', $id)->get()->first();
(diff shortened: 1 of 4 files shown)


app/Http/Controllers/PromptsController.php
```diff
@@ -102,4 +102,18 @@ public function getPrompts(Request $request)
...
+
+    /**
+     * Shows an individual prompt.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Contracts\Support\Renderable
+     */
+    public function getPrompt(Request $request, $id)
+    {
+        $prompt = Prompt::where('id', $id)->get()->first();
```
itinerare, line 114: Probably add a `if(!$prompt) abort(404);` or somesuch as a fallback/to prevent errors

resources/views/prompts/prompt.blade.php
```diff
@@ -0,0 +1,8 @@
+@extends('prompts.layout')
+
+@section('title') All Prompts @endsection
```
itinerare, line 3: Maybe adjust this to be something like "Prompts :: (prompt title)"?

resources/views/prompts/prompt.blade.php
```diff
@@ -0,0 +1,8 @@
...
+
+@section('content')
+    {!! breadcrumbs(['Prompts' => 'prompts', 'All Prompts' => 'prompts/prompts']) !!}
```
itinerare, line 6: Should probably add the prompt title and relative address to the breadcrumbs!

itinerare (changes requested):
Code broadly looks good too, just needs some minor adjustments!

AW0005:
@itinerare Changes should be up!

itinerare (approved):

SpeedyD (approved):
Figured I'd give it a do-over as well. Code looks good and clean.

Only question I had is why the change from `<h4>` to `<h3>` for the Details and Rewards, but that's kind of a minor non-issue.

AW0005:
@SpeedyD I think I nudged all the headers up by one from what is was before to fit better in the full page format.

SpeedyD:
> @SpeedyD I think I nudged all the headers up by one from what is was before to fit better in the full page format.

Fair enough! I'm honestly not sure if I'll even change it back myself.. I tried it out with my browser inspector and frankly, it doesn't look that bad either. Just made me go :thinking: basically. :+1:

PR #342, opened 2022-07-15
refactor: stop character's names from traversing off-screen

Ne-wt:


SpeedyD:
Pardon, but may I request the change done to `resources/views/user/characters.blade`
is repeated for `resources/views/user/sublist.blade`?

itinerare:
Could incorporate #343 since it'd cut that problem off at the pass also for future changes-- I can sort that out real quick if desired as well, just holler at me

SpeedyD (approved):
Seems fine, though as per my previous comment, https://github.com/corowne/lorekeeper/pull/342#issuecomment-1186018075, I'd like to see this applied to `resources/views/user/sublist.blade.php` as well.

Perhaps I can do the holler @itinerare requires? ^^;

itinerare:
Oh yeah I'll just merge and handle it, no worries.

PR #364, opened 2022-08-14
Feat: Adds Admin "Edit" button

preimpression:


resources/views/raffles/ticket_index.blade.php
```diff
@@ -27,7 +27,8 @@
                         <th>User</th>
                     </thead>
                     <tbody>
-                        @foreach ($raffle->tickets()->winners()->get() as $winner)
+                        @foreach ($raffle->tickets()->winners()->get()
+        as $winner)
```
preimpression, line 31: I will burn the auto-formatter's crops and salt their fields

itinerare: Understandable, have a nice day

app/Models/Species/Subtype.php
```diff
@@ -147,4 +147,24 @@ public function getUrlAttribute() {
...
+    }
+
+    /**
+     * Gets the power required to edit this model.
+     *
+     * @return string
+     */
+    public function getAdminPowerAttribute()
+    {
+        return url('edit_data');
```
itinerare, line 168: Is there a reason this is in the `url()` helper rather than just returning the string? Likewise for the other instances of this in the models.

itinerare (commented):
Broadly looks good and I like the implementation as a component!

app/Models/Species/Subtype.php
preimpression: oh my god.
the reason is im a dumbass let me fix that

itinerare: It Happens To The Best Of Us

itinerare (approved):

Ne-wt (approved):

PR #380, opened 2022-08-30
More misc tweaks/fixes

itinerare:
Primarily:
- Add a visibility toggle to traits
- Fix an exploit allowing users to purchase currency with another user's character's owned currency

Tested locally, requires `migrate`.


preimpression (approved):

SpeedyD (approved):
Code looks good to me.

PR #396, opened 2022-09-13
Making alt text for avatars consistent

SpeedyD:
As per title. Extremely minor fix.


SpeedyD:
As for why I didn't change the titles: It is because that aside from the gallery submission, it didn't make much sense.

Anywhere where an avatar is shown, it's primarily a specific user; for comments the name is mentioned right after, for the profile you're literally on that user's profile..

But for the gallery submission, the artist can be someone different than the person submitting!

(Nobody asked, I know, but my previous thought was to make those consistent too. In hindsight, not smart.)

itinerare (approved):

preimpression (approved):

PR #415, opened 2022-10-06
Adds sidebars to News and Sales

SpeedyD:
Sidebars display Recent News/Sales on posts, and 'On This Page' on the index pages.
For Sales specifically, both posts and the index pages also have a section 'For Sale', which are only __open__ sales.

Original inspiration, most of the code and idea by @itinerare.

### DIFF
--- a/resources/views/sales/_sidebar.blade.php
+++ b/resources/views/sales/_sidebar.blade.php
@@ -2,7 +2,7 @@
     <li class="sidebar-header"><a href="{{ url('sales') }}" class="card-link">Sales</a></li>
     <li class="sidebar-section">
         <div class="sidebar-section-header">For Sale</div>
-        @foreach (App\Models\Sales\Sales::visible()->orderBy('updated_at', 'DESC')->where('is_open', 1)->get() as $sales)
+        @foreach ($forsale as $sales)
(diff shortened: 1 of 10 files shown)


resources/views/sales/_sidebar.blade.php
```diff
@@ -0,0 +1,26 @@
+<ul>
+    <li class="sidebar-header"><a href="{{ url('sales') }}" class="card-link">Sales</a></li>
+    <li class="sidebar-section">
+        <div class="sidebar-section-header">For Sale</div>
+        @foreach (App\Models\Sales\Sales::visible()->orderBy('updated_at', 'DESC')->where('is_open', 1)->get() as $sales)
```
itinerare, line 5: No fault of your own, but it would probably be more practical to reimplement this as a variable supplied via `__construct()` in the controller... See [here](https://github.com/corowne/lorekeeper/commit/429d7fe8f84d71e3994f13d706d7302a69f2fad1) for example

itinerare (approved):
Approving this as-is since the changes that I'd want here are both noncritical and the fault of my own past code and I need to kick up a misc PR soonish anyway.

itinerare (approved):

preimpression (approved):

Issue #459, opened 2022-12-09
500 error when character on sale is deleted


SpeedyD:
### Bug as reported:

(Main branch, last commit parent: [da3ebb11](https://github.com/corowne/lorekeeper/commit/da3ebb11fee2c55e388727bed70e385f0c0592c5))

**URL of page:** `<site>/sales`, `<site>/sales/<id>` and `<site>/admin/sales/edit/<id>` (where <id> is the id of the sale in question)
**Description of bug:** When a character has been deleted entirely, and they were part of a sales post, the sale in question as well as the admin page on which it is editable, will give out a 500 error. (And of course any of the sales pages that sale is displayed on.)
**How was the bug caused (exact steps to reproduce):** Add a character to a sale. Delete the character. Go to the sale in either public area or the editing page of the admin section.
### Further details
..Yeah, I guess this is a thing that happens. Should be a decently easy fix, just have a check if it exists and if not display something like 'Character Not Found' like with gallery images.

~~PR #462 should fix this, eventually.~~

PR #462, opened 2022-12-09
Fix(Sales): Resolve "characters not found" issue

SpeedyD:
Referring to Issue #459 that is-

I feel like this is a form of band-aid solution, but it should look and work decently. Please look it over thoroughly.


resources/views/admin/sales/_character_select_entry.blade.php
```diff
@@ -14,7 +14,7 @@
                 <a href="#" class="float-right fas fa-close"></a>
                 <div class="form-group">
                     {!! Form::label('slug[]', 'Character Code') !!}
-                    {!! Form::text('slug[]', $character->character->slug, ['class' => 'form-control character-code']) !!}
+                    {!! Form::text('slug[]', $character->character ? $character->character->slug : null, ['class' => 'form-control character-code']) !!}
```
AW0005, line 17: Nit: I think you should be able to use the `??` operator here to make these a little shorter like this:
`$character->character->slug ?? null`

AW0005 (approved):
I left a little nitpick comment and thought the rest of this looked good!

But I also noticed in the description you felt like this was a band-aid solution so I did a little poking around and I think another option might be to update the relationship on `Sales` for the `SalesCharacter to be this:
`$this->hasMany('App\Models\Sales\SalesCharacter', 'sales_id')->has('character')`

So it would never return SalesCharacters as a part of the collection at all, if their character is undefined. The downside to that though, is then there's no messaging to the user that a character was deleted and that's why it's missing. 🤔 Not sure which would be more preferable.

itinerare:
Yeah I had the idea to have better checking on the interface itself so that the information entered pertinent to the sale was still accessible; ideally I think this would be the case for both the admin and public-facing interfaces so that records are maintained, just replacing relevant bits with [deleted character] or somesuch as is done for e.g. logs.

SpeedyD:
@AW0005, @itinerare had personally told me they'd look things over and work on it, so I'm leaving it 100% up to them from here. 
I just wanted to set up a temporary solution that worked.. and it does, even if it's not that clean. ^^;

resources/views/admin/sales/_character_select_entry.blade.php
SpeedyD: Would you believe I wasn't aware of ternary operators 3 months ago? Anyway, good note for @itinerare if/when they fix this. ^^;

itinerare:
Yeah I had some bold thoughts about getting character info by virtue of it being soft-deleted rather than deleted-deleted... though tbh this probably wouldn't work as well as I'd like considering that iirc the core character is the only thing to which that applies and not e.g. the image, traits, etc, which are kinda the relevant bits of info that would contextualize the sale info. At this rate probably what I would end up doing is just cleaning up the display in line w my comment above.

SpeedyD:
Just refreshing the branch by putting all the new stuff in..

SpeedyD:
..Hrm, I have an idea how to do this the way Merc maybe wants, but I think I need #615 resolved first. It might help the process along..

So I'm converting this to draft for now.

SpeedyD:
Don't know if still needed, but I've opted to at least push the dev branch in for now, resolving the one conflict

SpeedyD:
..Scratch that. Looking at the changes that have been done and the two edits in here do not even get reached anymore. Closing PR.

PR #475, opened 2022-12-22
feat(species): add visibility toggle

SpeedyD:
Adds a visible toggle for Species.
Will hide the species in the encyclopedia, search function of the masterlist/sublist/myo list and design update.

### DIFF
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -105,7 +105,7 @@ class WorldController extends Controller {
         }
 
         return view('world.subtypes', [
-            'subtypes' => $query->with('species')->visible()->orderBy('sort', 'DESC')->paginate(20)->appends($request->query()),
+            'subtypes' => $query->with('species')->orderBy('sort', 'DESC')->paginate(20)->appends($request->query()),
         ]);
     }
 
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/WorldController.php
```diff
@@ -105,7 +105,7 @@ public function getSubtypes(Request $request) {
         }
 
         return view('world.subtypes', [
-            'subtypes' => $query->with('species')->orderBy('sort', 'DESC')->paginate(20)->appends($request->query()),
+            'subtypes' => $query->with('species')->visible()->orderBy('sort', 'DESC')->paginate(20)->appends($request->query()),
```
itinerare, line 108: This needs to be a separate scope on the subtype model, probably checking the visibility of the parent species at minimum...

itinerare (changes requested):

app/Http/Controllers/WorldController.php
SpeedyD: You are absolutely right- for now I'm removing this and will put it on my to-do list for a future 'subtype-visibility' feature.
Until then, administration just needs to be careful not to add subtypes to species they have hidden..

itinerare (approved):

SpeedyD:
**DO NOT APPROVE!** Found critical bug.

itinerare (approved):

Draginraptor (approved):

PR #493, opened 2023-01-12
Adds Alias Login's to Develop

AW0005:
Same instructions as the extension for setting up:
http://wiki.lorekeeper.me/index.php?title=Extensions:Alias_Logins


Ne-wt:
might be a good idea to squash these commits to remove the original hard coding of TH (use git --reset soft)

itinerare:
I was planning on squash merging regardless for good commit formatting so it shouldn't be a trouble

itinerare:
Noting here that I've looked over the code and just want to play around w/ it a bit before approving!

resources/views/auth/login.blade.php
```diff
@@ -67,6 +67,16 @@
                 </div>
             </div>
         </form>
+        
+         <h3 class="text-center mt-5 pt-2">Alternate Logins</h3>
```
Draginraptor, line 71: Check if any sites are allowed for login, and show/hide the header based on if there are some/none.

resources/views/auth/register.blade.php
```diff
@@ -11,6 +11,16 @@
                 <h1>Register</h1>
             </div>
         </div>
+        <h3 class="text-center">Alternate Registrations</h3>
```
Draginraptor, line 14: Same comment as the login blade - could check if any of the sites are allowed, and show/hide the header accordingly.

Draginraptor (approved):

AW0005:
Update made for hiding the headers!

AW0005:
Just pushed an update for a fix for the twitter socialite provider that does require a little bit of additional setup that I added to the extension page as well:
http://wiki.lorekeeper.me/index.php?title=Extensions:Alias_Logins

The trouble is that OAuth1 version of socialite providers don't give a way to programmatically provide an alternate redirect url which this makes use of. Laravel Socialite however has a built in Oauth2 provider for twitter so this has updated twitter auth to be based on that. Tested regular auth as well as login, all seems to work fine.

I did also notice that tumblr seems to be the one other socialite provider we have that uses oauth1 - so may need to put a guard against that, since I don't see any other alternative from my quick googling, besides attempting to rebuild it with OAuth2.

itinerare:
I think setting a check up for tumblr and noting that atm it's not a supported option should be fine...

Ne-wt (approved):

itinerare:
Good to hear it + agreed-- ty for testing!

PR #495, opened 2023-01-13
Feat: add usage of gravatar for user avatars

Ne-wt:


itinerare (approved):

Draginraptor (approved):

app/Models/User/User.php
```diff
@@ -333,6 +333,28 @@ public function getAvatar()
...
+    public function getAvatarUrlAttribute()
+    {
+        if ($this->avatar == 'default.jpg') {
+            // check if a gravatar exists
+            $hash = md5(strtolower(trim($this->email)));
+            $url = 'https://www.gravatar.com/avatar/'.$hash.'??d=mm&s=200';
+            $headers = @get_headers($url);
+            return $url;
+            if (!preg_match("|200|", $headers[0])) {
+                return url('assets/images/avatars/default.jpeg');
```
SpeedyD, line 350: ...Isn't this a problem? Filename is `default.jpeg`, while the default file is `default.jpg`.

SpeedyD (changes requested):
Looks fine except for the jpeg vs jpg file extension..

app/Models/User/User.php
Draginraptor: I wanted to look into this more, and also realised that there's also `return $url;` on line 348, which means the if-else wouldn't be reached.

That said, `url('assets/images/avatars/default.jpeg')` should also be `url('images/avatars/default.jpg')` (inclusive of the jpeg to jpg change.

Draginraptor (changes requested):
Check over the comments on the User model's getAvatarUrlAttribute function

itinerare (approved):

AW0005 (approved):

PR #497, opened 2023-01-14
feat(world): search and sort traits by subtype

SpeedyD:
Adds both a subtype search function and a subtype sorting function to the Traits section of the Encyclopedia,
allowing users to search for subtypes or to sort by subtypes!


app/Http/Controllers/WorldController.php
```diff
@@ -199,6 +205,7 @@ public function getFeatures(Request $request) {
             'features'   => $query->paginate(20)->appends($request->query()),
             'rarities'   => ['none' => 'Any Rarity'] + Rarity::orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
             'specieses'  => ['none' => 'Any Species'] + Species::visible(Auth::check() ? Auth::user() : null)->orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
+            'subtypes'   => ['none' => 'Any Subtype'] + Subtype::visible(Auth::check() ? Auth::user() : null)->orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
```
Draginraptor, line 208: I'm getting an error where Auth is not found, presumably because `use Auth;` is missing.

Also this includes an attempt to use the `visible()` scope -  but it is not yet implemented for subtypes.

I think it would be better to merge this pull request and the draft pull request to create one pull request that covers both?

SpeedyD: Whoops! My intention was to leave this out and put it in #498, but I must've accidentally snuck it in regardless, my bad..

Auth error is confusing, though- as specieses right above it also uses it??

@itinerare, what is your thought on Auth problem above, but also about merging #498 into this??

itinerare: `use Auth;` should be added
also probably it's easier to just handle them at once

SpeedyD: Will do! Thanks @itinerare, @Draginraptor.

SpeedyD:
Added `use Auth;` as requested and merged in #498.

Draginraptor (approved):

itinerare (approved):

PR #498, opened 2023-01-14
feat(subtypes): add visibility toggle

SpeedyD:
Adds a visible toggle for Subtypes.

Will hide the subtype in the encyclopedia, search function of the masterlist/sublist and design update.

**DRAFT:** Waiting for #497, so I can hide the subtype in the search function too.


SpeedyD:
To be merged into #497 instead.

PR #507, opened 2023-01-18
feat(world): Search traits (and items & prompts)

SpeedyD:
...by those lacking species, subtypes and categories!

(Items and prompts specifically is category only, obviously.)

**NOTE:**
Something went wrong with my previous PR, so I redid it. My apologies for the other PR.

### DIFF
--- a/app/Http/Controllers/PromptsController.php
+++ b/app/Http/Controllers/PromptsController.php
@@ -52,7 +52,7 @@ class PromptsController extends Controller {
     public function getPrompts(Request $request) {
         $query = Prompt::active()->staffOnly(Auth::check() ? Auth::user() : null)->with('category');
         $data = $request->only(['prompt_category_id', 'name', 'sort', 'open_prompts']);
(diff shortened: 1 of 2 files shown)


Draginraptor (commented):
Would "No [__]" be better as a label? For a bit I was confused since I thought that "No Specific [__]" was the same as "Any [__]", and was confused by the use of whereNull until it finally clicked for me. This could also be a "just me" thing though.

itinerare:
I'd be inclined to agree with that, yeah

SpeedyD:
To be honest, deciding on the function's name has been the one thing I've been hoping to get comments on.

I've actually been thinking about "Without []", but am not sure..

SpeedyD:
@itinerare @Draginraptor
Altered the wording to be "Without []" instead, even altered the variable name to `withoutOption` for clarification and consistency.

..Feeling a LITTLE dumb about my quick replacing accidentally removing the spacing, but the PHP auto-job picked that up. :)

itinerare (approved):

AW0005 (approved):

PR #508, opened 2023-01-22
 feature(Characters): Added Previous and Next Buttons

SpeedyD:
Which you can change with a extension setting. By default off.

### DIFF
--- a/app/Http/Controllers/Characters/CharacterController.php
+++ b/app/Http/Controllers/Characters/CharacterController.php
@@ -21,6 +21,7 @@ use App\Services\DesignUpdateManager;
 use App\Services\InventoryManager;
 use Auth;
 use Illuminate\Http\Request;
+use Illuminate\Support\Facades\View;
 use Route;
 use Settings;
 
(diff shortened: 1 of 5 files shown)


app/Http/Controllers/Characters/CharacterController.php
```diff
@@ -51,6 +52,39 @@ public function __construct() {
...
+            }
+            $nextCharacter = $query->get()->last();
+            if (!$nextCharacter || $nextCharacter->id == $this->character->id) {
+                $nextCharacter = null;
+            }
+            else {
+                $nextCharName = $nextCharacter->fullName;
+                $nextCharUrl = $nextCharacter->url;
+            }
+            View::share('prevCharName', $prevCharName);
```
itinerare, line 84: Might be marginally cleaner to just put all the info in an array and just pass that over to the view instead? Since then it'd be just the one variable. This is more a nitpick than anything though

app/Http/Controllers/Characters/CharacterController.php
```diff
@@ -77,8 +111,10 @@ public function getCharacter($slug) {
      * @return \Illuminate\Contracts\Support\Renderable
      */
     public function getCharacterProfile($slug) {
+		$url_addition = "/profile";
         return view('character.profile', [
             'character' => $this->character,
+			'url_addition' => $url_addition,
```
itinerare, line 117: You can just put a string in the array that's supplied to the view.... you don't need to define a variable with it and supply it that way
We also don't generally use underscores in variables like this but that is more a style consistency nitpick than an actual problem,

resources/views/character/_header.blade.php
```diff
@@ -1,3 +1,23 @@
+@if (!$character->is_myo_slot && Config::get('lorekeeper.extensions.display_previous_and_next_characters'))
+    @if ($prevCharName || $nextCharName)
+        <div class="row mb-4">
+            @if ($prevCharName)
+                <div class="col text-left float-left">
+                    <a class="btn btn-outline-success text-success" href="{{ $prevCharUrl }}{!! !isset($url_addition) ? '' : $url_addition !!}">
+                        <i class="fas fa-angle-double-left"></i> Previous Character ・ <span class="text-primary">{!! $prevCharName !!}</span>
```
itinerare, line 7: Good styling on these buttons incidentally, looks good 👍

itinerare (changes requested):
No functionality concerns (obv it works just fine as-is), just some tidying that needs doing.

app/Http/Controllers/Characters/CharacterController.php
SpeedyD: Done and done.

app/Http/Controllers/Characters/CharacterController.php
SpeedyD: I made it a marginally larger variable, but this should be more unique and consistent at least.

resources/views/character/_header.blade.php
SpeedyD: Thank you! I also edited this part, obviously. :)

itinerare (approved):

itinerare (approved):

PR #509, opened 2023-01-22
fix(deactivation): Function doesn't exist, fixed

SpeedyD:
uc_first() is a function that doesn't exist.
ucfirst() does.


preimpression (approved):

preimpression:
Oops

itinerare (approved):

PR #522, opened 2023-02-01
add optional aliases

ScuffedNewt:


app/Console/Commands/AddSiteSettings.php
```diff
@@ -83,6 +83,8 @@ public function handle() {
 
         $this->addSiteSetting('comment_dislikes_enabled', 0, '0: Dislikes disabled, 1: Dislikes enabled.');
 
+        $this->addSiteSetting('enable_optional_alias', 0, '0: Optional alias disabled, 1: Optional alias enabled.');
```
itinerare, line 86: Going to switch this to config setting myself per discussion

itinerare (approved):

itinerare (approved):

preimpression (approved):

resources/views/account/aliases.blade.php
```diff
@@ -55,7 +55,7 @@
                         </div>
                         <div class="col-5 text-right">
                             <div class="logs-table-cell">
-                                @if (!$alias->is_primary_alias)
+                                @if (!$alias->is_primary_alias || !config('lorekeeper.settings.require_alias'))
```
AW0005, line 58: This bit of logic is causes the "make primary" button to show up even if an alias is already the primary alias, which then generates a `Invalid alias selected.` error if you click on it. 

Might be good to add a check to hide that button when an alias is already primary?

app/Http/Middleware/CheckAlias.php
```diff
@@ -18,7 +18,7 @@ public function handle($request, Closure $next) {
         if (Settings::get('is_maintenance_mode') == 1 && !$request->user()->hasPower('maintenance_access')) {
             return redirect('/');
         }
-        if (!$request->user()->has_alias) {
+        if (!$request->user()->has_alias && config('lorekeeper.settings.require_alias')) {
```
AW0005, line 21: I think the extra check added here miiiight not be necessary? Since `has_alias` will return true if the config setting is on.

AW0005 (approved):
Had a couple of thoughts I put down, but I also poked at this a bit locally and couldn't manage to break it :D

resources/views/account/aliases.blade.php
itinerare: This might just be misplaced from what I remember of the view. I'll take a look at it!

app/Http/Middleware/CheckAlias.php
itinerare: A good point, yeah.

preimpression (approved):

AW0005 (approved):

PR #523, opened 2023-02-03
hotfix: Fixing error with lowercase "auth"

preimpression:


SpeedyD:
If I may make a suggestion, aside from this file, the last remaining file with lowercase `auth::` is `app\Policies\CommentPolicy.php`..

I would like to ask if that one could also be added to this patch.

SpeedyD (approved):

itinerare (approved):

itinerare:
Holding this for a moment pending some other in-progress "hotfix" type work that's liable to go into main, but looks good.

PR #535, opened 2023-02-08
PHP 8.1 upgrade

itinerare:
Reportedly, Dreamhost is pushing upgrading sites to PHP 8.1 due to support ending for 7.4... which is understandable for security etc reasons, and probably an idea, but annoying from a logistical standpoint.
However this is salvageable without too significant changes since Laravel 8 supports PHP 8.1 at maximum and takes surprisingly few changes to arrive at a compatible set of packages. The main things are just allowing Laravel (within the same major version) and doctrine/dbal (fine since the issue we were prior having with it looks to be resolved anyhow) to update.

This has been tested a bit already, and looks to be working alright, hence the PR, but more is always welcome.
Necessary actions are just updating PHP itself and running composer.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.4",
+        "php": "~8.1",
         "boxconnect/deviant-php": "dev-master",
         "doctrine/dbal": "2.2",
         "erusev/parsedown": "^1.7",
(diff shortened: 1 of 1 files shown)


preimpression (approved):

AW0005 (approved):
Successfully used this on a fresh Dreamhost install a little while ago so I'd consider that a pretty good test :D

PR #555, opened 2023-03-09
Adjust Prev and Next Buttons to respect when 'character_pull_number' is 'all',

AW0005:
If a LK site was using `'character_pull_number' => 'all'` in settings and tried to use the new previous and next buttons a lot of characters would have oddly missing buttons to the non-sequential numbers between categories. This adjusts the prev and next buttons to respect the LK setting and only filter by category if it's set that way.


itinerare (approved):

SpeedyD (approved):
Didn't think about this angle. Nice job, I approve! :)

PR #556, opened 2023-03-11
Feat(character): Updating previous and next buttons to skip the gaps

SpeedyD:
Have been thinking for a while to update the previous and next buttons.. so they no longer skip over the gaps- displaying no previous or next button just because in a sequence of 1-2-4-5 there is no 3, so no previous on 4 and no next on 2.. that's silly.

Spoke about it with @itinerare who quickly produced code, but it required some testing.. and some fixing. And that's what I did.

Also special mention to @AW0005 whose recent edit to the code made me consider doing this. :)

### DIFF
--- a/config/lorekeeper/extensions.php
+++ b/config/lorekeeper/extensions.php
@@ -77,7 +77,7 @@ return [
     // Previous & Next buttons on Character pages - Speedy
     // Adds buttons linking to the previous character as well as the next character on all character pages.
     'previous_and_next_characters' => [
-	    'display' => 1,
-		'reverse' => 0, // By default, 0 has the lower number on the 'Next' side and the higher number on the 'Previous' side, reflecting the default masterlist order. Setting this to 1 reverses this.
-	],
+        'display' => 1,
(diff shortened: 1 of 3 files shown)


SpeedyD:
On advice from @itinerare, I will be editing the code to be more consistent with other code soon.
I will also be swapping the previous and next links to be more consistent with default ML order, but will also add an option to keep it as it currently is.

Please refrain from reviewing until this has been done. Thank you.

SpeedyD:
...aaand done. Ready for reviews.

config/lorekeeper/extensions.php
```diff
@@ -76,5 +76,8 @@
 
     // Previous & Next buttons on Character pages - Speedy
     // Adds buttons linking to the previous character as well as the next character on all character pages.
-    'display_previous_and_next_characters' => 0,
+    'previous_and_next_characters' => [
+        'display' => 1,
```
AW0005, line 80: Nit: Did you mean to leave it on by default? Wouldn't ask but it was turned off prior.

AW0005 (approved):
Poked around locally and it looks good to me :D

config/lorekeeper/extensions.php
SpeedyD: Nope, absolutely forgot to turn it off after testing. Thanks!

itinerare (approved):

PR #557, opened 2023-03-11
bug(prompts): Category Filter value being set to name

AW0005:
Just a teeny bug fix that I noticed, where the selected category on the prompts filter wasn't "sticking" on clicking search because it's value was being set to name instead.

### DIFF
--- a/resources/views/admin/features/features.blade.php
+++ b/resources/views/admin/features/features.blade.php
@@ -31,7 +31,7 @@
             {!! Form::select('rarity_id', $rarities, Request::get('rarity_id'), ['class' => 'form-control']) !!}
         </div>
         <div class="form-group mr-3 mb-3">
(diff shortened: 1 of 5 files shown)


SpeedyD (approved):
Wow! I literally recently changed the EXACT SAME ISSUE for traits! I should've checked if it was the same in prompts. Edit approved.

AW0005:
@SpeedyD You saying that made me think to go double check... and I found a few more of them throughout the code base and fixed those up too.

SpeedyD:
@AW0005 Glad to be of service then! I completely did not think of the admin section..

Which is funny because I am FAIRLY sure I've been having issues with the traits section there.

itinerare (approved):
Always good to see more of these squashed.

PR #559, opened 2023-03-13
feat(Prompts): Prompt Calendar

SpeedyD:
I should probably have made this an extension, but frankly, it fits better on the dev branch.

Adds a Prompt Calendar. Nothing fancy. Uses [FullCalendar](https://fullcalendar.io) through CDN. There's probably a better way to do this.


SpeedyD:
Screw it, too much minor things to fix still, so I'm setting it to draft for now.

itinerare:
I think my one immediate comment is to just fold the calendar JS into the files rather than access it via CDN...

SpeedyD:
> I think my one immediate comment is to just fold the calendar JS into the files rather than access it via CDN...

Done that, at least.

SpeedyD:
Screw it. It's been a year. I'm cancelling this draft and deleting my work. I think I can safely say Newt's version at #670 is doing this job better.

PR #569, opened 2023-03-30
bug(items): item images weren't deleting

AW0005:
Since the `$image` variable wasn't being passed to the `populateData` function the code to delete an existing image on an item was never being run.


itinerare (approved):

SpeedyD (approved):
Sorry for slow response- didn't understand what was happening from just the snippet. Finally took the time to look into it.. and now I understand why this is necessary. Absolutely approve.

PR #593, opened 2023-05-11
fix(galleries): Sidebar now shows active galleries + fix:(version) updated to 3.0.0

SpeedyD:
Minor patch, makes active galleries now show up properly

**Add: Also changed version number, see later comments why**


itinerare (approved):

SpeedyD:
..what the heck, refactor action?

...hm, maybe I should purposely trigger the blade one too, hang on

SpeedyD:
...and ofcourse the action errored out. Ah well.

AW0005 (approved):

PR #594, opened 2023-05-11
feat(gallery): Show All Recent Submissions

SpeedyD:
Adds a page that shows all recent submissions!

Is disabled by default. Will show up in sidebar and as button on gallery index depending on options set.
Added bonus is recent submissions showing up on the frontpage, which is another featured.. disabled by default.


AW0005 (approved):

preimpression (approved):

app/Http/Controllers/HomeController.php
preimpression, line 1: Maybe instead of immediately pulling the query, it could first check if it's enabled? That way it reduces processing power if it's disabled

SpeedyD: @itinerare if you see this: I'll look into this tomorrow. Don't merge, please. :)

SpeedyD: Fixed it!

preimpression (approved):

PR #601, opened 2023-05-19
patch(sales): Allows hidden characters to be added to sales.

SpeedyD:
Displays warning on characters hidden from public view!


itinerare (approved):

resources/views/home/_character.blade.php
preimpression, line 1: `Auth::check()` would be good here pre-`Auth::user()`

preimpression, line 1: `Auth::check()` would be good here pre-`Auth::user()`

preimpression (changes requested):

resources/views/home/_character.blade.php
SpeedyD: Thanks, I forgot! XD

preimpression (approved):

PR #614, opened 2023-06-03
Feat(site): Modifying pagination to allow page select on ...

SpeedyD:
~~All this PR does so far is replace the bootstrap defaults with.. the same views from the vendor.~~

~~However, that makes it accessible to modify, which was the goal.~~

As of commit https://github.com/corowne/lorekeeper/pull/614/commits/43681fcc80f5365f0fd6d105b971f8658908c04f, each '...' block now opens a popover where you can change the page.

The functionality itself wasn't too bad, but the fact that each popover had to be unique.. whew.

### DIFF
--- a/resources/views/layouts/_pagination.blade.php
+++ b/resources/views/layouts/_pagination.blade.php
@@ -1,0 +1,46 @@
+@if ($paginator->hasPages())
+    <nav>
+        <ul class="pagination">
+            {{-- Previous Page Link --}}
+            @if ($paginator->onFirstPage())
+                <li class="page-item disabled" aria-disabled="true" aria-label="@lang('pagination.previous')">
+                    <span class="page-link" aria-hidden="true">&lsaquo;</span>
+                </li>
+            @else
+                <li class="page-item">
+                    <a class="page-link" href="{{ $paginator->previousPageUrl() }}" rel="prev" aria-label="@lang('pagination.previous')">&lsaquo;</a>
+                </li>
+            @endif
+
+            {{-- Pagination Elements --}}
+            @foreach ($elements as $element)
+                {{-- "Three Dots" Separator --}}
+                @if (is_string($element))
+                    <li class="page-item disabled" aria-disabled="true"><span class="page-link">{{ $element }}</span></li>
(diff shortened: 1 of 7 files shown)


resources/views/layouts/_pagination.blade.php
```diff
@@ -16,7 +16,19 @@
...
+                    <li class="page-item pageSelectPopover{{ $instanceID }}" data-container="body" data-toggle="popover" data-placement="top" data-title="Jump to Page" data-html="true" data-content='
+                    <form>
+                    <div class="form-group justify-content-center">
+                            <input type="range" class="form-control-range custom-range" id="paginationPageRange{{ $instanceID }}" min="1" max="{{ $paginator->lastPage() }}" value="{{ $paginator->currentPage() }}" onchange="pageUpdateSelectText{{ $instanceID }}()">
+                            <input type="number" id="paginationPageText{{ $instanceID }}" min="1" max="{{ $paginator->lastPage() }}" value="{{ $paginator->currentPage() }}" onchange="pageUpdateSelectRange{{ $instanceID }}()">
+                            <button type="button" class="btn btn-primary" onclick="pageGo{{ $instanceID }}()">Go</button>
+                        </div>
+                    </form>
+                    '><span class="page-link">{{ $element }}</span></li>
+                    @include('layouts._pagination_js', ['ID' => $instanceID])
```
itinerare, line 33: Why feed ID to the include when `$instanceID` continues to work within the include and `$ID` isn't used?

SpeedyD: ...because that was a test I forgot to remove. *facepalm*

itinerare (approved):

SpeedyD:
Moif cleaned up the code extremely well, half want to addcredits just so @AW0005 would get co-credits XD

preimpression (approved):
Could do with a toggle for whether the three dots separator has the popover, but regardless looks good

itinerare:
I think it's both unobtrusive and useful enough that it's fine to just have by default.

itinerare (approved):

PR #615, opened 2023-06-04
Bug(Sales): MYO made characters show old MYO image

AW0005:
Sales image attribute references the "first" image for a character, which means that if you have a character made on a MYO slot with the default image still attached to them, that's what shows up in the sale image instead of the correctly updated active image.

Switching the attribute to reference the attached character's set active image instead fixes it up.

This is somewhere in the grey area of being a bug vs a feature enhancement but I figured it was close enough to unwanted behavior to title it a bug.

### DIFF
--- a/app/Models/Sales/SalesCharacter.php
+++ b/app/Models/Sales/SalesCharacter.php
@@ -178,6 +178,6 @@ class SalesCharacter extends Model {
      * @return App\Models\Character\CharacterImage
      */
     public function getImageAttribute() {
-        return CharacterImage::where('is_visible', 1)->where('character_id', $this->character_id)->orderBy('created_at')->first();
+        return $this->character->image;
     }
 }
--- a/app/Models/Sales/SalesCharacter.php
+++ b/app/Models/Sales/SalesCharacter.php
@@ -2,7 +2,6 @@
 
 namespace App\Models\Sales;
 
-use App\Models\Character\CharacterImage;
 use App\Models\Model;
 use Config;
 


itinerare:
Hmmm. I think probably the thing to do would just to make it toggleable, as I feel strongly that it should remain the "first" image for archival purposes and think having a sales post with a character from a MYO slot profoundly _weird_, but I also recognize that that is not the beginning and end of use cases for the thing.

SpeedyD:
I second @itinerare's suggestion. If you add a toggle, this is an instant approve from me. For archival purposes, we'd prefer the old image..

..In all honesty, I feel like there may need to be something more intricate to preserve old sales, but using first is the easiest way to do it rn.

itinerare:
A potential option re that is recording the image ID at the time the character is added to the sale (probably the best balance between ease-of-use and solving this kind of issue/covers the most use cases most readily), but that would be rather more work than just setting up a toggle.

SpeedyD:
For now, I'm suggesting @AW0005 just adds a toggle, and maybe we can look into _that_ another time, @itinerare. :)

itinerare:
Yeah, hence the [or you could just add a toggle and be done with it].

AW0005:
@itinerare @SpeedyD 
Updated to track the character image id at time of sale listing. I wasn't fond of the idea of leaving in something that works by coincidence over going ahead and implementing something that should work as expected for all cases.

app/Models/Sales/SalesCharacter.php
```diff
@@ -177,6 +178,6 @@ public function getPriceAttribute() {
      * @return App\Models\Character\CharacterImage
      */
     public function getImageAttribute() {
-        return $this->character->image;
+        return $this->image_id ? CharacterImage::find($this->image_id) : CharacterImage::where('is_visible', 1)->where('character_id', $this->character_id)->orderBy('created_at')->first();
```
itinerare, line 181: I think for safety's sake it'd be prudent to
1. add a relation for the image
2. switch this to `return $this->image_id && $this->image ? $this->image : CharacterImage::(etc etc)`
so that even if the image with the recorded ID is deleted, there's a fallback.

app/Services/SalesService.php
```diff
@@ -232,6 +232,7 @@ private function processCharacters($sales, $data) {
             // Record data/attach the character to the sales post
             SalesCharacter::create([
                 'character_id' => $character->id,
+                'image_id'     => $character->image->id,
```
itinerare: While functional in theory, this will essentially set the sales character's image every time the associated sales post is edited, as (on line 111 of this file) sales characters are all removed and recreated anew from the submitted data (a thing which happens a lot for ancillary objects created via clone form around LK, as the alternative is.... much more complex and prone to error if not implemented right).

A way around this is to add a (usually hidden, though in this case you could make a case for having it available for editing, since image IDs are exposed by default) field to the clone form and populate it with the image ID for existing sales characters, and then have something like `$charData[$key]['image_id'] ?? $character->image->id` here, thus preserving the already set ID.

itinerare (changes requested):

SpeedyD:
...I have nothing to add, I think Merc got everything that needed to be said. :sweat_smile:

itinerare (approved):

SpeedyD (approved):

resources/views/admin/sales/_character_select_entry.blade.php
```diff
@@ -7,6 +7,7 @@
                     <div class="character-image-blank hide">Enter character code.</div>
                     <div class="character-image-loaded">
                         @include('home._character', ['character' => $character->character])
+                        {!! Form::hidden('image_id', $character->image_id) !!}
```
itinerare, line 10: Wait, just realized that this needs to be `image_id[]` since it's in a clone form...

app/Services/SalesService.php
```diff
@@ -232,6 +232,7 @@ private function processCharacters($sales, $data) {
             // Record data/attach the character to the sales post
             SalesCharacter::create([
                 'character_id' => $character->id,
+                'image_id'     => $data['image_id'] ?? $character->image->id,
```
itinerare, line 235: And this needs to be `$data[$key]['image_id']` to suit.

itinerare (changes requested):

itinerare:
It gets us all sooner or later. |D

PR #631, opened 2023-06-19
Feat(Species Trait Index): Adds modals when clicking on traits

AW0005:
Adds the ability to have modals pop up when you click on a trait on the trait index with the traits description instead of navigating to a separate page.

### DIFF
--- a/resources/views/world/species_features.blade.php
+++ b/resources/views/world/species_features.blade.php
@@ -11,7 +11,7 @@
     <p>This is a visual index of all {!! $species->displayName !!}-specific traits. Click a trait to view more info on it!</p>
 
     @foreach ($features as $categoryId => $categoryFeatures)
-        @if ($categories[$categoryId]->is_visible)
+        @if (!isset($categories[$categoryId]) || $categories[$categoryId]->is_visible)
             <div class="card mb-3 inventory-category">
                 <h5 class="card-header inventory-header">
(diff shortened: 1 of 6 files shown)


resources/views/world/species_features.blade.php
```diff
@@ -11,7 +11,7 @@
     <p>This is a visual index of all {!! $species->displayName !!}-specific traits. Click a trait to view more info on it!</p>
 
     @foreach ($features as $categoryId => $categoryFeatures)
-        @if ($categories[$categoryId]->is_visible)
+        @if (!isset($categories[$categoryId]) || $categories[$categoryId]->is_visible)
```
SpeedyD, line 14: Apologies, but I don't understand this change? Could you explain this?

resources/views/world/species_features.blade.php
```diff
@@ -41,3 +41,17 @@
         @endif
     @endforeach
 @endsection
+
+@section('scripts')
+@if(Config::get('lorekeeper.extensions.species_trait_index.trait_modals'))
```
SpeedyD, line 46: Very, very minor note, but could you perhaps tab the @if and inside the @if section? ^^;;

resources/views/world/species_features.blade.php
AW0005: It's possible for traits to not have any category (and they're labeled with the "Miscellaneous" label on the lines just below this)

SpeedyD: ...so now I have two questions: How did I miss this when I added the is_visible check and how did it go through after a collective check XD

Thanks for fixing this!

SpeedyD (approved):

itinerare (approved):

PR #644, opened 2023-07-13
Update AuthServiceProvider.php

Ne-wt:
this file doesnt exist, we should just comment this out since it floods error logs otherwise


itinerare (approved):

preimpression (approved):

itinerare:
Turns out just catching develop up with main was also a good idea...

PR #668, opened 2023-08-15
add utility reward string creation

ScuffedNewt:
not implemented anywhere but allows assetArrays to be created as a reward string, using it in pets and seems like something vanilla could use to reduce code repetition


AW0005:
Could this possibly be used up front to replace the getBoxRewardString in [BoxService](https://github.com/corowne/lorekeeper/blob/678b9a0c7b5aead130e65cd9a95302753d6c7af2/app/Services/Item/BoxService.php#L160)?
(And should this be set to be merged into develop rather than main?)

itinerare:
Yeah I'd rather either see 1. an implemented use case in core, or 2. a noted use case in another ext before merging it in. And, yeah, this ought to go into develop.

Ne-wt:
> Could this possibly be used up front to replace the getBoxRewardString in [BoxService](https://github.com/corowne/lorekeeper/blob/678b9a0c7b5aead130e65cd9a95302753d6c7af2/app/Services/Item/BoxService.php#L160)? (And should this be set to be merged into develop rather than main?)

yeah that would be the intention

Ne-wt:
> Yeah I'd rather either see 1. an implemented use case in core, or 2. a noted use case in another ext before merging it in. And, yeah, this ought to go into develop.

image
image
image

itinerare:
Aah, I see. Yeah that could be used for boxes pretty readily...

Ne-wt:
> Aah, I see. Yeah that could be used for boxes pretty readily...

it'd probably need a consideration for raffles but that'd be fairly easily added (just snipe it from boxreward)

itinerare:
Yeah I don't imagine that'd be too difficult to manage.

itinerare (approved):

AW0005 (approved):

PR #669, opened 2023-08-15
New Feature: Adds a page that shows the Full Inventory

SpeedyD:
It's purely in Home, as it's for a user to see, in one easier glance, exactly which items they have where.

It's similar to Account Search, but instead of making you look through items you may not even have, it simply shows everything you do have.

### DIFF
--- a/app/Http/Controllers/Users/InventoryController.php
+++ b/app/Http/Controllers/Users/InventoryController.php
@@ -248,7 +248,7 @@ class InventoryController extends Controller {
 
         return view('home.inventory_full', [
             'categories' => $categories->keyBy('id'),
-            'items'      => $items->sortBy('name')->sortBy('updated_at')->groupBy(['item_category_id', 'id']),
+            'items'     => $items->sortBy('name')->sortBy('updated_at')->groupBy(['item_category_id', 'id']),
             'user'       => $user,
             'characters' => $characters,
         ]);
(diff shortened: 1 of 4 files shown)


itinerare (approved):

app/Http/Controllers/Users/InventoryController.php
```diff
@@ -198,6 +198,62 @@ public function getAccountSearch(Request $request) {
...
+    /**
+     * Show the full inventory page.
+     *
+     * @return \Illuminate\Contracts\Support\Renderable
+     */
+    public function getFullInventory() {
+        $user = Auth::user();
+
+        // Gather the user's characters and the items they own
+        $characters = Character::where('user_id', $user->id)->orderBy('slug', 'ASC')->get();
```
AW0005, line 210: Why not use $user->characters or $user->allCharacters? And should this filter out hidden characters?

AW0005 (approved):
Had one comment but overall looks good!

Also I kind of like this inventory view style better. :0
Wonder if it might be nice to factor it out as something that could be toggled on for the regular inventory?

app/Http/Controllers/Users/InventoryController.php
itinerare: Good point-- I *do* think it should filter out hidden chars.

SpeedyD: I based this off of the (Home) Inventory, Character Inventory and Account Search.. that is why.

I'll look into $user->allCharacters, that might be a better call.

AW0005 (approved):

PR #670, opened 2023-08-15
site event calendar

ScuffedNewt:
for sales/prompts/news


Ne-wt:
image
forgot this in main description

app/Http/Controllers/PageController.php
```diff
@@ -42,4 +42,87 @@ public function getCreditsPage() {
...
+     * Returns all events from news, sales and prompts formatted as json.
+     *
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function getCalendarEvents() {
+        $start_at = request()->get('start');
+        $end_at = request()->get('end');
+
+        // get all prompts within time frame, if hide_before_start is set or hide_after_end is set, filter them out if they are not within the prompt time frame
+        $prompts = \App\Models\Prompt\Prompt::where('is_active', 1)->where('start_at', '>', $start_at)->where('end_at', '<', $end_at)
```
AW0005, line 65: Nit, and add `use App\Models\Prompt\Prompt;` at the top, and same goes for the other Models referenced this way in here.
```suggestion
        $prompts = Prompt::where('is_active', 1)->where('start_at', '>', $start_at)->where('end_at', '<', $end_at)
```

app/Http/Controllers/PageController.php
```diff
@@ -42,4 +42,87 @@ public function getCreditsPage() {
...
+            ->where(function ($query) {
+                $query->where('hide_before_start', 0)->orWhere('hide_after_end', 0)
+                    ->orWhere(function ($query) {
+                        $query->where('hide_before_start', 1)->where('start_at', '<=', now());
+                    })
+                    ->orWhere(function ($query) {
+                        $query->where('hide_after_end', 1)->where('end_at', '>=', now());
+                    });
```
AW0005, line 73: Could this be replaced with `->active()`?

app/Http/Controllers/PageController.php
```diff
@@ -42,4 +42,87 @@ public function getCreditsPage() {
...
+
+        $array = [];
+        foreach ($prompts as $prompt) {
+            $array[] = [
+                'id'    => $prompt->id,
+                'title' => $prompt->name,
+                'url'   => $prompt->url,
+                'start' => $prompt->start_at->toW3cString(),
+                'end'   => $prompt->end_at->toW3cString(),
+                'color' => '#4c96d4',
```
AW0005, line 98: Would be nice if these colors were configurable or tied to a css variable in `lorekeeper.css`

app/Http/Controllers/PageController.php
Ne-wt: cannot due to wanting it within the specified timeframe rather than just those that are active, unless all are grabbed and queried down further later

app/Http/Controllers/PageController.php
Ne-wt: can add it to a config

SpeedyD (approved):
Looking at this again with a fresh mindset half a year later, I think it's decent. It'll need a bit of testing and fine tuning, I'm sure, but it's the develop branch.. we'll get there.

**EDIT:** In approval-hindsight, this may actually need a few updates to 3.1.0 first, won't it..?

SpeedyD:
Dear @ScuffedNewt, may I kindly request that, now that it has been officially _nine months_ since your last update, that you re-merge the develop branch into this? There are conflicts, unfortunately.

SpeedyD (approved):
Just. Re-approving just in case it's needed.

SpeedyD:
@ScuffedNewt The conflicts have once more arose. May I request some time out of your schedule to fix these? :)

ScuffedNewt:
again will do this when more approvals are given, there is no need to do this prior to changes being requested or approvals being given

LostInProgres (commented):
Not _quite_ remembering how PR reviews work, so bear with me-

I noticed that the icons on the buttons to go to a previous/next month  aren't displaying right for me.

Image

Didn't find any issues other than that!

PR #675, opened 2023-08-21
Feat(Inventories): Summarized View

SpeedyD:
- Added Summarized View to Home, User and Character inventories, similarly to the Masterlist view.
- Also fixed minor change on Full Inventory, and added a tag symbol for named items.

This kind of view for the other inventories was originally suggested by @AW0005 in https://github.com/corowne/lorekeeper/pull/669#pullrequestreview-1578925415 :)

### DIFF
--- a/resources/views/home/inventory.blade.php
+++ b/resources/views/home/inventory.blade.php
@@ -18,39 +18,78 @@
 
     <p>This is your inventory. Click on an item to view more details and actions you can perform on it.</p>
 
-    @foreach ($items as $categoryId => $categoryItems)
-        <div class="card mb-3 inventory-category">
-            <h5 class="card-header inventory-header">
-                {!! isset($categories[$categoryId]) ? '<a href="' . $categories[$categoryId]->searchUrl . '">' . $categories[$categoryId]->name . '</a>' : 'Miscellaneous' !!}
-                <a class="small inventory-collapse-toggle collapse-toggle" href="#categoryId_{!! isset($categories[$categoryId]) ? $categories[$categoryId]->id : 'miscellaneous' !!}" data-toggle="collapse">Show</a>
-            </h5>
-            <div class="card-body inventory-body collapse show" id="categoryId_{!! isset($categories[$categoryId]) ? $categories[$categoryId]->id : 'miscellaneous' !!}">
-                @foreach ($categoryItems->chunk(4) as $chunk)
-                    <div class="row mb-3">
-                        @foreach ($chunk as $itemId => $stack)
-                            <div class="col-sm-3 col-6 text-center inventory-item" data-id="{{ $stack->first()->pivot->id }}" data-name="{{ $user->name }}'s {{ $stack->first()->name }}">
-                                @if ($stack->first()->has_image)
-                                    <div class="mb-1">
-                                        <a href="#" class="inventory-stack">
-                                            <img src="{{ $stack->first()->imageUrl }}" alt="{{ $stack->first()->name }}" />
-                                        </a>
-                                    </div>
-                                @endif
-                                <div>
-                                    <a href="#" class="inventory-stack inventory-stack-name">{{ $stack->first()->name }} x{{ $stack->sum('pivot.count') }}</a>
-                                </div>
-                            </div>
-                        @endforeach
-                    </div>
-                @endforeach
-            </div>
+    <div class="text-right mb-3">
+        <div class="btn-group">
+            <button type="button" class="btn btn-secondary active def-view-button" data-toggle="tooltip" title="Default View" alt="Default View"><i class="fas fa-th"></i></button>
+            <button type="button" class="btn btn-secondary sum-view-button" data-toggle="tooltip" title="Summarized View" alt="Summarized View"><i class="fas fa-bars"></i></button>
         </div>
-    @endforeach
+    </div>
+
+    <div id="defView" class="hide">
+        @foreach ($items as $categoryId => $categoryItems)
+            <div class="card mb-3 inventory-category">
+                <h5 class="card-header inventory-header">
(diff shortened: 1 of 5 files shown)


SpeedyD:
My apologies for all the extra pushes, I'm trying to clean the code a little bit and.. finding out the blade formatter has other opinions that I wasn't aware of. :)

(Both formatters have run at this point, with no issues, so please review to your heart's content.)

AW0005 (approved):

resources/views/home/inventory.blade.php
```diff
@@ -18,39 +18,78 @@
...
+
+    <div id="defView" class="hide">
+        @foreach ($items as $categoryId => $categoryItems)
+            <div class="card mb-3 inventory-category">
+                <h5 class="card-header inventory-header">
+                    {!! isset($categories[$categoryId]) ? '<a href="' . $categories[$categoryId]->searchUrl . '">' . $categories[$categoryId]->name . '</a>' : 'Miscellaneous' !!}
+                    <a class="small inventory-collapse-toggle collapse-toggle" href="#categoryId_{!! isset($categories[$categoryId]) ? $categories[$categoryId]->id : 'miscellaneous' !!}" data-toggle="collapse">Show</a>
+                </h5>
+                <div class="card-body inventory-body collapse show" id="categoryId_{!! isset($categories[$categoryId]) ? $categories[$categoryId]->id : 'miscellaneous' !!}">
+                    @foreach ($categoryItems->chunk(4) as $chunk)
```
Ne-wt, line 36: there is a lot of repetition outside of this chunk, consider reducing code reuse and instead having hidden widgets outside of the boilerplate that are cloned or hidden/unhidden from within
you can set classes of the parents similarly

resources/views/home/inventory_full.blade.php
```diff
@@ -20,17 +20,23 @@
...
-                @foreach ($categoryItems as $itemId => $itemtype)
+                @foreach ($categoryItems as $itemtype)
                     <div class="col-lg-3 col-sm-4 col-12">
-                        @if ($categoryItems[$itemId]->first()->has_image)
-                            <img src="{{ $categoryItems[$itemId]->first()->imageUrl }}" style="height: 25px;" alt="{{ $categoryItems[$itemId]->first()->name }}" />
+                        @if ($itemtype->first()->has_image)
+                            <img src="{{ $itemtype->first()->imageUrl }}" style="height: 25px;" alt="{{ $itemtype->first()->name }}" />
                         @endif
-                        <a href="{{ $categoryItems[$itemId]->first()->idUrl }}">{{ $categoryItems[$itemId]->first()->name }}</a>
+                        <a href="{{ $itemtype->first()->idUrl }}">{{ $itemtype->first()->name }}</a>
```
Ne-wt, line 28: this is all practically the same code, on a similar vain, consider moving the whole thing in a widget

resources/views/user/inventory.blade.php
```diff
@@ -11,31 +11,79 @@
...
-                        @foreach ($chunk as $itemId => $stack)
-                            <div class="col-sm-3 col-6 text-center inventory-item" data-id="{{ $stack->first()->pivot->id }}" data-name="{{ $user->name }}'s {{ $stack->first()->name }}">
-                                <div class="mb-1">
-                                    <a href="#" class="inventory-stack"><img src="{{ $stack->first()->imageUrl }}" alt="{{ $stack->first()->name }}" /></a>
-                                </div>
-                                <div>
-                                    <a href="#" class="inventory-stack inventory-stack-name">{{ $stack->first()->name }} x{{ $stack->sum('pivot.count') }}</a>
+    <div class="text-right mb-3">
+        <div class="btn-group">
+            <button type="button" class="btn btn-secondary active def-view-button" data-toggle="tooltip" title="Default View" alt="Default View"><i class="fas fa-th"></i></button>
```
Ne-wt, line 16: again

resources/views/widgets/_inventory_view_js.blade.php
```diff
@@ -0,0 +1,46 @@
+<script>
+    $(document).ready(function() {
+
+        var $defButton = $('.def-view-button');
```
Ne-wt, line 4: use let instead of var, or better yet for unchanging like ids use const

resources/views/widgets/_inventory_view_js.blade.php
```diff
@@ -0,0 +1,46 @@
...
+        var $defView = $('#defView');
+        var $sumButton = $('.sum-view-button');
+        var $sumView = $('#sumView');
+
+        var view = null;
```
Ne-wt, line 9: dont initialise to null since youre calling initView immediately anyway

resources/views/widgets/_inventory_view_js.blade.php
```diff
@@ -0,0 +1,46 @@
...
+
+        function setView(status) {
+            view = status;
+
+            if (view == 'def') {
+                $defView.removeClass('hide');
+                $defButton.addClass('active');
+                $sumView.addClass('hide');
+                $sumButton.removeClass('active');
+                window.localStorage.setItem('lorekeeper_inventory_view', 'def');
```
Ne-wt, line 36: nitpick but use status / view instead of strings since youre going to the trouble of redefining

SpeedyD:
The only thing that repeats would be between the User Inventory in Home and the User Inventory in Profile, which use identical codes. The rest, not quite as identical, need more distinct solutions. There are big differences between the User Inventory and Character Inventory alone, for they use different item types. The Full Inventory is coded different once again as applies to both of these item types.

As for the Inventory views, I opted to be consistent with the Masterlist views. Most of this code is taken from there and altered with different variables so they don't interfere with each other. My suggestion would be to improve that first, and then look at where you can improve this code as well.

itinerare (approved):

PR #676, opened 2023-08-22
Bug(Shops): Shops with no image will show broken image icon

AW0005:
Shops with no image specified were still adding the image html and showing a broken image icon instead of hiding it.


itinerare (approved):

SpeedyD (approved):
No brainer change- I'm amazed nobody (including myself) caught this sooner. O_o;

PR #703, opened 2023-10-15
Fix(MYOs): Allows character images to work locally

SpeedyD:
Having found that you cannot make MYO slots because of an incorrect url issue, I investigated how this happened and found it was this use of the asset() function as opposed to the public_path() function.
I've done my research and there is not a single service that uses asset(), except the character manager. So I hereby propose this fix.

This admittedly only fixes the use of MYO slots on localhost. On a live site, this change has no effect. But this should make testing easier, at least.


AW0005 (approved):

itinerare (approved):

itinerare:
Holding for merge.

PR #741, opened 2023-12-11
Feature/wysiwyg comments

ScuffedNewt:


app/Http/Controllers/Comments/CommentController.php
```diff
@@ -89,7 +89,7 @@ public function store(Request $request, $model, $id) {
         }
 
         $comment->commentable()->associate($base);
-        $comment->comment = $request->message;
+        $comment->comment = config('lorekeeper.settings.wysiwyg_comments') ? $request->message : parse($request->message);
```
itinerare, line 92: This needs to be run through HTML purifier if WYSIWYG I think (with any needed changes to edit logging as well).

itinerare (changes requested):
Looks good, though I'd like to see it bolted down a little more to be safe.

app/Http/Controllers/Comments/CommentController.php
Ne-wt: parse is the purifier

Ne-wt: also edit logging has been updated already

itinerare (approved):

AW0005 (approved):

PR #742, opened 2023-12-11
Feature/username changes

ScuffedNewt:
this also adds proper laravel forms to avatar form which should prevent the button from disappearing on mobile (a common complaint)

also adds filetime query to avatar URL which should prevent the need to "hard refresh"


itinerare (approved):

AW0005 (approved):
This looks good! Had the thought while reading through that it might be nice in the future to add like the last 5 username change logs or something like that to the admin-side "Edit User" page for easy reference in case of issues.

itinerare:
They are logged admin-side!

itinerare (approved):

PR #743, opened 2023-12-11
Notify user of no reward

Ne-wt:
common complaint
it doesnt add danger flash, happy to move flashes into the boxrewardstring if needed


Ne-wt:
image

itinerare:
:(

AW0005 (approved):

PR #748, opened 2023-12-14
feature/general tidying

itinerare:
This one touches a lot of files and is consequently a bit of a monster, but it should be exclusively refactors and thus fairly safe across the board. It's essentially the sum of a few clean-up/refactoring tasks that I've been meaning to do for a bit and which we might as well get out of the way.

### DIFF
--- a/app/Models/User/UserAlias.php
+++ b/app/Models/User/UserAlias.php
@@ -32,7 +32,7 @@ class UserAlias extends Model {
      * Get the user this set of settings belongs to.
      */
     public function user() {
-        return $this->belongsTo('App\Models\User\User', 'user_id');
+        return $this->belongsTo(User::class, 'user_id');
     }
 
     /**********************************************************************************************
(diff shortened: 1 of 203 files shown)


SpeedyD (commented):
I'm at about 90 files in now, but it's midnight and I caught myself reading over a few things, so I'm quitting my deep-check for the night and will resume when I wake up.

Want to add though: 90 files in and no problems so far. :)

app/Models/User/UserAlias.php
```diff
@@ -3,7 +3,7 @@
 namespace App\Models\User;
 
 use App\Models\Model;
-use Config;
+use Illuminate\Support\Facades\Config;
```
AW0005, line 6: Nit: Should this have been removed?

AW0005 (approved):
Just one little thing I noticed but everything else looked great. Hyped for these changes!

app/Models/User/UserAlias.php
itinerare: Yes-- the linter doesn't seem to catch all of them when removing unused imports, for whatever reason, but I'm not overly worried about defunct ones since it's more a tidiness issue than a functionality one.

SpeedyD (approved):
203/203 files viewed. I see no issue in the changes. :)

..Barring the conflicts mentioned by Github.

PR #749, opened 2023-12-15
fix(images): Add hash and htaccess rule to prevent hidden images from being discoverable

AW0005:
Reference here for the funky looking htaccess rule:
http://tltech.com/info/referrer-htaccess/

I have successfully tested it with an apache based server.

composer update change should fix this formatter failing that I was also running into:
https://github.com/corowne/lorekeeper/actions/runs/7215907881


itinerare:
Do the htaccess changes interfere with images displaying in meta tags (e.g. in discord previews)?

AW0005:
Oh yes, it absolutely will... ah the first time I did this we opted to at least white list character images for that and I can probably do that here, but it would still block other embeds. 

Whitelisting domains that it could embed on is also difficult - discord alone does the pulling of the image behind the scenes and then reserves on their own url so no good way to know what the referrer would be for that. 

Hard problem, 100% open to ideas / adjusting what's put into core. v.v

itinerare:
I feel like the htaccess changes might be better kept out of core and put forward as an additional option with the caveat that it is a more intensive solution/not one-size-fits-all... also given that it already isn't given apache2 vs nginx etc etc.

SpeedyD:
Out of curiosity, is there a way to make this work retroactively? Like, especially for sites that already have hundreds.. or hundreds of thousands of images up?

Aside from that- and the fact that I agree with Merc above- I think it's approvable :)

itinerare:
Yeah I think it'd be prudent to add a command to do that.

AW0005:
Reasonable on the htaccess I'll edit this soon to remove!

this code is actually backwards compatible with existing images! If there's no hash it still returns the image url with just the id. Given that, do you still think the command is warranted?

itinerare:
Yes-- even if it doesn't break existing images, folk are liable to want to add hashes to 'em.

app/Console/Commands/AddImageHashes.php
```diff
@@ -0,0 +1,79 @@
...
+    $images = $images->concat(PromptCategory::where('has_image', 1)->whereNull('hash')->get());
+    $images = $images->concat(Rarity::where('has_image', 1)->whereNull('hash')->get());
+    $images = $images->concat(Shop::where('has_image', 1)->whereNull('hash')->get());
+    $images = $images->concat(Species::where('has_image', 1)->whereNull('hash')->get());
+    $images = $images->concat(Subtype::where('has_image', 1)->whereNull('hash')->get());
+
+    foreach ($images as $image) {
+      $oldName = $image->id . '-image.png';
+      $image->hash = randomString(10);
+      // Any service works, I can't use the abstract one
```
itinerare: Yeah this is fine/as expected.

itinerare (approved):

SpeedyD (approved):

PR #755, opened 2023-12-18
make text form disappear on galleries if images are uploaded

ScuffedNewt:
recently experienced some confusion with users thinking text = title so added some clarification and hide the form if an image is uploaded

### DIFF
--- a/resources/views/galleries/create_edit_submission.blade.php
+++ b/resources/views/galleries/create_edit_submission.blade.php
@@ -37,7 +37,7 @@
         <h2>Main Content</h2>
         <p>Upload an image and/or text as the content of your submission. You <strong>can</strong> upload both in the event that you have an image with accompanying text or vice versa.</p>
 
-        <div class="form-group">
+        <div class="form-group" id="imageForm">
             {!! Form::label('Image') !!}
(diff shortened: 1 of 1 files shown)


itinerare:
While I understand the desire to mitigate confusion here, that both image and text input are available at once is deliberate, e.g. if a user wants to add a supplementary image for a primarily text submission (or similar). However, I *do* support the clearer labeling for the text field.

resources/views/galleries/create_edit_submission.blade.php
```diff
@@ -57,8 +57,8 @@
             </div>
         </div>
 
-        <div class="form-group">
-            {!! Form::label('Text') !!}
+        <div class="form-group" id="writingForm">
+            {!! Form::label('Writing / Text') !!} {!! add_help('If you have a text submission, you can paste it here. You can also use the WYSIWYG editor to format your text. If you have an image submission, you can leave this blank.') !!}
```
SpeedyD, line 61: I like this text, but I'd like to also suggest adding to the final sentence, like.. "-you can leave this blank, or add a text to supplement your image submission."

PR #758, opened 2023-12-24
chore(workflows): consolidate linting, use reusable workflows

itinerare:
I was cleaning up, updating, and de-duplicating workflows for my own projects, so figured I might as well offer up similar here.
The main changes are:
- The pint and blade-formatter workflows are condensed into one "lint" workflow with two different jobs (on its own just a more compact rendering of the existing implementation)
- The workflows themselves are provided for by copies on a repo I made to house reusable workflows... the workflows themselves can be found [here](https://github.com/itinerare/github-actions/blob/main/.github/workflows/pint.yml) and [here](https://github.com/itinerare/github-actions/blob/main/.github/workflows/blade_formatter.yml); I figure this isn't functionally much different from calling external actions. They are much the same as the existing workflows, just updated (and easier for me to maintain on that front).
- The big functionality difference is the addition of concurrency group settings, which compels the jobs to queue nicely rather than run simultaneously and then clash if both have something they want to commit, resolving that particular issue.

It also adjusts the workflow to run only on PRs, though on most events associated with them. This sidesteps some annoying issues (since it allows setting the workflows to run on the actual PR branch rather than just the detached head as they would otherwise do, to my understanding) while arriving at functionally the same result, since the vast majority of changes pass through a PR first.


SpeedyD (approved):
From what I understand, this basically goes from running every commit to running every PR, using an external workflow- in your public repo 'github-actions'.

...Seems fine to me, I disliked the waiting for jobs over and over.

itinerare:
It *does* re-run on changes to a PR, though, since that may be fairly important depending on the circumstances.

SpeedyD:
I see no issue with that. If you make a change to a PR, that may require a secondary check. 🤷

itinerare:
Precisely, yeah.

itinerare:
Realized it should really run on PRs to *any* branch, since hotfixes *are* sometimes a thing.

SpeedyD (approved):

AW0005 (approved):

PR #760, opened 2023-12-25
Fix/command extension tracker backslashes

SpeedyD:
Linux servers do __not__ like back slashes. (**Boo to `\`!!**)

Linux servers **do** like forward slashes. (**Yay to `/`!!**)

TL;DR No matter how nice it looks locally, it won't run live unless forwardslash.


SpeedyD:
WRONG BRANCH IGNORE ME

PR #762, opened 2023-12-25
fix(user): sidebar set_active fix

SpeedyD:
For some reason, set_active does NOT play nice with just using $user->url..

This fixes it by pulling the consistent code from the other section.

Also.. item-logs was currency-logs?? Whoops.


itinerare (approved):

SpeedyD:
set_active actually expects a _relative path_. $user->url gives the entire url. That is why it doesn't work with $user->url.

AW0005 (approved):

PR #763, opened 2023-12-25
fix(user): Forced lowercase on name check

SpeedyD:
Imagine: you know a username, and you visit their profile by just typing their name.

But you forgot to capitalize their name. Now a box appears that their username was changed to "Username", just cause you missed the capitalization.

..Very minor bug, yes, but by forcing the check to be entirely lowercase, it eliminates the messagebox from popping up on missing caps.


itinerare (approved):

Ne-wt (approved):
silly oversights like these am i right

itinerare:
I mean, we all do it... nothing new there

PR #764, opened 2023-12-25
fix(galleries): Shows subgalleries of parent while in subgallery (and siblings and avunculi!)

SpeedyD:
While in a subgallery, the sidebar stopped showing the other subgalleries of it's parent.

This fixes it, and even works for sub-subgalleries.. Though only one level in.

~~This may require more finetuning in the model later, but I figured this is a fine start.~~

I actually added some fixes that makes it instead check for siblings and avunculi. (Aunts and uncles - aka the parent's siblings.)

You can now have up to Three levels in- which marks a vast improvement in my opinion.

### DIFF
--- a/app/Models/Gallery/Gallery.php
+++ b/app/Models/Gallery/Gallery.php
@@ -78,6 +78,26 @@ class Gallery extends Model
     }
 
     /**
+     * Get the sibling galleries of this gallery.
+     */
+    public function siblings() {
+		if($this->parent) {
+			return $this->parent->hasMany(self::class, 'parent_id')->sort();
+		}
+		return null;
+    }
+	
+    /**
+     * Get the avunculi galleries of this gallery.
+     */
+    public function avunculi() {
+		if ($this->parent && $this->parent->siblings()) {
+			return $this->parent->siblings()->sort();
+		}
+		return null;
+    }
+
+    /**
      * Get the submissions made to this gallery.
      */
     public function submissions()
--- a/resources/views/galleries/_sidebar.blade.php
+++ b/resources/views/galleries/_sidebar.blade.php
@@ -19,6 +19,15 @@
         </li>
     @endif
 
+    @if ($galleryPage && $sideGallery->parent && $sideGallery->parent->children->count())
+        <li class="sidebar-section">
+            <div class="sidebar-section-header">{{ $sideGallery->parent->name }}: Sub-Galleries</div>
+            @foreach ($sideGallery->parent->children()->visible()->get() as $child)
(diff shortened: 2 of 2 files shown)


itinerare (approved):

app/Models/Gallery/Gallery.php
```diff
@@ -77,6 +77,26 @@ public function children()
...
+		if($this->parent) {
+			return $this->parent->hasMany(self::class, 'parent_id')->sort();
+		}
+		return null;
+    }
+	
+    /**
+     * Get the avunculi galleries of this gallery.
+     */
+    public function avunculi() {
```
Ne-wt, line 93: I wonder if this could possibly be done recursively to get parent->grandparent->so and so forth then recurse those for children... not for this PR BUT I / any one else who is interested in it may be able to put some thinking to it

itinerare: It's possible; I've done recursion along those lines for some other stuff... though I guess it depends on how much utility something like that provides relative to the amount of effort? But open to hearing if you/someone else comes up with something neat in that vein.

SpeedyD: I've been considering it, but.. I figured three generations should be enough at this point. You can always go back to the last generation, and from there the last, if needed..

I'm also wondering who is really going to be crazy enough to make a Lorekeeper with sub-sub-sub-sub-subgalleries.. At some point, there needs to be a line drawn, don't you think so?

itinerare: Yeah, it's a fun little puzzle to consider, but I think for practical purposes this PR goes about as far as is necessary (arguably even a little further, so really, the bases should be covered here).

SpeedyD: My brain went brr. Here's code I just typed up.
```php
public function checkParents(object $childGallery, array $familyTree = []) {
  if ($childGallery->parent) {
    $parent = $childGallery->parent;
    $familyTree[$parent->name] = include $parent;
    checkParents($parent, $familyTree);
  }
  return $familyTree;
}
```
I'm not implementing this nor have I tested this and take no responsibility for this working or failing. :P

SpeedyD: > Yeah, it's a fun little puzzle to consider, but I think for practical purposes this PR goes about as far as is necessary (arguably even a little further, so really, the bases should be covered here).

And I forgot to comment on this specifically but I absolutely agree. Code I posted above was literally just a puzzle I tried to solve cause brain brr, but not cause of an actual need. 🤷

AW0005 (approved):

PR #771, opened 2023-12-31
Datetimepicker consistency fix, widget creation, localization and nitpicks

SpeedyD:
- Added Month and Year selection
- Made 'datepicker' consistenty be the class, as there are sometimes multiple pickers on a page.
- Made the 'datepicker' function a widget.
- Added localization, clicking Now will return site time rather than the time of the PC used.

- For some reason, the cooldown datepicker on a character's stats modal did not work, would not work, whatever- so I have done a lot of specific fixes just for that one.

This has been thoroughly tested and seems to be working fine now.


SpeedyD:
🤦
I make the filename consistent in code and then forget to change the actual filename X_x

SpeedyD:
.... don't approve just yet. Realized a stupid issue. I should've made it CLASS consistent, cause I realize now that I forgot to test when there's multiple datetimepickers, and IDs block that.

Be back after extensive tests.

SpeedyD:
..for the record, I do NOT understand what the blade formatter is doing, but whatever

SpeedyD:
It has been thoroughly checked - and I'm.. exceedingly confused why the stats modal's datepicker didn't work.

I _got_ it to work, but why DIDN'T it??

itinerare (approved):

SpeedyD:
...oh yeah, with my other PR just pushed through, this one doesn't need that script section at all! Yay!

AW0005:
https://github.com/SpeedyD/lorekeeper/pull/213

Little update coming through to keep the datepicker floaty while using it's "inline" setting

SpeedyD:
Absolutely, love that, thanks @AW0005!

SpeedyD:
@itinerare why is the formatter doing that weird spacing to one of the js lines? I purposely reverted it to the last part being part of the previous line. . . 

And NOW the final change is that it put it back on ONE line??? I do not understand the blade formatter / lint / whatever's rules.

itinerare:
Would that I could help you, but the blade formatter is not any less of a mystery to me sometimes...

Ne-wt (approved):

PR #779, opened 2024-01-02
Make back-hashing images optional

AW0005:
While I agreed with providing this command for those that wish to back-hash their image urls, I don't think it's a good idea to assume everyone updating to V3 will inherently want to do this - I certainly wouldn't.

Most existing images likely won't be unreleased content, _and_ it's not uncommon to use the static link to an image in other places of the site, via pages, news, sales, etc. Back-hashing will kill all of those links, which very much may not be expected or remotely wanted by those updating to V3.

So this opts to add that information as a part of the V3 update command and give the option to skip back-hashing images


itinerare:
That's a fine and fair way about it!

itinerare (approved):

SpeedyD (approved):

PR #781, opened 2024-01-03
suggestion(character): Sale value switch gets removed

SpeedyD:
I don't understand the reasoning behind it becoming impossible to change the sale value when the design is set to unsellable- and then keeping the sale value visible nonetheless.

Some sites prefer to assign values regardless of it being sellable, even if just to attribute to art value for an art trade., so I get why you'd want to KEEP the sale value even if a design can't be sold.. but why hide the input for the sale value..???

It's just awkward.

And as sidenote: Made it so that if the sale value is not higher than 0, it will just hide the sale value from the character.


itinerare (approved):

AW0005 (approved):

SpeedyD:
I have NO idea why it said merge conflicts, but I fixed them just now

PR #783, opened 2024-01-04
Suggestion: species visual index fix

SpeedyD:
Clicking either the image, name or rarity returns the modal.. highly annoying!

Even if you just want to copy paste the name, or want to get the rarity specifically, or maybe you want the link to the trait itself- any of these would make the modal pop up!

..Not anymore. This fixes it so only the IMAGE returns the modal, and the name and rarity work like the links they always were.


itinerare:
Makes enough sense to me.

itinerare (approved):

Ne-wt (approved):

PR #790, opened 2024-01-09
Allow for removal of last power on a rank

AW0005:
Discovered today that if you add a power to a rank, and it's the only power, and you try to remove it, it doesn't remove, because the code that deletes the old powers was only happening if at least one other power was being specified.


itinerare (approved):
Someday we'll manage to squash all of this kind of bug...

PR #792, opened 2024-01-11
feature/password reset fix

itinerare:
As on the tin. Fixes an error getting the password reset token from the route and adds flash messages back to the password reset flow. Note that the one for _requesting_ a password reset is very deliberately obtuse for security reasons/so as not to directly confirm/deny whether or not a given email address exists in the system (and since tanking UX a little for this sake is an acceptable sacrifice; it's not liable to be a high-traffic area).


resources/views/auth/passwords/reset.blade.php
Ne-wt, line 1: just adding the same route thing to email instead of $email

itinerare: Yeah I mostly just wanted to get your response on that before doing it, but agree

PR #794, opened 2024-01-14
Universal trait index merged into core

SpeedyD:
@CH3RVB's Universal Trait Index extension - but now merged into develop.
Additionally, it contains the same modal system the Species Trait Index has, for consistency, as well as the same options to disable it.. And of course, does not display hidden traits.

### DIFF
--- a/app/Models/Feature/Feature.php
+++ b/app/Models/Feature/Feature.php
@@ -93,8 +93,7 @@ class Feature extends Model {
      *
      * @return string
      */
-    public function getVisualTraitsUrlAttribute()
-    {
+    public function getVisualTraitsUrlAttribute() {
         return url('/world/universaltraits');
     }
 
--- a/app/Models/Feature/Feature.php
+++ b/app/Models/Feature/Feature.php
@@ -88,15 +88,6 @@ class Feature extends Model {
         return $this->belongsTo(FeatureCategory::class, 'feature_category_id');
     }
 
-    /**
-     * Gets the URL the visual index of universal traits.
-     *
-     * @return string
-     */
-    public function getVisualTraitsUrlAttribute() {
-        return url('/world/universaltraits');
-    }
-
     /**********************************************************************************************
 
         SCOPES
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -300,14 +300,12 @@ class WorldController extends Controller {
         ]);
     }
 
-     /**
+    /**
      * Shows a universal visual trait list.
      *
-     * @param  \Illuminate\Http\Request  $request
      * @return \Illuminate\Contracts\Support\Renderable
      */
-    public function getUniversalFeatures(Request $request)
-    {
+    public function getUniversalFeatures(Request $request) {
         $categories = FeatureCategory::orderBy('sort', 'DESC')->get();
         $rarities = Rarity::orderBy('sort', 'ASC')->get();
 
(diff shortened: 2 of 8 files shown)


SpeedyD:
"But Speedy", I hear you say. "Why is this a draft?"

Frankly, the Species Trait Index has subtype features in there that the Universal Trait Index does not. Though I asked @CH3RVB, and they were unaware of leaving them out, they also believed it weird that a universal trait would actually be restricted to a subtype.

ERGO! Before I make this a proper pull, I'd like to know opinions on this. :)

itinerare:
I mean I don't think a trait can be associated with a subtype without being associated with a species, so it should by default be impossible to include in this...

SpeedyD:
> I mean I don't think a trait can be associated with a subtype without being associated with a species, so it should by default be impossible to include in this...

...Now that you mention it, a subtype **must** be associated with a species.. But a trait can be associated with a subtype **without** being associated with a species.

..Perhaps a bugfix we may need to do?

app/Models/Feature/Feature.php
```diff
@@ -88,6 +88,15 @@ public function category() {
         return $this->belongsTo(FeatureCategory::class, 'feature_category_id');
     }
 
+    /**
+     * Gets the URL the visual index of universal traits.
+     *
+     * @return string
+     */
+    public function getVisualTraitsUrlAttribute() {
+        return url('/world/universaltraits');
```
itinerare, line 97: From at least a quick glance... is this used anywhere? If not, it should probably be removed...

itinerare:
Probably it should only allow selecting a subtype if a species is selected, consistent with other places...

app/Models/Feature/Feature.php
SpeedyD: ...Honestly, I have no idea? That's from @CH3RVB's original code. Lemme look around.

SpeedyD: Aight, I think I see what happened. I guess this was assumed necessary because the Species Trait Index needs it.. But nothing uses this. Removing.

itinerare: Ah, yeah, that tracks.

SpeedyD:
> Probably it should only allow selecting a subtype if a species is selected (and only subtypes of that species), consistent with other places...

Agreed. I'll see if I can get that done as bugfix PR for Release/v3.0.0 - for now, I'll keep the subtype out, get my update in and put this on ready for review.

itinerare:
Sounds good to me!

itinerare (approved):

itinerare (approved):

SpeedyD:
It's part of the Species Visual Trait Index, too.. I haven't written this
code.

Either way, there IS a difference- To explain, it's a ternary operator that
starts the line before: `$features = count($categories) ?`
Then at the end of the first part it has an `:` instead of a `;`, allowing
the "if false" result.

And as distinct difference, the first has
`->orderByRaw('FIELD(feature_category_id,'.implode(',',
$categories->pluck('id')->toArray()).')')`, but the second does not.

So long story short, the orderByRaw can ONLY trigger if $categories has a
count, and then IN the orderByRaw it's id gets plucked..

(What I personally don't get is why it's groupedBy feature_category_id even
if there's none.. But as I said, not my code. Technically @itinerare's, I
think?)

AW0005 (approved):
Oh yea my brain totally missed the ternary operator before!

itinerare:
There is to be fair probably a much cleaner way of writing it... but I'm not especially concerned about it right at this very moment.

PR #805, opened 2024-01-17
Extension/multiple subtypes

Ne-wt:
by nature is "optional", recommended by @SpeedyD to PR, open to discussion on whether this should be vanilla or not but probably best had on the PR anyway

### DIFF
--- a/resources/views/admin/masterlist/create_character.blade.php
+++ b/resources/views/admin/masterlist/create_character.blade.php
@@ -216,8 +216,8 @@
     </div>
 
     <div class="form-group" id="subtypes">
-        {!! Form::label('Subtype (Optional)') !!} @if($isMyo) {!! add_help('This will lock the slot into a particular subtype. Leave it blank if you would like to give the user a choice, or not select a subtype. The subtype must match the species selected above, and if no species is specified, the subtype will not be applied.') !!} @endif
-        {!! Form::select('subtype_id', $subtypes, old('subtype_id'), ['class' => 'form-control disabled', 'id' => 'subtype']) !!}
+        {!! Form::label('Subtypes (Optional)') !!} @if($isMyo) {!! add_help('This will lock the slot into a particular subtype. Leave it blank if you would like to give the user a choice, or not select a subtype. The subtype must match the species selected above, and if no species is specified, the subtype will not be applied.') !!} @endif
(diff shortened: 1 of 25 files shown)


Ne-wt:
will update base from main at a later point, just want to have code review of the update right now + aforementioned discussion

resources/views/admin/masterlist/create_character.blade.php
```diff
@@ -259,7 +259,16 @@
       var myo = '<?php echo($isMyo); ?>';
       $.ajax({
         type: "GET", url: "{{ url('admin/masterlist/check-subtype') }}?species="+species+"&myo="+myo, dataType: "text"
-      }).done(function (res) { $("#subtypes").html(res); }).fail(function (jqXHR, textStatus, errorThrown) { alert("AJAX call failed: " + textStatus + ", " + errorThrown); });
+      }).done(function (res) { 
+        $("#subtypes").html(res); 
+        $( "#subtype" ).selectize({
+            maxItems: 10
```
itinerare, line 265: I'll do a more in-depth code review at some later point, but for the sake of discussion... While this is inherently "optional", I wonder if it might be useful to utilize this characteristic of selectize (and probably some additional validation on the back end) to have a configurable cap on the number of subtypes that can be added. Provided that, I think, I wouldn't have an issue merging this in, since it could (by default) provide the same level of functionality as current while leaving the option open/effectively be toggleable.

itinerare:
I don't think supporting exts directly like that is within scope for core... that's the responsibility of ext devs.

Ne-wt:
> I do think there's a few bits missing
yeah I can go ahead and add those in when they're all merged

> I do think it'd be neat to have an automated check whether the Second Subtype ids
if it's merged as a part of main I think it'd be best to be left out a la what merc said, but really the additions aren't *too* hard

> And.. there's also the overlooked Third Subtype ext
should be the same as second subtype, just adding another entry into the CharacterImageSubtype create part

resources/views/admin/masterlist/create_character.blade.php
Ne-wt: I've not put any backend validation (for now) but have added the configurable frontend. Shouldn't be a rush anyway considering this will be waiting on aforementioned PRs by speedy

itinerare: Sounds good, and yeah, really no rush.

PR #813, opened 2024-01-24
Prevent Horizontal Stretch on <main>

spectrology:
The flex direction limit will prevent blocks from stretching the main container since it's set to flex. This is only on #app right now but it should be on main as well.

As for break-word, I think Lorekeeper ships with name length limits, but this prevents cases where users add header blocks to their comments / descriptions / other stuff it'd be odd to impose a limit on. I understand removing that property but definitely recommend keeping flex-direction.

### DIFF
--- a/public/css/lorekeeper.css
+++ b/public/css/lorekeeper.css
@@ -84,6 +84,7 @@ body {
 main {
     flex-grow: 1;
     display: flex;
+    flex-direction: column;
 }
 
 main > .row {
--- a/public/css/lorekeeper.css
+++ b/public/css/lorekeeper.css
@@ -85,6 +85,7 @@ main {
     flex-grow: 1;
     display: flex;
     flex-direction: column;
+    word-wrap: break-word;
 }
 
 main > .row {


itinerare (approved):
Yeah, validation etc rules can only really cover so many cases intrinsically... so this kind of alternate solution makes a lot of sense, by my reckoning. I think `break-word` works in this case as well/covers a lot of cases that otherwise result in awkwardness (and if it _also_ results in things looking a little funky, it looks at least to me like it's generally going to be the lesser of those evils).

itinerare:
(Thanks also for contributing and for following conventional commit format! 👍)

SpeedyD (approved):
> (and if it _also_ results in things looking a little funky, it looks at least to me like it's generally going to be the lesser of those evils).
Real talk? If it ends up looking a little funky, that will be the responsibility of the site owner to fix- I think with this we've done all we can to make it work best as possible from our end.

itinerare:
Not necessarily... some things just can't be helped (especially depending on device/viewport size/etc), is really the bottom line there.

PR #822, opened 2024-02-05
Fix/hidden traits and unreleased items visible for staff

SpeedyD:
This makes hidden traits and unreleased items visible for staff.

...I swore this was part of the point of the v3 release.

### DIFF
--- a/resources/views/world/species_features.blade.php
+++ b/resources/views/world/species_features.blade.php
@@ -28,7 +28,7 @@
                                     @endif
                                     <p>
                                         @if (!$feature->first()->is_visible)
-                                             <i class="fas fa-eye-slash mr-1"></i>
+                                            <i class="fas fa-eye-slash mr-1"></i>
                                         @endif
                                         {!! $feature->first()->displayName !!}
                                         @if ($feature->first()->subtype)


app/Http/Controllers/WorldController.php
```diff
@@ -147,7 +147,7 @@ public function getFeatureCategories(Request $request) {
      * @return \Illuminate\Contracts\Support\Renderable
      */
     public function getFeatures(Request $request) {
-        $query = Feature::visible()->with('category')->with('rarity')->with('species');
+        $query = Feature::visible(Auth::check() ? Auth::user() : null)->with('category')->with('rarity')->with('species');
```
AW0005, line 150: Doesn't Auth::user() already return null if they aren't logged in? 🤔

AW0005 (approved):
Had one thought, but otherwise looks good

app/Http/Controllers/WorldController.php
itinerare: This I'd have to check... but this kind of check is how it's broadly done, so if nothing else it's consistent with the rest of the project.

itinerare (approved):

app/Http/Controllers/WorldController.php
SpeedyD: I did it this way mostly so it'd be consistent with the rest of the project, as itinerare noticed. 🤷

SpeedyD: @AW0005 and @itinerare
Sorry for tagging on an already merged-in PR, but I wanted to add onto this answer: I've done my research, you see.

There is a bit of fun behind this call.. You see, Auth::check() checks if there is, in fact, an Auth::user(). If it exists, it returns true. If not, it returns false. The whole function is literally `return ! is_null($this->user());`.

..And this is safer for an initial check. You see, if you do Auth::user() from the get go, and the user is in fact a guest (aka: not logged in).. it won't return null. It will _**error out** because there is **no** Auth::user()._

..It is because of this that I think that `(Auth::user() ?: null)` would not work the same, as it errors out on guests rather than return true or false, if there's no user. I have not tested this, but in theory, it should not work. You could probably use !is_null or isset yourself, but then you may as well just use Auth::check.

itinerare: Oh, I've since tried it and `Auth::user() ?? null` works fine iirc. It's just not the standard for the project at present.

SpeedyD: > Oh, I've since tried it and `Auth::user() ?? null` works fine iirc. It's just not the standard for the project at present.

In all honesty, last night Moif and I got into a short discussion about it, and we agreed that ?? would indeed likely work fine- the difference between ?: and ?? being that ?? **also** checks for _is not null_. So it gives a FALSE return if null, unlike ?: which errors out if Auth::user() does not exist.

Perhaps something to implement later on? I've read running both check and user can have an impact, even if it's rather unnoticeable right now.. but maybe on bigger databases..?..


Either way- I'll be resolving this conversation. It's been open long enough.

PR #824, opened 2024-02-07
Bug(Visual Trait Index): Hidden Traits were still showing when they shouldn't

AW0005:
Turns out Feature's scopeVisible function was a special snowflake compared to the others so having the same check sent through wasn't working right. This updates it to be the same, and also updates the `getDropdownItems` function since it was originally dependent on the different format.


itinerare:
Ahh the dropdown... that'd do it.

itinerare (approved):

SpeedyD (approved):
Haha, and I wrote the withHidden code too.. 😰

PR #827, opened 2024-02-09
Feature/currency conversion

ScuffedNewt:
Allows staff to setup direct currency conversions that users can apply in the bank page
also displays conversions on world page


resources/views/home/_bank_convert.blade.php
itinerare, line 1: Where is this file used? Implicitly, it's supposed to be included somewhere, but I couldn't spot where.

resources/views/home/bank.blade.php
```diff
@@ -42,6 +42,28 @@
         <a href="{{ url(Auth::user()->url . '/currency-logs') }}">View logs...</a>
     </div>
 
+    <h3>Convert Currency</h3>
```
itinerare, line 45: Maybe only display this section if any currency conversion options exist? Would probably prevent confusion on sites that don't make use of the feature.

itinerare (changes requested):
Seems like a reasonable inclusion given it's not too large and I feel like folk keep wanting something like this, and broadly looks good!

resources/views/home/_bank_convert.blade.php
Ne-wt: returned from ajax request

resources/views/home/bank.blade.php
Ne-wt: good idea!

resources/views/home/_bank_convert.blade.php
itinerare: Ahhh gotcha

itinerare (approved):

AW0005 (approved):

PR #828, opened 2024-02-09
fix: add back has_image checks and also change honeypot to return for better UX

ScuffedNewt:
some has images got removed way back sometime and its always bugged me, fix for that + mentioned honeypot change from ages ago


Ne-wt:
I put this on release since its just fixes / not a "new" feature

itinerare (approved):

AW0005 (approved):

PR #829, opened 2024-02-09
Bug(Shops): Round up shop purchase quantity

AW0005:
Because otherwise 0.5 because a whole item at half price.


itinerare (approved):

SpeedyD (approved):
Seems good to me. But should this not be in release, rather than main??

itinerare:
No I ok'd it, it's all good

SpeedyD:
I had no problem with it regardless. 👍

PR #847, opened 2024-02-22
add optional gallery submissions on prompts

ScuffedNewt:


app/Http/Controllers/Users/SubmissionController.php
```diff
@@ -86,21 +87,31 @@ public function getNewSubmission(Request $request) {
         $closed = !Settings::get('is_prompts_open');
         $inventory = UserItem::with('item')->whereNull('deleted_at')->where('count', '>', '0')->where('user_id', Auth::user()->id)->get();
 
+        if (config('lorekeeper.settings.allow_gallery_submissions_on_prompts')) {
+            $gallerySubmissions = GallerySubmission::where('user_id', Auth::user()->id)->where('status', 'Accepted')->orderBy('id', 'DESC')->get()->pluck('title', 'id');
```
itinerare, line 91: Would be worth including submissions on which the user collaborated as well.

app/Models/Submission/Submission.php
```diff
@@ -38,7 +39,7 @@ class Submission extends Model {
      * @var array
      */
     public static $createRules = [
-        'url' => 'nullable|url',
+        'url' => 'nullable',
```
itinerare, line 42: The `url` validation rule was added as part of a way to harden that field... is there a reason it's removed?

itinerare (changes requested):
Broadly looks good!

SpeedyD (approved):
Approved on the condition @itinerare's comments are taken into consideration. :)

app/Models/Submission/Submission.php
Ne-wt: in order for it to be used as a `title` field if needs be, but I can always just add it back!

app/Http/Controllers/Users/SubmissionController.php
Ne-wt: true!

app/Models/Submission/Submission.php
itinerare: Hmm. I think it'd probably be safest to add it back, though open to other solutions that achieve the desired functionality *and* safety.

itinerare (changes requested):

itinerare (approved):

PR #848, opened 2024-02-22
Fix(admin/traits:) Adding "without" filter options

SpeedyD:
I call this a bug fix rather than a new feature as it should've been here.

Allows you to filter traits in the admin section to "without" category, species or subtype.


AW0005 (approved):

itinerare (approved):

itinerare:
This is... still arguably a feature, regardless of whether or not it "should" be present.

PR #849, opened 2024-02-22
feat(shops): Add shop image to meta-img, if it exists

SpeedyD:
See title. 🤷


itinerare (approved):
Oh I like this, this is a good addition

AW0005 (approved):

PR #850, opened 2024-02-22
fix(titles): A lot of layout title fixes

SpeedyD:
I present to you: A TON of title fixes, consistency edits and replacement of old `@stop` Laravel code.

Exceptions include:
- account/banned.blade.php (I figured when banned, the title Banned is straight forward, should not need Account in front)
- account/deactivated.blade.php (I figured when deactivated, the title Deactviated is straight forward, should not need Account in front)

As well as the following:
- admin/submissions/submission.blade.php
- home/submissions.blade.php
- home/bug_report_index.blade.php
- home/report.blade.php

Of which all require quite a bit more edits than just this, and will be done in a seperate PR.

PR #851, opened 2024-02-22
Fix: Correcting sections

SpeedyD:
Three pages in home and one in the admin section were not properly placed. This fixes it.

(It's the four pages left unchanged in #850.)

### DIFF
--- a/resources/views/home/bug_report_index.blade.php
+++ b/resources/views/home/bug_report_index.blade.php
@@ -1,10 +1,10 @@
-@extends('layouts.app')
+@extends('home.layout')
 
-@section('title')
+@section('home-title')
     Bug Reports
 @endsection
 
-@section('content')
+@section('home-content')
     {!! breadcrumbs(['Reports' => 'reports']) !!}
     <h1>
         Bug Reports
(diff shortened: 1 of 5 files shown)


AW0005 (approved):

resources/views/home/bug_report_index.blade.php
```diff
@@ -1,10 +1,10 @@
-@extends('layouts.app')
+@extends('home.layout')
```
itinerare, line 1: Mind that the bug report index itself does not require being logged in to view-- so it *not* using the home layout is arguably more sensible.

itinerare (changes requested):

resources/views/home/bug_report_index.blade.php
SpeedyD: Hrm.. I suppose that is true.. Perhaps it should be moved altogether?

itinerare: That would probably be the better call.

SpeedyD: Never mind. Undone this change.

itinerare (approved):

PR #854, opened 2024-02-26
fix(many): @comments overwriting scripts section

SpeedyD:
This is not a problem in main, but only in release (and dev) because of the updated comments function.

Without this `@parent`, many pages which have an `@comments` tag will find their scripts function completely non-functional, as the scripts section in the comments page just.. overrides any other.

(For example: The admin side of handling reports, the Assign button did NOTHING.)

### DIFF
--- a/resources/views/comments/comments.blade.php
+++ b/resources/views/comments/comments.blade.php
@@ -80,6 +80,7 @@
 @endauth
 
 @section('scripts')
+    @parent
     <script>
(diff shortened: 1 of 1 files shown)


itinerare:
*Ah*

itinerare (approved):

SpeedyD:
> _Ah_.

_Yeah._

AW0005 (approved):

PR #863, opened 2024-02-26
Deprecate laravelcollective/html forms

itinerare:
The package [is abandoned](https://packagist.org/packages/laravelcollective/html) in favor of [spatie/laravel-html](https://github.com/spatie/laravel-html), and will need to be replaced likely before we update to Laravel 11.

However, as it stands we rely *heavily* on the `Form` utility provided by the former-- and this same is true of extensions, which largely follow the core project. `spatie/laravel-html`, meanwhile, uses [a considerably different syntax](https://spatie.be/docs/laravel-html/v2/introduction) for its HTML generation utilities; however, having migrated a few of my own projects to get a sense for how it works, the amount of effort involved, etc. I don't think it's going to be a functionality issue as much as just a lot of work. 

Still, sooner or later all of those forms and form fields are going to need to migrated to the new syntax. Conveniently, the packages seem to be able to coexist without issue, which is good because that migration is not going to happen overnight for the core project, let alone for extensions.

Consequently, it makes sense to deprecate use of the existing `Form` utility in favor of adopting and gradually migrating existing forms/form fields over to the new `html()` utilities. This is in the interests of eventually removing `laravelcollective/html` entirely, likely in v4 at the earliest (as that alone will constitute a breaking change). Ideally the migration process can happen during v3.x, as it would be good to be able to update Laravel at next convenience.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "socialiteproviders/tumblr": "^4.1",
         "socialiteproviders/twitch": "^5.3",
         "spatie/laravel-feed": "^4.1",
-        "spatie/laravel-honeypot": "^4.1"
+        "spatie/laravel-honeypot": "^4.1",
+        "spatie/laravel-html": "^3.5"
     },
     "require-dev": {
         "fakerphp/faker": "^1.9.1",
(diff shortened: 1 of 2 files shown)


SpeedyD:
O h.

Oh no.

AW0005 (approved):

Issue #181, opened 2021-09-29
Content Warning thumbnail for Characters


SpeedyD:
### Feature Request

It has recently come to my attention that the Gallery got a specific thumbnail for content warnings.
I feel it would be a good idea to extend this to the character masterlist.

itinerare:
I'm of a mind to tackle this but stake no claim in the event that someone else wants to first!

PR #876, opened 2024-03-16
Character Warnings ext to core

SpeedyD:
Pulled @redcr0wz 's extension and updated it to v3.

Current stage is _draft_ as some tweaks beyond what I've already done will need to be done.

Once completed, should officially fix issue #181.


SpeedyD:
Requesting @Ne-wt, @itinerare and @AW0005 to look things over and comment where possible. Even though I jumped into this with all vigor, my brain is tired now. All ideas welcome!!

I'll also need to figure out how to make a migration that swaps the text from Characters to CharacterImages, possibly with an extra check..

And quite a bit of extra cleanup is necessary... I already did a few minor changes to make it all consistent, but I wouldn't be surprised if missed something obvious. ^^;;

Also: severely **_untested_** at this current point.

app/Models/User/User.php
```diff
@@ -441,6 +441,24 @@ public function getBirthdayDisplayAttribute() {
...
+                return null;
+            break;
+            case 1:
+                return null;
+            break;
+            case 2:
+                return null;
+            break;
+        }
+    }
```
SpeedyD, line 461: ...I have to actually wonder why this exists. I don't think it's used anywhere?

app/Services/UserService.php
```diff
@@ -275,7 +275,22 @@ public function disableTwoFactor($code, $user) {
...
+     *
+     */
+    public function updateWarningVisibility($data, $user)
+    {
+        $user->settings->warning_visibility = $data;
+        $user->settings->save();
+
+        return true;
+    }
+
```
SpeedyD, line 291: This WILL require additional tweaking. I'll probably change it to be similar to the birthday update.

SpeedyD (commented):
These're two points I need to remember to edit soon. Comments welcome.

database/migrations/2023_10_12_011826_add_character_warning_to_character_profile.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    /**
+     * Reverse the migrations.
+     *
+     * @return void
+     */
+    public function down()
+    {
+        Schema::table('character_warning', function (Blueprint $table) {
```
Ne-wt, line 33: going to make a pr for some changes but this migration reverse is incorrect

SpeedyD: Oh shit, it is! Can you fix that in your PR as well?

app/Models/User/User.php
SpeedyD: Fixed by @redcr0wz.

database/migrations/2023_10_12_011826_add_character_warning_to_character_profile.php
SpeedyD: Thanks for the notif, @redcr0wz fixed it.

SpeedyD:
Nice work, @itinerare and @ScuffedNewt!..

Looks like there are a few conflicts to be resolved, lemme look into those quickly.

itinerare:
Happy to help :sparkles:

SpeedyD:
Is that change in [App/Services/UserService.php](https://github.com/corowne/lorekeeper/pull/876/files#diff-28a0b3aa7c8bcb21672fcfd472204061d19039d51ef8b0da41b268c070eae085) still required? I intended to give it a change similar to the birthday update back when, but.. as with everything regarding this specific PR, it fell off the grid =_=;

ScuffedNewt:
> Is that change in [App/Services/UserService.php](https://github.com/corowne/lorekeeper/pull/876/files#diff-28a0b3aa7c8bcb21672fcfd472204061d19039d51ef8b0da41b268c070eae085) still required? I intended to give it a change similar to the birthday update back when, but.. as with everything regarding this specific PR, it fell off the grid =_=;

yup iirc! it's used on the settings page

SpeedyD:
On suggestion of @ScuffedNewt, I'm putting this on ready for review. I'll resolve that BrowseController conflict in a moment.

SpeedyD:
Technically left to do:
- The earlier comment regarding UserService
- Check out all the empty TODOs, they were added for some reason, I'd assume.

Nonetheless, please review as is. :)

SpeedyD:
..helloyello, the pint action didn't run cause it's announcing
>  ! app/Models/Character/CharacterImage.php Parse error: syntax error, unexpe…  
..wish I knew the rest of that line.

itinerare:
That means the file is malformed somehow-- looks like a function got its `}`(s) chopped off.

SpeedyD:
Closing after discussion with @ScuffedNewt. Please review #1052 instead. :)

PR #900, opened 2024-04-06
Bug(User): Couldn't update user emails

AW0005:
The updateAccount function wasn't actually updating email even though the email is editable by admins from the user blade.


itinerare:
That's because the email _isn't_ editable by admins-- the field is disabled.

AW0005:
ough I did not stare at the line long enough to see that - this came up with one of my LKs I maintain because if alias login site goes down users don't have a way into their accounts if they didn't add an email account. Admins then may need to go in and edit the email on behalf of the user to get them access to their account - and it's easier if you can do it from the UI. 

I'm ok closing this if it still wouldn't be wanted but figured I'd give the usecase first. Otherwise I can remove the disable from the input as well.

itinerare:
I think that's a problem that's probably better solved by other means, e.g. suggesting that alias login folk add a different valid alias and/or email/password so they have some kind of fallback. It might even make sense to _require_ it.

PR #913, opened 2024-04-17
fix(world): Items and traits duplication issue on encyclopedia

SpeedyD:
On LKs with high amounts of items/traits, eventually items/traits start duplicating over pages. An item/trait that was halfway page 8 is also at the top of page 9.

And worst of all: When you get to the last page, you definitely have not seen all items/traits. Some of them have just poofed into the ether.

With this ingenious solution I have finally prevented this duplication!
(Note: solution is actually really, really dumb. I can't believe this fixed it.)

### DIFF
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -50,7 +50,7 @@ class WorldController extends Controller {
         }
 
         return view('world.currencies', [
-            'currencies' => $query->orderBy('name')->paginate(20)->appends($request->query()),
+            'currencies' => $query->orderBy('name')->orderBy('id')->paginate(20)->appends($request->query()),
         ]);
     }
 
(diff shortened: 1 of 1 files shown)


itinerare (approved):

itinerare:
I'd assume it can occur for most objects in /world, then, it just usually isn't a trouble/they usually aren't in significant enough numbers for the issue to manifest.

SpeedyD:
> I'd assume it can occur for most objects in /world, then, it just usually isn't a trouble/they usually aren't in significant enough numbers for the issue to manifest.

This can ABSOLUTELY occur for most objects in /world - I'm in fact surprised it's not doing it on the masterlist or userlists.

I just figured that Currencies, Rarities, Species, Species Categories, Trait Categories, Subtypes and Item Categories are unlikely to hit the same amount.

Even if so.. the same fix can be easily applied.

SpeedyD:
...and so I did. Seems like a good idea.

itinerare (approved):

PR #926, opened 2024-04-26
Feature/image blocking

CH3RVB:
This is my first PR to here, so be nice to me /hj

This is my third (i think) crack at making something like this. Everything else was either too clunky or felt... annoying to use? So this is my final solution. LOL. 

Adds a modular image blocking system for users. Users can curate their own blocklist by blocking a specific model + ID. Blocked models + IDs have their image blurred out until they become unblocked again by the user. I say model + ID because right now it's an image only thing, but theoretically you could apply this in other ways? 

With a small dosage of Moif magic, it automatically applies to all mentions of the blocked image across the site, as long as they hotlink directly to the image URL. Very easy to use across exts, simply plug in the block widget, block an image, and go.

I'm unsure about some aspects, such as the route for the image blocking, as well as how many different world/other entries I should even put this thing on. Maybe the button is also too annoying to look at? Dunno. Either way, it's here :')

### DIFF
--- a/app/Helpers/Helpers.php
+++ b/app/Helpers/Helpers.php
@@ -7,7 +7,7 @@
 |
 | Miscellaneous helper functions, primarily used for formatting.
 |
-*/
+ */
 
 /**
  * Returns class name if the current URL corresponds to the given path.
@@ -17,7 +17,8 @@
  *
  * @return string
  */
-function set_active($path, $class = 'active') {
+function set_active($path, $class = 'active')
+{
     return call_user_func_array('Request::is', (array) $path) ? $class : '';
 }
 
@@ -28,8 +29,9 @@ function set_active($path, $class = 'active') {
  *
  * @return string
  */
-function add_help($text) {
-    return '<i class="fas fa-question-circle help-icon" data-toggle="tooltip" title="'.$text.'"></i>';
+function add_help($text)
+{
+    return '<i class="fas fa-question-circle help-icon" data-toggle="tooltip" title="' . $text . '"></i>';
 }
 
 /**
@@ -39,10 +41,11 @@ function add_help($text) {
  *
  * @return string
  */
-function breadcrumbs($links) {
+function breadcrumbs($links)
+{
     $ret = '<nav><ol class="breadcrumb">';
     $count = 0;
-    $ret .= '<li class="breadcrumb-item"><a href="'.url('/').'">'.config('lorekeeper.settings.site_name', 'Lorekeeper').'</a></li>';
+    $ret .= '<li class="breadcrumb-item"><a href="' . url('/') . '">' . config('lorekeeper.settings.site_name', 'Lorekeeper') . '</a></li>';
     foreach ($links as $key => $link) {
         $isLast = ($count == count($links) - 1);
 
@@ -53,7 +56,7 @@ function breadcrumbs($links) {
         $ret .= '">';
 
         if (!$isLast) {
-            $ret .= '<a href="'.url($link).'">';
+            $ret .= '<a href="' . url($link) . '">';
         }
         $ret .= $key;
         if (!$isLast) {
@@ -77,12 +80,14 @@ function breadcrumbs($links) {
  *
  * @return string
  */
-function format_date($timestamp, $showTime = true) {
-    return $timestamp->format('j F Y'.($showTime ? ', H:i:s' : '')).($showTime ? ' <abbr data-toggle="tooltip" title="UTC'.$timestamp->timezone->toOffsetName().'">'.strtoupper($timestamp->timezone->getAbbreviatedName($timestamp->isDST())).'</abbr>' : '');
+function format_date($timestamp, $showTime = true)
+{
+    return $timestamp->format('j F Y' . ($showTime ? ', H:i:s' : '')) . ($showTime ? ' <abbr data-toggle="tooltip" title="UTC' . $timestamp->timezone->toOffsetName() . '">' . strtoupper($timestamp->timezone->getAbbreviatedName($timestamp->isDST())) . '</abbr>' : '');
 }
 
-function pretty_date($timestamp, $showTime = true) {
-    return '<abbr data-toggle="tooltip" title="'.$timestamp->format('F j Y'.($showTime ? ', H:i:s' : '')).' '.strtoupper($timestamp->timezone->getAbbreviatedName($timestamp->isDST())).'">'.$timestamp->diffForHumans().'</abbr>';
+function pretty_date($timestamp, $showTime = true)
+{
+    return '<abbr data-toggle="tooltip" title="' . $timestamp->format('F j Y' . ($showTime ? ', H:i:s' : '')) . ' ' . strtoupper($timestamp->timezone->getAbbreviatedName($timestamp->isDST())) . '">' . $timestamp->diffForHumans() . '</abbr>';
 }
 
 /**
@@ -94,8 +99,9 @@ function pretty_date($timestamp, $showTime = true) {
  *
  * @return string
  */
-function format_masterlist_number($number, $digits) {
-    return sprintf('%0'.$digits.'d', $number);
+function format_masterlist_number($number, $digits)
+{
+    return sprintf('%0' . $digits . 'd', $number);
 }
 
 /**
@@ -106,12 +112,13 @@ function format_masterlist_number($number, $digits) {
  *
  * @return string
  */
-function parse($text, &$pings = null) {
+function parse($text, &$pings = null)
+{
     if (!$text) {
         return null;
     }
 
-    require_once base_path().'/vendor/ezyang/htmlpurifier/library/HTMLPurifier.auto.php';
+    require_once base_path() . '/vendor/ezyang/htmlpurifier/library/HTMLPurifier.auto.php';
 
     $config = HTMLPurifier_Config::createDefault();
     $config->set('Attr.EnableID', true);
@@ -153,7 +160,8 @@ function parse($text, &$pings = null) {
  *
  * @return string
  */
-function parseUsers($text, &$users) {
+function parseUsers($text, &$users)
+{
     $matches = null;
     $users = [];
     $count = preg_match_all('/\B@([A-Za-z0-9_-]+)/', $text, $matches);
@@ -163,7 +171,7 @@ function parseUsers($text, &$users) {
             $user = App\Models\User\User::where('name', $match)->first();
             if ($user) {
                 $users[] = $user;
-                $text = preg_replace('/\B@'.$match.'/', $user->displayName, $text);
+                $text = preg_replace('/\B@' . $match . '/', $user->displayName, $text);
             }
         }
     }
@@ -180,7 +188,8 @@ function parseUsers($text, &$users) {
  *
  * @return string
  */
-function parseUsersAndAvatars($text, &$users) {
+function parseUsersAndAvatars($text, &$users)
+{
     $matches = null;
     $users = [];
     $count = preg_match_all('/\B%([A-Za-z0-9_-]+)/', $text, $matches);
@@ -190,7 +199,7 @@ function parseUsersAndAvatars($text, &$users) {
             $user = App\Models\User\User::where('name', $match)->first();
             if ($user) {
                 $users[] = $user;
-                $text = preg_replace('/\B%'.$match.'/', '<a href="'.$user->url.'"><img src="'.$user->avatarUrl.'" style="width:70px; height:70px; border-radius:50%; " alt="'.$user->name.'\'s Avatar"></a>'.$user->displayName, $text);
+                $text = preg_replace('/\B%' . $match . '/', '<a href="' . $user->url . '"><img src="' . $user->avatarUrl . '" style="width:70px; height:70px; border-radius:50%; " alt="' . $user->name . '\'s Avatar"></a>' . $user->displayName, $text);
             }
         }
     }
@@ -207,7 +216,8 @@ function parseUsersAndAvatars($text, &$users) {
  *
  * @return string
  */
-function parseUserIDs($text, &$users) {
+function parseUserIDs($text, &$users)
+{
     $matches = null;
     $users = [];
     $count = preg_match_all('/\[user=([^\[\]&<>?"\']+)\]/', $text, $matches);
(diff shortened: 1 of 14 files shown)


app/Helpers/Helpers.php
```diff
@@ -444,8 +462,33 @@ function prettyProfileName($url) {
...
+
+    $block = App\Models\ImageBlock::where([
+        ['item_id', '=', $item->id],
+        ['item_type', '=', $model],
+        ['user_id', '=', $user->id],
+    ])->first();
+
+    if ($block) {
+        return true;
+    } else {
```
itinerare, line 491: Generally speaking, with returns, you don't need to encase the second in the if/else-- nothing beyond a  `return` is executed  to begin with, so you can simply have the first in the if, then the second freestanding.

app/Models/ImageBlock.php
```diff
@@ -0,0 +1,51 @@
...
+class ImageBlock extends Model
+{
+
+    /**
+     * The attributes that are mass assignable.
+     *
+     * @var array
+     */
+    protected $fillable = [
+        'item_id', 'item_type', 'user_id'
```
itinerare, line 17: Referring to non-specific objects as "items" when "item" is a concept already extant in the codebase is... very confusing.

resources/views/layouts/app.blade.php
```diff
@@ -82,6 +82,18 @@
     @endif
 
     @include('feed::links')
+
+    {{-- Image Blocking --}}
+    @if (Auth::check() && count(Auth::user()->blockedImages))
+        <style>
+            @foreach (Auth::user()->blockedImages as $image)
+                [src="{{ $image->item->imageUrl ? $image->item->imageUrl : ($image->item->currencyImageUrl ? $image->item->currencyImageUrl : ($image->item->avatarUrl ? $image->item->avatarUrl : '')) }}"] {
```
itinerare, line 90: This logic would be better done in an attribute on the image block model, I think. Would be a good use case for a switch statement, potentially.

resources/views/layouts/app.blade.php
```diff
@@ -82,6 +82,18 @@
     @endif
 
     @include('feed::links')
+
+    {{-- Image Blocking --}}
+    @if (Auth::check() && count(Auth::user()->blockedImages))
```
itinerare, line 87: The count here should just be able to be `Auth::user()->blockedImages->count()`, since iirc what you get out of a relation is either directly or functionally a collection.

resources/views/widgets/_item_block.blade.php
```diff
@@ -0,0 +1,7 @@
+ {!! Form::open(['url' => 'account/image-blocks/block/' . $item->id]) !!}
+ @php
+     $model = get_class($item);
+ @endphp
+ <input type="hidden" name="item_type" value="{{ $model }}" />
+ {!! Form::submit(checkItemBlock($item, Auth::user()) ? 'Unblock' : 'Block', ['class' => 'btn btn-success btn-sm '.(isset($float) && $float ? 'float-right' : '')]) !!}
```
itinerare, line 6: Probably this should be clearer about *what's* being blocked.

resources/views/widgets/_item_block.blade.php
```diff
@@ -0,0 +1,7 @@
+ {!! Form::open(['url' => 'account/image-blocks/block/' . $item->id]) !!}
+ @php
+     $model = get_class($item);
+ @endphp
+ <input type="hidden" name="item_type" value="{{ $model }}" />
```
itinerare, line 5: While the risk is low in a project like this that's *already* open source, it's generally not the best idea to plunk stuff like model names out in the open; comments already went through this, so have a look [at the form for those](/corowne/lorekeeper/blob/develop/resources/views/comments/_form.blade.php), etc. for an idea as to what to do about it.

routes/lorekeeper/members.php
```diff
@@ -52,6 +52,8 @@
     Route::post('bookmarks/edit/{id}', 'BookmarkController@postCreateEditBookmark');
     Route::get('bookmarks/delete/{id}', 'BookmarkController@getDeleteBookmark');
     Route::post('bookmarks/delete/{id}', 'BookmarkController@postDeleteBookmark');
+
+    Route::post('image-blocks/block/{id}', 'AccountController@postBlockUnblockImage');
```
itinerare, line 56: If you go a similar route to comments (wrt the above), this'll probably need to be changed accordingly, but the general gist and placement of the route are fine.

app/Services/UserService.php
```diff
@@ -643,4 +661,47 @@ public function reactivate($user, $staff = null) {
...
+        DB::beginTransaction();
+
+        try {
+            $model = $data['item_type'];
+            $item = $model::find($id);
+            $user = Auth::user();
+
+            //check it's not already blocked, if not, make the block
+
+            $block = ImageBlock::where([
```
itinerare, line 684: You could probably trim off a little work and express this as `$user->blockedImages->where([non-user-id-conditions etc etc])->first()` too!

app/Services/UserService.php
```diff
@@ -643,4 +661,47 @@ public function reactivate($user, $staff = null) {
...
+     *
+     * @return bool
+     */
+    public function blockUnblockImage($data, $id)
+    {
+        DB::beginTransaction();
+
+        try {
+            $model = $data['item_type'];
+            $item = $model::find($id);
```
itinerare, line 679: It might be worthwhile to double-check that the object exists just to be safe, since it's technically in the 'accepting semi-arbitrary input from users' zone.

itinerare (changes requested):
A good stab at an arguably relatively complex task!

itinerare:
Belatedly, also... it might seem counterintuitive, but it might be an idea to have a page (akin to bookmarks) with the current user's blocked images, as a ready point of reference.

app/Helpers/Helpers.php
Ne-wt, line 1: nit: run composer lint to stop these weird style changes

app/Helpers/Helpers.php
Ne-wt: +1, can even just 
`return $block ? true : false`

app/Http/Controllers/Users/AccountController.php
Ne-wt, line 1: again with composer lint

app/Http/Controllers/Users/AccountController.php
```diff
@@ -475,4 +504,26 @@ public function postReactivate(Request $request, UserService $service) {
...
+     *
+     */
+    public function postBlockUnblockImage(Request $request, UserService $service, $id)
+    {
+        $data = $request->only([
+            'item_type'
+        ]);
+
+        if ($service->blockUnblockImage($data, $id)) {
+
```
Ne-wt, line 519: should this be blank? if yes, change it to `!$service...` and remove else block

app/Models/ImageBlock.php
Ne-wt: +1, recommend `block_object_id` or `image_id` `image_type` etc

app/Models/ImageBlock.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace App\Models;
```
Ne-wt, line 3: should this be in the user space, since it is user specific, and be `UserImageBlock`? imo I think it should be since there is no user agnostic rows

app/Services/UserService.php
Ne-wt, line 1: +1 to lint

database/migrations/2024_04_25_052128_add_image_blocking.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration
```
Ne-wt, line 7: this should be `class FileName extends Migration` to stop errors on older installs

resources/views/user/_profile_content.blade.php
```diff
@@ -6,6 +6,9 @@
     <div class="col-md-2 text-center">
         <!-- User Icon -->
         <img src="{{ $user->avatarUrl }}" class="img-fluid rounded-circle" style="max-height: 125px;" alt="{{ $user->name }}'s Avatar">
+        @if(Auth::check())
+                @include('widgets._item_block', ['item' => $user])
+            @endif
```
Ne-wt, line 11: nit: indents

Ne-wt: double nit: item should not be the variable name

resources/views/widgets/_item_block.blade.php
```diff
@@ -0,0 +1,7 @@
+ {!! Form::open(['url' => 'account/image-blocks/block/' . $item->id]) !!}
+ @php
+     $model = get_class($item);
```
Ne-wt, line 3: further on prior nit: `$item` is not the best variable name

resources/views/world/_currency_entry.blade.php
```diff
@@ -8,6 +8,9 @@
         <h3>{{ $currency->name }} @if ($currency->abbreviation)
                 ({{ $currency->abbreviation }})
             @endif
+            @if(Auth::check())
+                @include('widgets._item_block', ['item' => $currency, 'float' => true])
```
Ne-wt, line 12: float seems to be true in most places, consider it being default to true and explicitly pass it as false, but that's mostly preference

ScuffedNewt (changes requested):
mostly nits, agree with what mercury has said
apologies if comments are short - in work PR review mode

Ne-wt:
Also unrelated, I find this a little bit strange to include in the default LK install, considering things like toyhouse etc etc don't employ this on general images, and if a specific lorekeeper has such images on items etc that may need to be blocked its best installed individually, imo
@itinerare thoughts?

itinerare:
On some level, I agree... but users may need to block seemingly innocuous stuff in certain cases, so it's not necessarily sufficient to install it on a case-by-case basis/only where more common points of friction occur.

itinerare:
To some degree, I feel like if it's worth including in core at all then there's no sense in undercutting it by putting it behind a toggle... 

I think the other part of it is that TH has a different use case than we do, so it's not necessarily practical to make a direct comparison there; TH doesn't really have images that persistently exist around the site in the same way (items, etc.) to be an issue in the first place. Generally speaking, you have more leeway to simply not look at a character or image on there that causes issues, and iirc can block characters-- so it does have _some_ manner of comparable function, just that the context is different.
The more direct analogue would probably be pet sites-- and imo the fact that those generally don't have this kind of option isn't in their favor.

It's one of those things where it's the safer and more considerate thing to be proactive about it and spare folks unnecessary stress, even if it feels like "overkill" in a lot of cases/to include in core.

Ne-wt:
> The more direct analogue would probably be pet sites

definitely, hit the nail on the head! brain wasn't braining yday

> It's one of those things where it's the safer and more considerate thing to be proactive about it and spare folks unnecessary stress, even if it feels like "overkill" in a lot of cases/to include in core.

firmly in the camp of <this is probably not a feature conducive? to core> since most items etc should not be causing these issues ideally, especially when compared to Lorekeepers mainstream counterparts
That being said, I think this would be better placed as having `Characters` as default -- since when #876 is inevitably cleaned and finished, users may still find that they want to block a certain character even if does is not necessarily fall under a "warning needed" category. 
In fact, I think that would be a much more requested feature than items

TLDR; IMO, if this feature is to be added...

- Change the primary focus of the default configuration to be for characters
- A toggle, if not (since other use cases would need to be manually added by a developer)

app/Models/ImageBlock.php
```diff
@@ -0,0 +1,51 @@
...
+    protected $fillable = [
+        'item_id', 'item_type', 'user_id'
+    ];
+
+    /**
+     * The table associated with the model.
+     *
+     * @var string
+     */
+    protected $table = 'image_blocks';
```
Ne-wt, line 25: following up on my prior comment about user space too, ideally the table should be `user_image_blocks` in similar fashion

itinerare:
> definitely, hit the nail on the head! brain wasn't braining yday

Haha, all good, been there.

Anyhow, yeah, I'm inclined to agree that characters are the bigger/more probable use case, so I think it's sensible enough to aim this more firmly _at_ that use case-- with support for other things essentially as a side benefit (albeit one that I do still find in favor of, personally). I'd agree that most items, etc. _shouldn't_ cause issues/it's generally good to avoid making ones that are liable to, but I'd rather provide users the ability to address these things for themselves.

app/Helpers/Helpers.php
itinerare: Probably I'm just going to switch it to running on push for all branches again so neither reviewers nor contributors need to think about it, since my clever idea wrt running on PRs only was stymied (fairly reasonably) by repo perms... It was worth a shot |D

Ne-wt: 👌 
> It was worth a shot |D

can't work out all the time!

app/Helpers/Helpers.php
Ne-wt: actually follow up on this i'd do

`function checkImageBlock($image, $user)
{
    return App\Models\ImageBlock::where([
        ['image_id', '=', $image->id],
        ['image_type', '=', get_class($image)],
        ['user_id', '=', $user->id],
    ])->exists();
}`

app/Helpers/Helpers.php
itinerare: Did that while I was thinking about it-- so it _should_ now be as easy as just merging in latest develop. Hopefully.

app/Models/ImageBlock.php
itinerare: Good call yeah, inclined to agree

database/migrations/2024_04_25_052128_add_image_blocking.php
itinerare: _Does_ this cause issues? My understanding was that upstream (wrt Laravel) this change was following PHP supporting it, so it doesn't make much sense to me that (assuming PHP is upgraded, which needs to happen regardless) this should be a problem.

app/Models/ImageBlock.php
itinerare: Likewise, agree on this

database/migrations/2024_04_25_052128_add_image_blocking.php
Ne-wt: I've found on SOME sites it can be an issue, & quiet often on local (people on windows / older xampp)
perhaps in this case it *should* be fine since this is on v3 which should implicitly mean that the php version is up to par, so take this comment with a grain of salt

itinerare: I'm inclined to let this be since yeah, v3 explicitly requires PHP be reasonably up-to-date (whereas on current release it's supported but not technically _forced_)...

CH3RVB:
cancelling due to feedback

PR #944, opened 2024-05-06
fix(reports): include parent scripts in scripts section

itinerare:
Probably just escaped a bit in one of the prior edits; easy fix either way.


SpeedyD (approved):
Oops. That's my bad- fixed something else, and broke this. Sorry bout that.
(See https://github.com/corowne/lorekeeper/pull/857/files)

itinerare:
All good, it happens!

AW0005 (approved):

PR #945, opened 2024-05-07
fix gallery queue url in notif

rattusquo:
'Gallery Submission Rejected' notification was using the wrong url, causing a 404 error when clicked


itinerare:
Looks like this is present in v3 as well, so this should be targeting the release branch (and absent any content from develop).

PR #947, opened 2024-05-11
Bug(Character): Hidden characters not showing for admin

AW0005:
Fixes case found in the discord where if you create a Sale you can't see hidden characters despite having the appropriate permissions as an admin.

### DIFF
--- a/app/Models/Character/Character.php
+++ b/app/Models/Character/Character.php
@@ -216,7 +216,11 @@ class Character extends Model {
      *
      * @return \Illuminate\Database\Eloquent\Builder
      */
-    public function scopeVisible($query) {
+    public function scopeVisible($query, $user = null) {
+        if ($user && $user->hasPower('manage_characters')) {
+            return $query;
+        }
+        
         return $query->where('is_visible', 1);
     }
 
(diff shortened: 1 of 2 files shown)


AW0005:
... why did the blade formatter update the package json file

itinerare:
I think it's a weird artifact of how npm behaves or somesuch but not actually representative of any real change...

itinerare (approved):

itinerare:
Anyhow, I think it'd also be prudent to go through and adjust existing queries to reflect this, but it'd probably be reasonable to do that in a separate PR/targeting develop, etc. This change on its own is kinda scooting through since it *is* answering a standing bug....

SpeedyD (approved):
I'm unsure what precisely this actually fixes, but I see no problem with adding this. Useful for future development, I'm sure.

itinerare:
Presently the query for adding characters to sales supplies the user to the visibility scope... despite that not actually being supported... so it's impossible to add hidden characters, even if you are otherwise able to see them.

SpeedyD:
> Presently the query for adding characters to sales etc. supplies the user to the visibility scope... despite that not actually being supported... so it's impossible to add hidden characters, even if you are otherwise able to see them.

...I swore I fixed this prior, but I've been shown what the problem is and.. this is caused by the selectize update =_=;...

itinerare:
I mean, prior, that was _intended_ behavior... and, to be fair, the culprit is mostly that particular scope not being consistent with most of the others within the project.

PR #949, opened 2024-05-13
Feat(items/currency): Block direct transfer by site settings

SpeedyD:
Don't want your users to directly transfer items or currency, and instead pass it through a trade?

Aight, just change your site settings. :)

### DIFF
--- a/app/Console/Commands/AddSiteSettings.php
+++ b/app/Console/Commands/AddSiteSettings.php
@@ -82,6 +82,10 @@ class AddSiteSettings extends Command {
         $this->addSiteSetting('deactivated_key', 0, 'Optional key to view the deactivated list. Enter "0" to not require one.');
 
         $this->addSiteSetting('comment_dislikes_enabled', 0, '0: Dislikes disabled, 1: Dislikes enabled.');
+        
+        $this->addSiteSetting('can_transfer_currency_directly', 1, 'Can users directly transfer currency to other users without trading? 0: No, they cannot, 1: Yes, they can.');
+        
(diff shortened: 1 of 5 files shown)


app/Console/Commands/AddSiteSettings.php
```diff
@@ -82,6 +82,10 @@ public function handle() {
         $this->addSiteSetting('deactivated_key', 0, 'Optional key to view the deactivated list. Enter "0" to not require one.');
 
         $this->addSiteSetting('comment_dislikes_enabled', 0, '0: Dislikes disabled, 1: Dislikes enabled.');
+        
+        $this->addSiteSetting('can_transfer_currency_directly', 1, 'Can users directly transfer currency to other users without trading? 0: No, they cannot, 1: Yes, they can.');
```
itinerare, line 86: These descriptions should probably be written more dryly/in the fashion of the existing site settings text, for consistency.

SpeedyD: I based this style of writing approximately off of how the blacklist / deactivation options are written..

May I request a suggestion for how to rewrite this?

itinerare: Hmm, maybe something like... `Whether or not users can directly transfer [items/currency] to other users without trading. 0: Users cannot directly transfer [items/currency]. 1: Direct [item/currency] transfers are allowed.`

SpeedyD: I will copy paste this in. 👍

itinerare (approved):

AW0005 (approved):

itinerare:
Just needs conflict resolution.

SpeedyD:
> Just needs conflict resolution.

Done and done.

PR #957, opened 2024-05-18
feat: Add ability for users to cancel their own design requests

preimpression:
Adds the ability for users to cancel their own design requests. Optional by way of config. By default it is set to ON.

### DIFF
--- a/app/Http/Controllers/Characters/DesignController.php
+++ b/app/Http/Controllers/Characters/DesignController.php
@@ -52,8 +52,8 @@ class DesignController extends Controller {
         }
 
         return view('character.design.request', [
-            'request' => $r,
-            'canCancel' => config('lorekeeper.extensions.design_return_to_draft')
+            'request'   => $r,
+            'canCancel' => config('lorekeeper.extensions.design_return_to_draft'),
         ]);
     }
 
--- a/app/Http/Controllers/Characters/DesignController.php
+++ b/app/Http/Controllers/Characters/DesignController.php
@@ -53,7 +53,7 @@ class DesignController extends Controller {
 
         return view('character.design.request', [
             'request'   => $r,
-            'canCancel' => config('lorekeeper.extensions.design_return_to_draft'),
+            'canCancel' => config('lorekeeper.extensions.design_return_to_draft') ?? 0,
         ]);
     }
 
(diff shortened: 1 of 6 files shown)


app/Services/DesignUpdateManager.php
```diff
@@ -819,21 +820,30 @@ public function cancelRequest($data, $request, $user) {
             // to add a comment to it. Status is returned to Draft status.
             // Use when rejecting a request that just requires minor modifications to approve.
 
-            // Set staff comment and status
-            $request->staff_id = $user->id;
-            $request->staff_comments = $data['staff_comments'] ?? null;
-            $request->status = 'Draft';
-            if (!isset($data['preserve_queue'])) {
+            if ($self == null) {
```
Ne-wt, line 823: this can just be !$self

app/Services/DesignUpdateManager.php
```diff
@@ -819,21 +820,30 @@ public function cancelRequest($data, $request, $user) {
...
+            $request->status = 'Draft';
             $request->save();
 
-            // Notify the user
-            Notifications::create('DESIGN_CANCELED', $request->user, [
-                'design_url'    => $request->url,
-                'character_url' => $request->character->url,
-                'name'          => $request->character->fullName,
-            ]);
+            if ($self == null) {
```
Ne-wt, line 838: same here

Ne-wt (approved):
good feature, only minor comments 👍

app/Http/Controllers/Characters/DesignController.php
```diff
@@ -52,7 +52,8 @@ public function getDesignUpdate($id) {
         }
 
         return view('character.design.request', [
-            'request' => $r,
+            'request'   => $r,
+            'canCancel' => config('lorekeeper.extensions.design_return_to_draft'),
```
itinerare, line 56: Could probably make this `'canCancel' => config(etc etc) ?? 0` or similar and skip the `isset`s later.

preimpression: Good call; I initially had the isset because I was under the assumption that admins had a different function to get to this page but after reviewing it again, it seems to all be through this one.

itinerare: Ahhh yeah that makes sense. I was gonna say, I think the config helper would be fine on its own, but if you want to be _absolutely_ sure...

itinerare (approved):

itinerare (approved):

PR #958, opened 2024-05-19
Remove duplicate edit-credits in js

Cylunny:
I noticed on my host when editing the credits of a character, the selectize would sometimes act freaky and appear empty/non functional. After digging for a while, I realized that the edit credits modal request gets sent twice. I think the reason is the event listener being duplicated. After removing it, the issue was resolved on my host - no more empty select and double requests.

(I checked the wiki on how to get small fixes in, so I hope I did it right! Sorry if not orz)


itinerare (approved):
Oh, that's weird... Anyhow, yep, you're in the right place/everything's in order!

Ne-wt (approved):

PR #959, opened 2024-05-19
Fix: Character Image Size Validation

ScuffedNewt:
[currently the size validation is incorrectly set to 20mb ish instead of 2mb](https://laravel.com/docs/11.x/validation#validating-files:~:text=(1024)-,%2D%3Emax(12%20*%201024)%2C,-%5D%2C)
also added validation on admin reuploads
the necessity of the validation can be debated, my thought process is if an admin tries to reupload a large image and brings down their PHP process, it can cause a lot of delays unless there is an active coder around to reboot etc

### DIFF
--- a/app/Models/Character/Character.php
+++ b/app/Models/Character/Character.php
@@ -78,8 +78,8 @@ class Character extends Model {
         'slug'                  => 'required|alpha_dash',
         'description'           => 'nullable',
         'sale_value'            => 'nullable',
-        'image'                 => 'required|mimes:jpeg,jpg,gif,png|max:20000',
-        'thumbnail'             => 'nullable|mimes:jpeg,jpg,gif,png|max:20000',
+        'image'                 => 'required|mimes:jpeg,jpg,gif,png|max:2048',
+        'thumbnail'             => 'nullable|mimes:jpeg,jpg,gif,png|max:2048',
         'owner_url'             => 'url|nullable',
     ];
 
@@ -109,8 +109,8 @@ class Character extends Model {
         'description' => 'nullable',
         'sale_value'  => 'nullable',
         'name'        => 'required',
-        'image'       => 'nullable|mimes:jpeg,gif,png|max:20000',
-        'thumbnail'   => 'nullable|mimes:jpeg,gif,png|max:20000',
+        'image'       => 'nullable|mimes:jpeg,gif,png|max:2048',
+        'thumbnail'   => 'nullable|mimes:jpeg,gif,png|max:2048',
     ];
 
     /**********************************************************************************************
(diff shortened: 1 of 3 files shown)


AW0005:
🤔 I wonder if this is something that changed with a laravel version or something because I would swear it limited at about 4mb when I first started building my LK (and eventually removed it), but it doesn't look like it was ever anything but KB from the docs...

In any case I'd argue for at least 4mb so it matches TH - I'd suspect otherwise there are likely to be complaints about it being smaller than that.

AW0005:
Hrm, maybe not _hard_ to change per se, but it is located in places I don't think the average non-coder LK owner will think about going to change, and then when it becomes a problem, they come into the LK server asking about how to change it because they can't find it, which is really my main concern.

I'd probably at least see about putting them all on a site/config setting? Makes it easier to change them all together and easier to find.

itinerare:
Yeah, hooking it up to a config setting would probably be a good course of action, especially since then that presents the opportunity to provide info/text saying why you should/shouldn't adjust the limit. |D

Ne-wt:
because they're static variables you can't assign a dynamic config to them unless it's in a method
If someone else wants to take on that refactor they're more than welcome, or I can do it in a separate pr at a later date, but considering this is just intended to correct unwanted behaviour I wont be doing it in this PR
image
image

itinerare:
That's reasonable, especially considering going into the release it really shouldn't be adding a new feature like that...

itinerare (approved):

SpeedyD (approved):
Honestly, I'm very curious how this happened...

...also, a sidenote, since Laravel documentation states they allow '2kb' as input as well, I'm half wondering why we're even specifiying bytes, but.. I guess that's more nitpick than constructive criticism.

itinerare:
I vaguely recalled as much/that elsewhere I've used that in validation rules, yeah...

app/Models/Character/Character.php
```diff
@@ -109,8 +111,8 @@ class Character extends Model {
         'description' => 'nullable',
         'sale_value'  => 'nullable',
         'name'        => 'required',
-        'image'       => 'nullable|mimes:jpeg,gif,png|max:20000',
-        'thumbnail'   => 'nullable|mimes:jpeg,gif,png|max:20000',
+        'image'       => 'nullable|mimes:jpeg,gif,png|max:2048',
+        'thumbnail'   => 'nullable|mimes:jpeg,gif,png|max:2048',
```
SpeedyD, line 115: These still say 2048 🤔

ScuffedNewt: whoopsies

itinerare (approved):

SpeedyD (approved):

PR #967, opened 2024-06-03
Fix: Optimalizations and redundancy removal

SpeedyD:
Fixes included in this PR:
- For consistency, every single admin page has cols and form-group merged. I found this to be a common occurrence and this saves on a bit of space.
- There was a single redundant bracket on `resources/views/prompts/_prompt_entry.blade.php`. It's now removed. (This was found by to iinkt on the LK discord.)
- I found a minor mistake on the Character Designs/Art page; the art version's breadcrumbs had the wrong url.
- I realized the Character Designs/Art page shares most of it's code with the characters, myos and sublist pages. I have thus reused the same code.
- Also for these pages, I added an 'userpage_exts' check, this so that extension additions exclusive to userpage characters have a space to be added. This is left open with a blade comment. (Mostly done with co-owners in mind.)
- Likewise, I also added an 'owner' check, which hides the owner of the character if it's true, as the characters on the page would then belong to the user. If it's false, it shows the owner, assuming here that the characters on the page do not belong to the user.
- Finally, I realized the Character Designs/Art page lacked the username in the header, it was in the page title and all other character pages seemed to have it, so I figured I'd add it for consistency.
- Consolidated the visual indexes and their modal, secured it slightly better too.

From anyone reviewing, I'd also like to know the opinions on the following:

~~1. I was wondering if I shouldn't simply consolidate all the different characters pages into a single page. . . It seems like it can be done? Just pass along an extra variable from the controller..~~
**This has been decided to be unnecessary.**

~~2. Unrelated to the fixes above, but part of this PR was intended to be the consolidation of the trait modals of the species and universal visual trait indexes. I opted not to as there's some minor differences, but.. the biggest difference is that one demands the species, which seems to be more due to consistency with the modal url than actually necessary.. So I think merging the two can save some space. (Also keeping in mind a future subtype visual trait index, which I believe was a request..)~~
**This has been added to the PR.**

### DIFF
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -234,7 +234,7 @@ class WorldController extends Controller {
         if (!$species) {
             abort(404);
         }
-        if (!config('lorekeeper.extensions.species_trait_index.enable')) {
+        if (!config('lorekeeper.extensions.visual_trait_index.enable_species_index')) {
             abort(404);
         }
 
@@ -278,29 +278,6 @@ class WorldController extends Controller {
     }
 
     /**
-     * Provides a single trait's description html for use in a modal.
-     *
-     * @param mixed $speciesId
-     * @param mixed $id
-     *
-     * @return \Illuminate\Contracts\Support\Renderable
-     */
-    public function getSpeciesFeatureDetail($speciesId, $id) {
-        $feature = Feature::where('id', $id)->first();
-
-        if (!$feature) {
-            abort(404);
-        }
-        if (!config('lorekeeper.extensions.species_trait_index.trait_modals')) {
-            abort(404);
-        }
-
-        return view('world._feature_entry', [
-            'feature' => $feature,
-        ]);
-    }
-
-    /**
      * Shows a universal visual trait list.
      *
      * @return \Illuminate\Contracts\Support\Renderable
@@ -309,7 +286,7 @@ class WorldController extends Controller {
         $categories = FeatureCategory::orderBy('sort', 'DESC')->get();
         $rarities = Rarity::orderBy('sort', 'ASC')->get();
 
-        if (!config('lorekeeper.extensions.universal_trait_index.enable')) {
+        if (!config('lorekeeper.extensions.visual_trait_index.enable_universal_index')) {
             abort(404);
         }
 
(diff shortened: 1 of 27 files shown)


itinerare:
`form-group` will eventually become nothing fancier than `mb-3` with the transition to bootstrap 5, so fair enough there.

I don't think it's necessary to consolidate *every* instance of a similar view into one, especially if significant portions can be reused between them.
And, ultimately, the motivation is mostly just 'whatever makes things easier to maintain, long-term'.

itinerare (approved):
Also, in general, it's still a good call to split things up so that each PR has *one* single thing it's addressing; this makes them easier to review and document.

SpeedyD:
> Also, in general, it's still a good call to split things up so that each PR has _one_ single thing it's addressing; this makes them easier to review and document.

Fair enough, I'll leave that part be, then. 👍 

Regarding the Visual Trait Indexes, though? We currently have two different setups, two different modal routes.. when it could simply be one and the same function. Easier to maintain, review _and_ document, since it's all the same thing. The only differences should be that the Species version has Subtypes added to it. (The Universal one does not, as Subtypes are related to the species.. And when I make the Subtype version, it will also miss it.)

itinerare:
Yeah, I have no qualms with those modals being condensed.

SpeedyD:
> Yeah, I have no qualms with those modals being condensed.

Then allow me to ask you do not yet merge this until I get that done. :)

SpeedyD:
There we go. Review and merge away. :)

app/Http/Controllers/WorldController.php
```diff
@@ -336,23 +313,20 @@ public function getUniversalFeatures(Request $request) {
...
      */
-    public function getUniversalFeatureDetail($id) {
-        $feature = Feature::where('id', $id)->first();
+    public function getFeatureDetail($id) {
+        $feature = Feature::visible(Auth::user() ?? null)->where('id', $id)->first();
 
         if (!$feature) {
             abort(404);
         }
-        if (!config('lorekeeper.extensions.universal_trait_index.trait_modals')) {
```
itinerare, line 353: Is there a reason you omitted this?

SpeedyD: It's effectively useless.

In normal use, the user will never end up at this function unless they've gone through the trait index already, access to which is sealed off by the links being hidden _and_a 404 error in the functions for them. It's effectively the THIRD barrier to a modal that's.. otherwise not that approachable anyway.

Having the check whether the trait is MEANT to be visible is much more powerful, for blocking use, though it won't get there either, considering the checks on both visual indexes.

And finally: The fact the two visual indexes are consolidated into this one means we'd need two checks done at the same time, possibly need to add another variable just to distinguish between each.. that's a lot of extra work and effort for what I honestly think is a useless check.

SpeedyD: I forgot to add: For non-conventional approaches, I am appalled that we forgot to have at least an 'is visible' check here.

Major fail on our part, good thing common users do NOT know how to approach the feature entries from the modal's route.. And that is also why I added that check.

itinerare: Fair enough!

itinerare (approved):

AW0005 (approved):

PR #970, opened 2024-06-03
feat(world): Subtype Trait Index and code condensation

SpeedyD:
Subtype Trait Index. A requested feature.


SpeedyD:
Hang on, I'd like to add something. Setting this to draft.

SpeedyD:
There we go. Now you can flip a switch to include all traits from the same species that are NOT subtype specific, so that you can have all traits on one page that a single subtype can officially have.

app/Http/Controllers/WorldController.php
```diff
@@ -277,6 +277,93 @@ public function getSpeciesFeatures($id) {
...
+                        if ($feature->subtype) {
+                            if ($feature->subtype->id != $subtype->id) {
+                                return false;
+                            } else {
+                                return $feature->subtype->is_visible;
+                            }
+                        }
+
+                        return true;
+                    })
```
AW0005, line 355: Since the filter comes after the `get()` anyway it might be cleaner to do this on a variable of the results of the `count($categories) ?` ternary instead of duplicating that bit of code?

SpeedyD: ...because I honestly have no idea how.

Besides, the regular Species Visual Index also does it similar to this.. So.. consistency? ^^;;;

SpeedyD: So I did the following:
```php
        } else {
            $species = Species::visible(Auth::user() ?? null)->where('id', $subtype->species->id)->first();
            
            $features = count($categories) ?
                $species->features()
                    ->visible(Auth::user() ?? null)
                    ->orderByRaw('FIELD(feature_category_id,'.implode(',', $categories->pluck('id')->toArray()).')')
                    ->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')')
                    ->orderBy('has_image', 'DESC')
                    ->orderBy('name')
                    ->get() :
                $species->features()
                    ->visible(Auth::user() ?? null)
                    ->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')')
                    ->orderBy('has_image', 'DESC')
                    ->orderBy('name')
                    ->get();
                    
            $features->filter(function ($feature) use ($subtype) {
                        if ($feature->subtype) {
                            if ($feature->subtype->id != $subtype->id) {
                                return false;
                            } else {
                                return $feature->subtype->is_visible;
                            }
                        }

                        return true;
                    })
                    ->groupBy(['feature_category_id', 'id']);
        }
```
and now it gives me
`Too few arguments to function Illuminate\Database\Eloquent\Builder::chunk(), 1 passed in [LK FOLDER]\vendor\laravel\framework\src\Illuminate\Support\Traits\ForwardsCalls.php on line 23 and exactly 2 expected`

(Though it says line 23, it marks down line 37 instead: `@foreach ($categoryFeatures->chunk(4) as $chunk)`)

SpeedyD: @AW0005 is doing a PR, we've been in discussion over Discord. :)

SpeedyD:
It should be review ready once again. @AW0005 and I are the SQUASH SQUAD. 😎

itinerare (approved):
Nice condensing these down!

AW0005 (approved):

PR #971, opened 2024-06-04
Fix: Remove old fa-close line from character selection

MarskyMessier:
My first PR, I hope I'm doing everything right!

Noticed that there was still old unused lines of code in the character selection menus concerning submissions, sales, galleries and where you'd have select_entry and select.blade I believe. Didn't do anything nor was linked to any JS and wasn't visible on v3.0.0 unless you had an extension that made it visible because of custom `<a>` classes like Theme Manager.

`remove-character` is the correct class given it's in the JS and the copy file to delete the selected entry when on-click.

### DIFF
--- a/resources/views/admin/sales/_character_select.blade.php
+++ b/resources/views/admin/sales/_character_select.blade.php
@@ -19,7 +19,6 @@
                     </div>
                 </div>
                 <div class="col-md-10">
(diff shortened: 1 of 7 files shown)


SpeedyD (approved):
Not sure why it's there.. probably an old leftover that got missed. Nice catch. :)

itinerare (approved):

PR #974, opened 2024-06-16
hotfix(security|requests): add middleware that parses post fields

Ne-wt:
# Context
user reported a security concern verified locally (can utilise HTML in character names for example)

# Description
Applies a middleware to all requests that:
- parses all fields, minus `excludedFields`
- additionally strips those in `strippedFields` for extra security (since `parse()` doesn't remove HTML tags such a `<b>` etc)

### DIFF
--- a/app/Helpers/Helpers.php
+++ b/app/Helpers/Helpers.php
@@ -101,7 +101,6 @@ function parse($text, &$pings = null) {
     $config->set('Attr.EnableID', true);
     $config->set('HTML.DefinitionID', 'include');
     $config->set('HTML.DefinitionRev', 2);
-	$config->set('Cache.DefinitionImpl', null); // TODO: remove this later!
     if ($def = $config->maybeGetRawHTMLDefinition()) {
         $def->addElement('include', 'Block', 'Empty', 'Common', array('file*' => 'URI', 'height' => 'Text', 'width' => 'Text'));
 		$def->addAttribute('a', 'data-toggle', 'Enum#collapse,tab');
--- a/app/Http/Middleware/ParsePostRequestFields.php
+++ b/app/Http/Middleware/ParsePostRequestFields.php
@@ -1,0 +1,38 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
+use Illuminate\Http\Request;
+
+class ParsePostRequestFields
+{
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure(\Illuminate\Http\Request): (\Illuminate\Http\Response|\Illuminate\Http\RedirectResponse)  $next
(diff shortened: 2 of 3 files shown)


app/Helpers/Helpers.php
```diff
@@ -101,7 +101,6 @@ function parse($text, &$pings = null) {
     $config->set('Attr.EnableID', true);
     $config->set('HTML.DefinitionID', 'include');
     $config->set('HTML.DefinitionRev', 2);
-	$config->set('Cache.DefinitionImpl', null); // TODO: remove this later!
```
itinerare, line 104: This *may* backfire, since I recall the cache being weirdly long-lasting... Though at this point, I suspect there are more sites using the updated settings than not!

itinerare (approved):

app/Helpers/Helpers.php
Ne-wt: yeahhh, i could see the need to keep it but I would think that `php artisan optimize` would clear that out, but could be mistaken

itinerare: I don't think it has in the past...? But it's been a long time since I've had to deal with it...

app/Http/Middleware/ParsePostRequestFields.php
```diff
@@ -0,0 +1,63 @@
...
+     */
+    public function handle(Request $request, Closure $next) {
+        if ($request->isMethod('post')) {
+            $excludedFields = ['_token', 'password', 'email', 'description', 'text'];
+            $strippedFields = ['name', 'title'];
+
+            $parsedFields = [];
+            foreach ($request->except($excludedFields) as $key => $value) {
+                if (is_array($value)) {
+                    $parsedFields[$key] = parseArray($value, $strippedFields);
```
itinerare, line 25: Needs to be `$this->parseArray`

app/Http/Middleware/ParsePostRequestFields.php
```diff
@@ -0,0 +1,63 @@
...
+     * Recursively parse array values.
+     *
+     * @param  array  $array
+     * @param  array  $strippedFields
+     * @return array
+     */
+    private function parseArray(array $array, array $strippedFields) : array {
+        foreach ($array as $key => $value) {
+            if (is_array($value)) {
+                $array[$key] = parseArray($value, $strippedFields);
```
itinerare, line 51: Likewise (`$this->parseArray`)

preimpression (approved):
This looks good! I'd be inclined to test thoroughly to make sure it isn't going to get finicky with anything, particularly thinking in the admin panel, but I trust you guys have been doing so!

itinerare (changes requested):

app/Http/Middleware/ParsePostRequestFields.php
Ne-wt: ah yes, so used to helpers

PR #1013, opened 2024-08-07
Fix Gallery Performance

AW0005:
The Gallery Index was pulling in all submissions for a gallery before counting them / only taking the top 4 which over time can cause performance issues as the number of submissions in each gallery grows.

This seemed like more of a bug than a feature so I put it on release? But feel free to let me know if I should redirect to develop.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-    "name": "lorekeeper",
+    "name": "lorekeeper-extensions",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
--- a/resources/views/galleries/index.blade.php
+++ b/resources/views/galleries/index.blade.php
@@ -51,20 +51,18 @@
                     @endif
                 </div>
                 <div class="card-body">
-                    @if ($gallery->submissions->where('status', 'Accepted')->count())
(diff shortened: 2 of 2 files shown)


itinerare:
As mentioned previously, this'd probably be best paired with some eager loading; I might go in and see what I can do (in a separate PR) about that, since I'm of the agreement that these kinds of necessary optimizations are close enough to bugfixes/worth putting on release. |D

itinerare (approved):

SpeedyD (approved):

SpeedyD:
> As mentioned previously, this'd probably be best paired with some eager loading; I might go in and see what I can do (in a separate PR) about that, since I'm of the agreement that these kinds of necessary optimizations are close enough to bugfixes/worth putting on release. |D

I'm intrigued by this, could you elaborate in this seperate PR? :)

AW0005:
@itinerare Yea honestly I _started_ to mess around a bit but I wasn't feeling sure enough about what pieces would actually make sense to do that on / would make a notable difference. With a little more spare time I'd definitely be willing to help do more with this kind of thing

itinerare:
Yeah, I figure since it's something I already have experience with, it would probably make sense for me to play around with...

PR #1019, opened 2024-08-12
feat: make comment links become clickable to prevent weird views

ScuffedNewt:
### Before
image

### After
image

### DIFF
--- a/app/Models/Comment/Comment.php
+++ b/app/Models/Comment/Comment.php
@@ -133,14 +133,15 @@ class Comment extends Model {
     }
 
     /**
-     * Returns the comment contents but with links in clickable format
-     * 
+     * Returns the comment contents but with links in clickable format.
+     *
      * @return string
      */
     public function getCommentAttribute() {
         if (config('lorekeeper.settings.wysiwyg_comments')) {
             return preg_replace('/(https?:\/\/[^\s]+)/', '<a href="$1" target="_blank">link</a>', $this->attributes['comment']);
         }
+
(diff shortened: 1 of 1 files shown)


app/Models/Comment/Comment.php
```diff
@@ -131,4 +131,17 @@ public function getTopCommentAttribute() {
...
+     * Returns the comment contents but with links in clickable format.
+     *
+     * @return string
+     */
+    public function getCommentAttribute() {
+        if (config('lorekeeper.settings.wysiwyg_comments')) {
+            return preg_replace('/(?<!\()(https?:\/\/[^\s]+)/', '<a href="$1" target="_blank">link</a>', $this->attributes['comment']);
+        }
+
+        return preg_replace('/(?<!\()(https?:\/\/[^\s]+)/', '<a href="$1" target="_blank">link</a>', $this->attributes['comment']);
```
AW0005, line 145: I might suggest using the $1 in place of the literal `link` text so that way you can still see where the link is going

ScuffedNewt: That defeats the purpose of the normalise stopping the links from breaking comment formatting, however I could add a a data-toggle on the wysiwyg comment but for the markdown I don't have ideas

ScuffedNewt: I've made it so it displays the domain instead of "link"
image

SpeedyD (approved):
I like this, and approve of it.

preimpression (approved):

PR #1031, opened 2024-08-27
chore(deps-dev): set up mix, rebuild assets

itinerare:
- update bootstrap to 4.6
- update some included JS deps
- move some included JS deps to be handled by mix
- defer loading non-critical JS and CSS
- add mix build to CI workflow, rename
- clean up JS deps
- remove redundant lockfile
- use consistent formatting for masterlist images toggle (fixes label displaying incorrectly)

Did what I've been saying I'd do for a while and set up mix for our use. There shouldn't actually be any obvious changes/existing functionality should be preserved as expected-- I used an arrangement of dependencies I've tested fairly heavily elsewhere as well, so things should be fairly good on that front.

The mix manifest is likewise committed and tracked so that everything works out of the box, and so that folks wanting to make changes that way can just run it locally/not worry about running it on the server.

I also set things up so that the linting workflow can now rebuild mix assets as needed, e.g. in the wake of dependency updates.


ScuffedNewt (approved):

ScuffedNewt:
🙏 🙏

SpeedyD (approved):

ScuffedNewt:
@SpeedyD u legend

SpeedyD:
> @SpeedyD u legend

I don't think I'm that legendary, I basically took awhile to do my research as to what all this changes before hitting approve XD

itinerare:
Hahaha, thank you both

PR #1035, opened 2024-08-31
feat(comments): Add first time flag for page load

SpeedyD:
Some users on a site I administrate, as well as admittedly I myself, found it a bit annoying that every single page with comments had the url change instantly. It made a clean link one riddled with sort parameters, when the sorting wasn't even altered.

So I figured I'd make a simple first time flag. :)

PR #1037, opened 2024-08-31
Fix multiple comment display & loading, multiple subtypes command

ScuffedNewt:
updates subtypes command to use the orm so `update` can be used
makes it so initial load on comments doesn't modify URL, since that's the default URL state anyway
allows multiple comment instances to exist on one page (gallery approval for example)

thanks to @SpeedyD for pointing these out to me!

### DIFF
--- a/resources/views/comments/comments.blade.php
+++ b/resources/views/comments/comments.blade.php
@@ -33,7 +33,7 @@
                             'newest' => 'Newest First',
                             'oldest' => 'Oldest First',
                         ],
-                        Request::get('sort') ?: 'newest',
+                        Request::get($commentType . '-sort') ?: 'newest',
                         ['class' => 'form-control', 'id' => $commentType.'-sort'],
                     ) !!}
                 </div>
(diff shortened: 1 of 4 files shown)


resources/views/comments/comments.blade.php
```diff
@@ -102,28 +106,30 @@ function sortComments() {
...
                         page: '{{ request()->query('page') }}',
                     },
                     success: function(data) {
-                        $('#comments').html(data);
+                        $('#{{ $commentType }}-comments').html(data);
                         // update current url to reflect sort change
-                        var url = new URL(window.location.href);
-                        url.searchParams.set('sort', $('#sort').val());
-                        url.searchParams.set('perPage', $('#perPage').val());
+                        if ($('#{{ $commentType }}-sort').val() != 'newest' && $('#{{ $commentType }}-perPage').val() != 5) { // don't add to url if default
```
ScuffedNewt, line 116: https://github.com/corowne/lorekeeper/pull/1035 uses a first time flag which would allow the URL to be updated back to the original after fiddling, so this could be add as well on top now that I look again

itinerare: Whatever you feel is best there tbh

itinerare (approved):

SpeedyD (approved):

itinerare (approved):

PR #1038, opened 2024-08-31
hotfix(subtypes): update multiple subtypes command

ScuffedNewt:
i dont know if this is needed waiting on @SpeedyD


app/Console/Commands/ConvertCharacterSubtype.php
```diff
@@ -50,7 +50,7 @@ public function handle() {
             // make the string into an array
             foreach ($updates as $update) {
                 $update->update([
-                    'subtype_ids' => $update->subtype_id,
+                    'subtype_ids' => [$update->subtype_id],
```
SpeedyD, line 53: This is **not** the move. Results in `[null]` in the database if there's data.

PR #1039, opened 2024-08-31
temp(characters): Need to figure out the problem with filesize validation

SpeedyD:
After hounding @ScuffedNewt about this on #959, it is discovered that Laravel gives a nice `The given value "2mb" does not represent a valid number.` error.

Despite that Laravel documentation at https://laravel.com/docs/10.x/validation#validating-files-file-sizes says that 2mb is a very valid number.

This PR is a temporary hotfix to fix the 'does not represent a valid number' issue, but we need to look into _why_ 2mb isn't accepted and roll this PR back eventually.

### DIFF
--- a/app/Models/Character/Character.php
+++ b/app/Models/Character/Character.php
@@ -78,8 +78,8 @@ class Character extends Model {
         'slug'                  => 'required|alpha_dash',
         'description'           => 'nullable',
         'sale_value'            => 'nullable',
-        'image'                 => 'required|mimes:jpeg,jpg,gif,png|max:2mb',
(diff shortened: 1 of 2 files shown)


ScuffedNewt:
wow that sucks lmao

ScuffedNewt:
commenting to add its likely because we are using the string validation rather than the `min()` and `max()` functions listed in the doc

ScuffedNewt (approved):

itinerare:
ahhhhhh maybe so

itinerare (approved):

PR #1042, opened 2024-09-01
fix(features): update searchUrl

ScuffedNewt:


itinerare:
Wasn't this fixed in #1008 ?

PR #1047, opened 2024-09-02
feat(commands): improve second subtype support in subtype conversion command

itinerare:
- make second subtype support work out of the box
- add additional info text around design updates
- add a progress bar for design updates

Technically emerged as a bugfix, but took the opportunity to put some polish on it and head off some future headaches while I was at it. Figured if there was already implicit support, _explicit_/out-of-the-box support was probably the better move and the one that keeps things simpler on the support front.

Tested both with and without second subtype columns, so shouldn't cause issues either way.

### DIFF
--- a/app/Console/Commands/ConvertCharacterSubtype.php
+++ b/app/Console/Commands/ConvertCharacterSubtype.php
@@ -38,61 +38,90 @@ class ConvertCharacterSubtype extends Command {
      */
     public function handle() {
         if (Schema::hasColumn('character_images', 'subtype_id')) {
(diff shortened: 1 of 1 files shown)


SpeedyD (approved):

ScuffedNewt (approved):
should this also include third subtype extension or do we think no?

SpeedyD:
> should this also include third subtype extension or do we think no?

...Tbh Merc and I unanimously decided to do that in a seperate PR. XD

itinerare:
I was thinking it made the most sense to get this in asap/make sure what's on the branch is known working, first, but otherwise yeah it's probably for the best

PR #1048, opened 2024-09-02
feat(commands): add third subtype support in subtype conversion command

SpeedyD:
Based on #1047 that coincidentally passed just minutes ago!

- make _third_ subtype support work out of the box

That's all!

### DIFF
--- a/app/Console/Commands/ConvertCharacterSubtype.php
+++ b/app/Console/Commands/ConvertCharacterSubtype.php
@@ -40,9 +40,9 @@ class ConvertCharacterSubtype extends Command {
         if (Schema::hasColumn('character_images', 'subtype_id')) {
             // Check for second subtype columns up front to save on some redundant checks
             $secondSubtype = false;
+            $thirdSubtype = false;
             if (Schema::hasColumn('character_images', 'subtype_id_2') && Schema::hasColumn('design_updates', 'subtype_id_2')) {
                 $secondSubtype = true;
-                $thirdSubtype = false;
                 if (Schema::hasColumn('character_images', 'subtype_id_3') && Schema::hasColumn('design_updates', 'subtype_id_3')) {
                     $thirdSubtype = true;
                 }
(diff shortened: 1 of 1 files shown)


itinerare:
Wow! I would have never expected this very PR!

app/Console/Commands/ConvertCharacterSubtype.php
```diff
@@ -42,11 +42,17 @@ public function handle() {
             $secondSubtype = false;
             if (Schema::hasColumn('character_images', 'subtype_id_2') && Schema::hasColumn('design_updates', 'subtype_id_2')) {
                 $secondSubtype = true;
+                $thirdSubtype = false;
```
itinerare, line 45: This should be set above alongside `$secondSubtype`  (or similar) so that it's always set and doesn't cause errors from it... not being set...

app/Console/Commands/ConvertCharacterSubtype.php
```diff
@@ -42,11 +42,17 @@ public function handle() {
...
+                if (Schema::hasColumn('character_images', 'subtype_id_3') && Schema::hasColumn('design_updates', 'subtype_id_3')) {
```
itinerare, line 46: Is it safe to assume that these columns will only be present if the second subtype ones are? I assume yes, but...

itinerare (changes requested):
Just the one significant note... otherwise looks good!

app/Console/Commands/ConvertCharacterSubtype.php
SpeedyD: D'oh! See, I started with this part, and I thought later "oh yeah, I should fix that one part in the beginning"...

..and then forgot. Lemme fix that quick.

app/Console/Commands/ConvertCharacterSubtype.php
SpeedyD: Third Subtype is DIRECTLY built off of Second Subtype. It's literally your code + mine.

So if SecondSubtype columns are missing, the error lies with the site tech, not the code.

itinerare: I figured, but wanted to double-check, just in case |D

itinerare (approved):

ScuffedNewt (approved):

PR #1051, opened 2024-09-04
fix(subtypes): stop error on slots

ScuffedNewt:


SpeedyD:
.....Um, sorry, but what does this actually fix? Seems more like it's removing functionality.

ScuffedNewt:
it fixes the slot tag 
if you view the code the subtypes are treated differently further done, this code is unneeded since the subtype IDs are not contained in the character image

ScuffedNewt:
you can test it yourself to see, but I had a few big reports across multiple sites - this has fixed it

SpeedyD:
Right, fair enough. I think this error is primarily caused by the fact that that slot tags haven't been converted as did Design Updates and Character images, though new & edited slots shouldn't have this issue altogether.

On the other hand, the code does seem superfluous, as it's a relation now. It may be an issue for sites who have hard-coded a check for which image to use on subtypes, but that requires at most a check to fix.

SpeedyD (approved):

itinerare (approved):

PR #1071, opened 2024-09-23
prevent throttling on different links

ScuffedNewt:
adds a url to the throttle key so users arent limited from preforming different actions
also adds an exemption to admin routes

**why?**
- this feature is intended to stop spamming, a blanket limit (especially at 10 seconds) can be inhibiting on a click / request heaving application like lorekeeper
- admin routes have similar reasoning to the above, and are at lower risk of spamming being an issue


itinerare (approved):

SpeedyD (approved):
To be honest, this is a necessary and welcome change cause I tend to open multiple things in tabs and rapid fire approve when I'm doing approvals... _and I think you can guess the problem._

So thanks, Newt!

PR #1073, opened 2024-09-25
feat(optimisation): chunk image query for content warning tags

ScuffedNewt:
chunks the image query so not everything is loaded in memory

### DIFF
--- a/app/Http/Controllers/BrowseController.php
+++ b/app/Http/Controllers/BrowseController.php
@@ -397,57 +397,56 @@ class BrowseController extends Controller {
             });
         }
         if ($request->get('excluded_tags') || $request->get('included_tags')) {
-            $excludedTags = $request->get('excluded_tags');
-            $includedTags = $request->get('included_tags');
-            $images = $imageQuery->get();
-
-            $filteredImageQuery = $images;
-
-            // first exclude, then include
-            if ($excludedTags) {
-                $filteredImageQuery = $images->filter(function ($image) use ($excludedTags) {
-                    if (!$image->content_warnings) {
-                        return true;
-                    }
-
-                    if (in_array('all', $excludedTags) && $image->content_warnings) {
-                        return false;
-                    }
-
-                    foreach ($excludedTags as $tag) {
-                        if ($image->content_warnings && in_array($tag, $image->content_warnings)) {
+            $filteredImageIds = collect();
+            $excludedTags = $request->get('excluded_tags', []);
+            $includedTags = $request->get('included_tags', []);
+        
+            $imageQuery->chunk(100, function ($images) use (&$filteredImageIds, $excludedTags, $includedTags) {
+                // excluded tags
+                if (!empty($excludedTags)) {
+                    $images = $images->reject(function ($image) use ($excludedTags) {
+                        if (!$image->content_warnings) {
                             return false;
                         }
-                    }
-
-                    return true;
-                });
-            }
-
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/BrowseController.php
```diff
@@ -397,57 +397,56 @@ private function handleMasterlistSearch(Request $request, $query, $imageQuery, $
...
-                        return false;
-                    }
-
-                    foreach ($excludedTags as $tag) {
-                        if ($image->content_warnings && in_array($tag, $image->content_warnings)) {
+            $filteredImageIds = collect();
+            $excludedTags = $request->get('excluded_tags', []);
+            $includedTags = $request->get('included_tags', []);
+
+            $imageQuery->chunk(100, function ($images) use (&$filteredImageIds, $excludedTags, $includedTags) {
```
SpeedyD, line 404: This line is missing a closing tag, isn't it?

And from what I can tell from Laravel Docs, chunk is just.. chunk(#number#), not a whole line of extra text..?

ScuffedNewt: nope! not missing a closing tag - you should be able to test by running (and it not failing haha)

you're thinking of chunk on an eloquent without applying filtering, since we're filtering the chunk we need to pass the variables to the lambda function (which is the `function ($images)...` part)

chunking the images to 100 is because the entire query must be executed before the pagination is returned, there is no benefit from chunking only 24 since the entire query must be iterated through anyway!

ScuffedNewt: i saw your edit re the opens 3 and closes 2, the closing line occurs further down after the lambda filtering (should be like `});`

SpeedyD: > you're thinking of chunk on an eloquent without applying filtering, since we're filtering the chunk we need to pass the variables to the lambda function (which is the `function ($images)...` part)

I sure am. I couldn't find anything like this on the Laravel docs page.. but an expanded Google search showed me plenty of people using it like this. Sorry about that.
 
> chunking the images to 100 is because the entire query must be executed before the pagination is returned, there is no benefit from chunking only 24 since the entire query must be iterated through anyway!

Right, now you mention it- it would not show page 2 if you only have 24; or worse, if your search result ends up finding multiple images of a single character, you may not even have 24 results on the first page..
Does chunking it like this work well for future pages, though?

> nope! not missing a closing tag - you should be able to test by running (and it not failing haha)

> i saw your edit re the opens 3 and closes 2, the closing line occurs further down after the lambda filtering (should be like `});`

Ah, on line 444, right? Yeah, I missed that, sorry.

I probably should be looking into chunk more. The way I've seen it used before was primarily in blades, so I'm likely missing some entire possibility I'm unaware of.

ScuffedNewt: > Right, now you mention it- it would not show page 2 if you only have 24; or worse, if your search result ends up finding multiple images of a single character, you may not even have 24 results on the first page..
Does chunking it like this work well for future pages, though?

nope! basically what its doing is processing the query in "chunks" as to not take up as much memory! so basically by doing 100 at a time, we are iterating slightly faster than in blocks of 24, but we use a bit more memory
let me know if that makes sense!

SpeedyD (approved):
Can't believe I forgot to approve this afterward, sorry Newt

itinerare (approved):

PR #1079, opened 2024-10-03
fix(content warnings): included warnings not querying correctly

ScuffedNewt:
edit: should be tested on clean copy but I believe this is correct


SpeedyD (approved):
..Logically, this seems good to me. Safer bet than the in_array().

AW0005 (approved):

itinerare:
Needs conflict resolution, but otherwise should be good to go.

PR #1093, opened 2024-10-26
Bug(DesignManager): $subtype doesn't exist if unspecified

AW0005:
Fix error occurring when subtype isn't specified at all in a design request.

### DIFF
--- a/app/Services/DesignUpdateManager.php
+++ b/app/Services/DesignUpdateManager.php
@@ -391,9 +391,6 @@ class DesignUpdateManager extends Service {
             if (!$species) {
                 throw new \Exception('Invalid species selected.');
             }
-            if (isset($subtype) && $subtype && $subtype->species_id != $species->id) {
-                throw new \Exception('Subtype does not match the species.');
-            }
 
             // Clear old features
             $request->features()->delete();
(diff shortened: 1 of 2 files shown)


app/Services/DesignUpdateManager.php
```diff
@@ -391,7 +391,7 @@ public function saveRequestFeatures($data, $request) {
             if (!$species) {
                 throw new \Exception('Invalid species selected.');
             }
-            if ($subtype && $subtype->species_id != $species->id) {
+            if (isset($subtype) && $subtype && $subtype->species_id != $species->id) {
                 throw new \Exception('Subtype does not match the species.');
```
SpeedyD, line 395: This may be a nitpick, but.. if you add isset, can't the original '$subtype' check be removed? Think that's why it was originally there, for a check if it exists..

SpeedyD: WAIT A SECOND HANG ON I'm an idiot

This ENTIRE CHUNK is COMPLETELY UNNECESSARY!! The subtype and species matching is happening way earlier!!!

AW0005: Yea looks like you're right - updated to completely remove the old check

SpeedyD (approved):

ScuffedNewt (approved):

PR #1098, opened 2024-10-29
fix(galleries): more efficient queries

AW0005:
Making a couple more gallery related queries more efficient.

### DIFF
--- a/resources/views/galleries/gallery.blade.php
+++ b/resources/views/galleries/gallery.blade.php
@@ -26,7 +26,7 @@
         </p>
     @endif
     <p>{!! $gallery->description !!}</p>
-    @if (!$gallery->submissions->count() && $gallery->children->count() && $childSubmissions->count())
+    @if (!$gallery->submissions()->count() && $gallery->children->count() && $childSubmissions->count())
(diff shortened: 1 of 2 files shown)


SpeedyD (approved):

resources/views/galleries/gallery.blade.php
```diff
@@ -26,7 +26,7 @@
         </p>
     @endif
     <p>{!! $gallery->description !!}</p>
-    @if (!$gallery->submissions->count() && $gallery->children->count() && $childSubmissions->count())
+    @if (!$gallery->submissions()->count() && $gallery->children->count() && $childSubmissions->count())
```
itinerare, line 29: What's the reasoning here?

AW0005: Without the parenthesis it's pulling all submissions into memory to count them instead of letting sql count them

itinerare: I thought it was the reverse? Or is there some particular nuance of relation-vs-query I'm missing here

AW0005: Nope, not the reverse:
https://laravel.com/docs/10.x/eloquent-relationships#querying-relations

itinerare: Ahhhh I see

itinerare (approved):

PR #1100, opened 2024-10-29
feat(fontawesome): fontawesome update & ability to change between versions

### LINKED ISSUE #888: FontAwesome V6
While we're on the subject of upgrading- FontAwesome also comes to mind. V6 has been out for quite a while..

But we're still on V5 nonetheless.

I found this document on migrating to v5: (https://docs.fontawesome.com/web/setup/upgrade).

Fortunately, it seems that upgrading THAT is less of a hassle, as they heavily support backwards compatibility.

liwoyadan:
Finally getting to #888 ...sorry this is late!

With this, the webfont files for v5 and v6 versions of FontAwesome are contained within their own subfolders within `public/webfonts`. FontAwesome v5's original `all.min.css` has been renamed to be `allv5.min.css`, and `allv6.min.css` and `allvmix.min.css` have been added. Site owners can freely choose between these three stylesheets for their FontAwesome. I just shoved in a new helper class that will return the correct asset URL based on what the setting is.

I've just made it a site setting for ease of editing, so `php artisan add-site-setting` needs to be run after this is pulled.
- **0** makes the site use FontAwesome v5 only (this means any v6 classed icons will not work)
- **1** makes the site use FontAwesome v6 only (and it automatically supports v5 classes, so all v5 classed icons will show the v6 version)
- **2** makes the site use a mixed v5-v6 FontAwesome version: this means icons with v5 classes (i.e. `fas fa-crown`) will show their v5 versions, while icons with v6 classes (i.e. `fa-solid fa-crown`) will show the v6 versions

Here are visual previews of each of the settings:
A preview of setting 0, where the icons are only v5
A preview of setting 1, where the icons are only v6
A preview of setting 2, where the icons are a mix of v5 and v6

If there's any people out there running the pro versions of FontAwesome instead and want a mixed version of pro compiled, feel free to give me a shout and I can get it done pretty quickly. (Bless FontAwesome coming with all their sass files, hahaha...)


SpeedyD (approved):
Seems good to me.

I do want to point out though: If you do get someone's Pro files, do **NOT** PR them for Lorekeeper. That's a licensing disaster waiting to happen.
I'd strongly suggest making it a form of extension instead, offering it for those who have Pro. (And make it very clearly that people without shouldn't get it due to licensing issues.)
Though re-reading your message, you're likely already aware of it and are just offering it to those who happen to read your message.. ^^;

itinerare:
Oh *very* good work... thank you for taking the time to do this!

app/Console/Commands/AddSiteSettings.php
```diff
@@ -87,6 +87,8 @@ public function handle() {
 
         $this->addSiteSetting('can_transfer_items_directly', 1, 'Whether or not users can directly transfer items to other users without trading. 0: Users cannot directly transfer items. 1: Direct item transfers are allowed.');
 
+        $this->addSiteSetting('fontawesome_version', 0, 'What version of FontAwesome the site uses. 0: Version 5. 1: Version 6. 2: A mixed version where icons with v5 classes (i.e. fas) show the v5 icons and icons with v6 classes (i.e. fa-solid) show the v6 icons.');
```
itinerare: I think the one thing I'm thinking on is whether or not this should be a config setting instead...

SpeedyD: Gonna be honest, that _did_ hit my mind as well, but.. kinda wondering _how important_ the distinction is, cause we got a lot of settings in site settings as-is that in my personal opinion should be in config..

Like, I thought the distinction was Site settings is stuff that you may need to sometimes switch, whereas config is stuff you set once and then could forget forever.. 🤔

itinerare: Yeah, that's broadly the logic-- though it's a little more nuanced than that, e.g. to some degree it makes sense to have settings that relate to game function in site settings, sometimes, even if not changed very frequently (or at all). 

In this case, my vague feeling is that, because this deals with the function of the site on a mechanical level/is associated with how underlying assets are handled and loaded, it probably makes more sense to live with... the rest of the code...

ScuffedNewt (approved):
approving with the note that i think this is better in config like mercury stated

liwoyadan:
Sure thing! I had originally opted for site setting just because I know a number of site owners who are more comfortable decorating pages and site through the admin panel/TinyMCE editor and such but avoid touching any actual code files, but a config setting is no problem at all. Changed it!

Also no worries on the pro stuff haha, I've already had to tell a few people off for uploading my pro kits to public repositories. >< I just know there's a couple of LK site owners around who have pro kits through seats on my own pro subscription, so I figured I'd throw out the offer to compile just the css for it (they would still require the correct pro webfont files themselves) in case any of them were reading this PR and wanted the option. :-)

itinerare:
Much appreciated!

itinerare (approved):

PR #1111, opened 2024-11-05
feat(jquery-ui): allow jquery-ui mouse interactions to be used on touch screens

### LINKED ISSUE #1110: Adding drag-drop-touch for mobile sorting
[https://www.npmjs.com/package/drag-drop-touch](https://www.npmjs.com/package/drag-drop-touch)

I've had this package manually added to one of the JS folders for a while and it works pretty well on to fix touchscreens not liking the drag-and-drop in the character sort page (you add `draggable="true"` to the div under the foreach in `characters.blade.php` in home) 

Maybe it can be added to dev or v3? I don't know which JS folder is best to put it in or how to hook up packages installed via NPM myself though (like bootstrap) if that'd be more efficient than my temp implementation though

liwoyadan:
Tossing this into dev upon suggestion in #1110. jQuery UI's interactable widgets (i.e. draggable, resizable, selectable, etc...but most prominently for Lorekeeper's use: sortable) relies on mouse events and thus won't register on touchscreen devices such as mobile phones or iPads. This addition includes the [furf/jquery-ui-touch-punch](https://github.com/furf/jquery-ui-touch-punch) library that, as long as it's included on the page where the widget is, will allow touchscreen touch events to interact with and manipulate the widget as well. It works automatically and you do not need to adjust any further code.

There's a write up on [the extension page I made](http://wiki.lorekeeper.me/index.php?title=Extensions:Touchscreen_Sortable) but both I and several others have tested this across multiple different Android, iPhone, iPad, and few other touchscreen devices and it works smoothly. The only remark I've seen of it not working is someone in a Stack Overflow comment from 2014 mentioning it wasn't working on mobile Internet Explorer on their Windows Phone, but given that it's 10 years from then now, IE Mobile lost support in 2023, and the Windows Phone's final release was in 2015...that's at least not likely to be an issue that will be run into nowadays.

Touch Punch is dual licensed under MIT and GPL Version 2 licenses and can be freely used, modified, and distributed but in the case of the last option it does need to be attributed. The author requests that the [Touch Punch website](http://touchpunch.furf.com/) be linked, but it seems that the website had gone down sometime in the last few months, so instead I've just linked directly to the repository as credit. You _can_ access the website through the Wayback Machine, though, and you can find a preview of it [here](https://web.archive.org/web/20160827114740/https://touchpunch.furf.com/) if you want to try testing the touch interactions on your own mobile device. (Though the site may not load everything very smoothly, given that it's just a capture of it from the past.)

One additional feature I have included in this is specifically on the `/characters` (My Characters) and `/character/{slug}/images` (Character Images) pages: there's just a little bit of additional JavaScript. It will attempt to create a touch event on the user's device to detect if it's a touchscreen or not. If it does detect a touchscreen, it will clone and append a round drag handle icon on each sortable element within the overall sortable and set it as the sortable's handle, making users touch and drag that icon specifically to reorder their characters/character images. The styling for this drag handle can be adjusted in `lorekeeper.css`. 

This feature is particularly for people on mobile phones - so they can touch and scroll the page without triggering the sortable function, given that on mobile viewports the sortable essentially takes up the entire width of the screen. These drag handles are also triggered if you are on browser developer tools with touch simulation enabled.

A screenshot of the drag handle on the /characters page visible on touch devices.

All you need to do is pull and merge; there are no additional commands necessary. For anyone reading this who wants this functionality on their own site but aren't on the dev branch, I will still have the v3 branch up and it is linked [on the extension page](http://wiki.lorekeeper.me/index.php?title=Extensions:Touchscreen_Sortable).

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,3 +75,7 @@ You will need to send yourself the verification email and then link your social 
 ## Contact
 
 If you have any questions, please feel free to ask in the Discord server: https://discord.gg/U4JZfsu
+
+## Credits
+
+This minor extension uses the [furf/jquery-ui-touch-punch](https://github.com/furf/jquery-ui-touch-punch) library to enable touch on jQuery UI elements.
--- a/README.md
+++ b/README.md
@@ -75,7 +75,3 @@ You will need to send yourself the verification email and then link your social 
 ## Contact
 
 If you have any questions, please feel free to ask in the Discord server: https://discord.gg/U4JZfsu
-
-## Credits
-
-The [furf/jquery-ui-touch-punch](https://github.com/furf/jquery-ui-touch-punch) library is used to enable touch on jQuery UI elements.
--- a/public/css/lorekeeper.css
+++ b/public/css/lorekeeper.css
@@ -68,6 +68,17 @@ body {
     height: auto !important;
 }
 
+.mobile-handle {
+    font-size: 1.75rem;
+    position: absolute;
+    top: -1rem;
+    right: 0;
+}
+
+.mobile-handle i {
+    transform: rotate(-25deg);
+}
+
 /**************************************************************************************************
 
     Layout
--- a/public/js/jquery.ui.touch-punch.min.js
+++ b/public/js/jquery.ui.touch-punch.min.js
@@ -1,0 +1,11 @@
+/*!
+ * jQuery UI Touch Punch 0.2.3
+ *
+ * Copyright 2011–2014, Dave Furfero
+ * Dual licensed under the MIT or GPL Version 2 licenses.
+ *
+ * Depends:
+ *  jquery.ui.widget.js
+ *  jquery.ui.mouse.js
+ */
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -75,3 +75,7 @@ You will need to send yourself the verification email and then link your social
 ## Contact
 
 If you have any questions, please feel free to ask in the Discord server: https://discord.gg/U4JZfsu
+
+## Credits
+
+The [furf/jquery-ui-touch-punch](https://github.com/furf/jquery-ui-touch-punch) library is used to enable touch on jQuery UI elements.
```
SpeedyD: I don't think this is a good idea.. Why not just add it to `/resources/views/pages/credits.blade.php`, alongside credits for yourself?

SpeedyD:
Also, how about sortables in the admin section? Trait category and rarity comes to mind..

liwoyadan:
I can change where the credits are but I'm not sure what you mean by the admin section comment. As mentioned, they're automatically affected? Nothing further needs to be done as long as the script is included in `app.blade.php` or otherwise on whatever page the jQuery UI widget is located.

SpeedyD:
Ah.. right, sorry, I was only skimming the code earlier, so I hadn't realized all that happens in the character section is cloning that hand image. My bad entirely.

But please edit the credits. :)

liwoyadan:
And done! 👍

itinerare (approved):
Thanks as always for taking the time to do this!

ScuffedNewt (approved):

PR #1112, opened 2024-11-05
feat(news/sales): Allow admin with manage rights to view news/sales that arent visible

SpeedyD:
Let us preview the news and sales! We're admin! We got rank powers!

..Basically, allows users with the manage_news rank power to view news posts while they're not viewable
and likewise allows users with the manage_sales rank power to view sales posts while they're not viewable.

Adds a 'View Post' button to both create_edit pages when the post has an id, that links to said post.
And of course, my traditional fa-eye-slash icon for those invisible items.

...Only question I had was.. should I also add non-visible posts to the sidebar for admin only? Or.. just.. leave 'em only accessible through the admin area..?

### DIFF
--- a/app/Http/Controllers/NewsController.php
+++ b/app/Http/Controllers/NewsController.php
@@ -45,17 +45,11 @@ class NewsController extends Controller {
      * @return \Illuminate\Contracts\Support\Renderable
      */
     public function getNews($id, $slug = null) {
-        $news = News::where('id', $id)->first();
+        $news = News::where('id', $id)->visible(Auth::user() ?? null)->first();
         if (!$news) {
             abort(404);
         }
 
-        if (!$news->is_visible) {
-            $user = Auth::user() ?? null;
-            if ($user == null || !$user->hasPower('manage_news')) {
-                abort(404);
-            }
-        }
 
         return view('news.news', ['news' => $news]);
     }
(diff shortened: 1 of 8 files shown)


app/Http/Controllers/NewsController.php
```diff
@@ -45,11 +45,18 @@ public function getIndex() {
...
         if (!$news) {
             abort(404);
         }
 
+        if (!$news->is_visible) {
+            $user = Auth::user() ?? null;
+            if ($user == null || !$user->hasPower('manage_news')) {
+                abort(404);
+            }
+        }
```
itinerare: I think at this point it'd be better just to adjust the existing `visible()` scope for news to accommodate this...

app/Http/Controllers/SalesController.php
```diff
@@ -83,11 +83,18 @@ public function getIndex(Request $request) {
...
         if (!$sales) {
             abort(404);
         }
 
+        if (!$sales->is_visible) {
+            $user = Auth::user() ?? null;
+            if ($user == null || !$user->hasPower('manage_sales')) {
+                abort(404);
+            }
+        }
```
itinerare: Likewise for sales.

itinerare:
I think essentially just having a preview button on the admin end is a fine solution to this, personally.

SpeedyD:
This is what you meant, right, @itinerare? I have no idea why I didn't do it like this in the first place . . .

itinerare (approved):
Yep! Exactly.

ScuffedNewt (approved):

Issue #1113, opened 2024-11-06
Making items use actual Rarities rather than arbitrary numbering


ScuffedNewt:
as the title states

ScuffedNewt:
examples can be seen on item variants and a few other extensions that benefit from the use of rarities, plus I think logically it makes most sense for rarities to be consistent across the board

itinerare:
On the other hand, sites may not use the same rarity systems between characters/traits and items-- so any solution for this needs to accommodate that possibility as well as allow conditional loot tables to work (i.e. still correlate rarities to numeric values).

ScuffedNewt:
I think this should be solved on the rarity side, though, by making subcategories or divisions rather than on the item side. It is disruptive to the flow of how other things work especially considering most extensions use rarities as-is rather than creating a new system which ideally is how it should be

itinerare:
Oh, I'm not arguing for any particular resolution-- I'm just saying that the kind of functionality that's presently there needs to be maintained one way or another. Making the rarity system more robust feels like a reasonable way about it, though.

ScuffedNewt:
do you prefer if the rarity subcategories go in a new PR or as a separate issue to be discussed on how best to approach it?

itinerare:
Just directly to new PR might be best, unless you have specific questions regarding the implementation?

ScuffedNewt:
not really honestly, perhaps the only thing would be having varying searches based on the categorisations
I'll mull it over since I won't get to it for a few hours for sure anyway

itinerare:
Yeah, very fair. Sounds good, anyhow

PR #1114, opened 2024-11-06
Feat: allow items to use built-in rarities instead of individual system

ScuffedNewt:
#1113 

tested loot conditionals etc also


itinerare:
I think my one worry is that it might not be obvious what correlates to what wrt conditional loot tables, and/or that, if it's by sort, adjusting the sort without thinking to adjust the loot tables could have unintended adverse affects. Which is to say, I don't think tying it to something that can change without the end user really realizing it is wise....

ScuffedNewt:
I think that a small alert appearing next to the sort of any conditional tables exist could be a decent compromise, I think an alert on loot tables may go unnoticed by themselves but with both it could be a decent solution

otherwise do you have another idea for how to let the site admins know ?

itinerare:
Displayed when sorting rarities, you mean?

ScuffedNewt:
yeah I meant if any conditional table not of lol

itinerare:
Haha, yeah, gotcha.
Anyhow, that would probably work-- agreed that I don't think an alert by loot tables would do much.... 

I think it'd also be necessary to display the sort value for each rarity somewhere (perhaps also on the rarity index for quick/easy ref) so that admins know what to set conditional table value(s) to.

ScuffedNewt:
they don't have to set it to a number, they set it to the rarity itself

ScuffedNewt:
I think you misunderstood the setup! in which case I think the sort alert is also unneeded ?

itinerare:
Ahhh, I was remembering conditional loot tables just having you punch in a number. All good, then.

itinerare (approved):

itinerare:
That being said, I don't think just adding a small note of [changing sort order significantly may impact conditional tables] or somesuch would go amiss.

SpeedyD (approved):
Yeah, I'm gonna be honest, never understood why there was a numeric rarity system for items. I approve, vastly, of this change.

itinerare:
Realized that there needs to be some means of shuffling existing item rarities over to the new setup as well... ideally one that's reasonably easy to adjust.

SpeedyD (approved):

itinerare (approved):
Oh, I like that solution for converting existing rarities a *lot*, well done.

PR #1119, opened 2024-11-10
fix: remove extra $ from _profile_content.blade.php

ScuffedNewt:


itinerare (approved):

SpeedyD:
Augh, sorry about that. Overlooked it. 🙇

PR #1124, opened 2024-11-11
fix criteria for ItemRarity and ItemCategoryRarity loot tables

rattusquo:
updated rarity-based loot tables to pass '=' criteria as '==' to resolve an error in evaluating sort order. thank you to Newt for pointing out this fix :) please note that any existing loot tables with this criteria will need to be updated in the admin panel, as this fix resets their criteria field.


ScuffedNewt:
ideally this would be retroactive but what can you do!

ScuffedNewt (approved):

itinerare:
For my part, I'd definitely ask for this either to not break existing tables or for existing tables to be handled via some means (conversion command, etc.); it's a must one way or another.

SpeedyD (approved):
I thought I approved this MONTHS ago O_o

itinerare (changes requested):
Formalizing my earlier comment as a review ~~mostly so the PR is flagged appropriately and I don't keep wandering in here having forgotten what's up~~...
> For my part, I'd definitely ask for this either to not break existing tables or for existing tables to be handled via some means (conversion command, etc.); it's a must one way or another.

itinerare:
Closed due to inactivity/in favor of #1191.

PR #1125, opened 2024-11-11
item rarity searches

rattusquo:
added dropdowns for item rarity filtering to various areas of the site. search dropdown is contained within the config check for item entry expansion's extra fields alongside artist, since item rarity only applies when that portion of the ext is enabled :)


itinerare (approved):

SpeedyD:
I like this, I in fact spoke with Newt the other day that we should add this, haha..

I just have ONE question: Could you also add a "No Rarity" option? If not, I can do that later, no worries. But since Item Rarity is _optional_, it'd be nice for folk to be able to get results that specifically don't have a rarity. :)

rattusquo:
good call, should be updated now :)

itinerare (approved):

app/Http/Controllers/Admin/Data/ItemController.php
```diff
@@ -210,6 +217,7 @@ public function getItemIndex(Request $request) {
             'items'      => $query->paginate(20)->appends($request->query()),
             'categories' => ['none' => 'Any Category'] + ItemCategory::orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
             'artists'    => User::whereIn('id', Item::whereNotNull('artist_id')->pluck('artist_id')->toArray())->pluck('name', 'id')->toArray(),
+            'rarities'   => ['none' => 'Any Rarity'] + ['no_rarity' => 'No Rarity'] + Rarity::orderBy('rarities.sort', 'DESC')->pluck('name', 'id')->toArray(),
```
ScuffedNewt: instead of having `['none' => 'Any Rarity']` please instead use a `placeholder` on the form

app/Http/Controllers/Admin/Data/ItemController.php
```diff
@@ -176,6 +176,13 @@ public function getItemIndex(Request $request) {
         if (isset($data['artist']) && $data['artist'] != 'none') {
             $query->where('artist_id', $data['artist']);
         }
+        if (isset($data['rarity_id']) && $data['rarity_id'] != 'none') {
```
ScuffedNewt: using placeholder means you do not have to check `!= none`

resources/views/character/inventory.blade.php
```diff
@@ -41,6 +41,9 @@
                 {!! Form::select('item_category_id', $categories->pluck('name', 'id'), Request::get('item_category_id'), ['class' => 'form-control', 'placeholder' => 'Any Category']) !!}
             </div>
             @if (config('lorekeeper.extensions.item_entry_expansion.extra_fields'))
+                <div class="form-group ml-3 mb-3">
+                    {!! Form::select('rarity_id', $rarities, Request::get('rarity_id'), ['class' => 'form-control']) !!}
```
ScuffedNewt: `['class' => 'form-control', 'placeholder' => 'Any Rarity']`

ScuffedNewt (changes requested):

rattusquo:
done, thanks Newt! also brought artist/category selections in line with this change where relevant.

ScuffedNewt (approved):
Awesome!! thank you!

app/Http/Controllers/Users/InventoryController.php
```diff
@@ -69,6 +77,7 @@ public function getIndex(Request $request) {
             'userOptions' => User::visible()->where('id', '!=', Auth::user()->id)->orderBy('name')->pluck('name', 'id')->toArray(),
             'user'        => Auth::user(),
             'artists'     => ['none' => 'Any Artist'] + User::whereIn('id', Item::whereNotNull('artist_id')->pluck('artist_id')->toArray())->pluck('name', 'id')->toArray(),
```
itinerare: This pseudo-placeholder probably should be taken out...

rattusquo: missed that one, whoops

itinerare (approved):

SpeedyD:
It's already merged in, but I still wanted to say I heavily approve of this, and I thank you for taking the time to do this, @rattusquo! :)

PR #1135, opened 2024-11-24
feat(pages): Allow admin with manage rights to view pages that arent visible

SpeedyD:
Let us preview the pages! We're admin! We got rank powers!.. wait, I feel like I said this before.

Anyway, allows users with the edit_pages rank power to view pages while they're not viewable.

And of course, my traditional fa-eye-slash icon for those invisible items.

...Feels like I should've done this at the same time I added the functionality for sales and news posts, huh..

### DIFF
--- a/app/Models/SitePage.php
+++ b/app/Models/SitePage.php
@@ -56,7 +56,7 @@ class SitePage extends Model {
         SCOPES
     **********************************************************************************************/
     /**
-     * Scope a query to only include visible posts.
+     * Scope a query to only include visible pages.
      *
      * @param \Illuminate\Database\Eloquent\Builder $query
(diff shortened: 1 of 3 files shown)


app/Models/SitePage.php
```diff
@@ -52,6 +52,28 @@ class SitePage extends Model {
         'text'  => 'nullable',
     ];
 
+    /**********************************************************************************************
+        SCOPES
+    **********************************************************************************************/
+    /**
+     * Scope a query to only include visible posts.
```
itinerare: Should be switched to pages, but that's minor, all told.

itinerare (approved):

AW0005 (approved):

PR #1143, opened 2024-12-01
fix(commands): Currency icon fails to transfer correctly due to hash being changed way before it gets to the currency icons

SpeedyD:
See title.


ScuffedNewt (approved):

SpeedyD:
Hang on, there's another issue..

SpeedyD:
There we go.

ScuffedNewt (approved):

itinerare (approved):

PR #1155, opened 2024-12-14
fix(comments): Stops replacement regex at any opening tag

SpeedyD:
Found out earlier that links, pre-formatted or by the user, cause issues due to the regex, as it includes the ending </a> tag.

example made using regexr

With my suggested change, this no longer occurs.

example made using regexr 2

RegEx check done on https://regexr.com/, and yes, double checked with https://regex101.com/ and a LIVE SITE.

### DIFF
--- a/app/Models/Comment/Comment.php
+++ b/app/Models/Comment/Comment.php
@@ -153,7 +153,7 @@ class Comment extends Model {
     public function getCommentAttribute() {
         if (config('lorekeeper.settings.wysiwyg_comments')) {
             return preg_replace_callback(
-                '/(?<!href=")(?<!src=")(?<!\()(https?:\/\/[^\s]+)/',
+                '/(?<!href=")(?<!src=")(?<!\()(https?:\/\/[^\s<]+)/',
                 function ($matches) {
                     $url = $matches[1];
                     $parsedUrl = parse_url($url);


AW0005 (approved):

ScuffedNewt (commented):
should this instead stop at the next closing anchor? i could see this causing issues with nested styling

SpeedyD:
> should this instead stop at the next closing anchor? i could see this causing issues with nested styling

It starts at `https://` or `http://` and ends before the first `<`. If there's nested styling, it should not affect the nested styling, from what I can tell- it would just put it AROUND the link rather than inside.

Issue is that I actually found a link that was like `<a href="http://www.link.com"><strong>http://www.link.com</strong></a><br />` and the regex triggered _wrongly_ on `http://www.link.com</strong></a>`, making that the entire link. 

Since it literally stops at the first `<` it should only trigger on the url, which likely makes nested styling style _around_ the link, seeing how the replacement code work.

Thing is.. I've seen this firsthand with someone who just copy pasted from the address bar. Wysiwyg editor posted them as links and they broke. Had `</a><br` in the link right behind the actual intended url. This NEEDS to be fixed, and I'm surprised it hadn't come up before.

**EDIT:**
Here's with nested styling actually:
with nested

Here's the same, but WITHOUT my change:
with nested sans fix
Before my fix, it added the extra styling INTO THE REPLACEMENT URL. You can see how it actually finds the ENTIRE CHUNK BEHIND THE LINK. Because the regex takes everything until it hits a space.

(Screenshots courtesy of https://regex101.com/ - again I swap between it and https://regexr.com/ )

ScuffedNewt:
thanks for the explanation! missed / forgot that it started on https... even though I wrote it lol

ScuffedNewt (approved):

PR #1161, opened 2024-12-23
fix(pagination): A ductape solution to the missing input problem

SpeedyD:
Pagination works again! No idea why inputs don't work otherwise, though..
pagination

SOMETHING changed in between release and develop, breaking it..

### DIFF
--- a/resources/views/layouts/_pagination.blade.php
+++ b/resources/views/layouts/_pagination.blade.php
@@ -45,12 +45,13 @@
             @endif
         </ul>
     </nav>
-    
+
     @once
-                        <div class="pagination-popover-origin d-flex align-items-center hide" style="gap: 10px;">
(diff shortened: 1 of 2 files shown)


SpeedyD:
Turns out this is NOT the full solution I was hoping for, the 'press enter to instantly go' function is dead, and I don't know how to fix that.

Regardless, I think this is good to put in right now, and see if we can fix _that part_ later. Since now it's fully broken.

SpeedyD:
Never mind previous comment, I found why- it's a function I renamed (because I thought `onClick` was too generic) and forgot to rename it in the appropriate section as well.

SO YEAH: This somehow fixes the problem. 🤷

itinerare (approved):

ScuffedNewt (approved):

PR #1097, opened 2024-10-28
Extended Mentions: Re-Extended

SpeedyD:
Based on the conversation in issue #1083 and the original hashing PR #749, I bring you:
Extended Mentions: Re-Extended

..in draft, because I feel like taking a break right now, but wanted to put it up for potential comments.

**PR currently adds:**
parseItems
parseTraitThumbs
parseItemThumbs
parsePrompts
parsePromptThumbs
(+ idUrl attribute to Prompts Model)

**Still on the planning:**
parseRarityThumbs
parseSpeciesThumbs
parseSubtypeThumbs
parseShopTumbs
parseCurrencyThumbs
parseCurrencyCategoryThumbs
parseCharacterCategoryThumbs
parsePromptCategoryThumbs
parseTraitCategoryThumbs
parseItemCategoryThumbs

PR #1167, opened 2024-12-27
fix(models): Adding IdUrl to Prompts

SpeedyD:
This is originally part of #1097, but I figure this may have a more general use.

Considering it should really be there, even if currently unused, I consider this more a bugfix than anything.. hence why it's on release rather than develop.

### DIFF
--- a/resources/views/prompts/_prompt_entry.blade.php
+++ b/resources/views/prompts/_prompt_entry.blade.php
@@ -6,9 +6,9 @@
         <x-admin-edit title="Prompt" :object="$prompt" />
         <div class="mb-3">
             @if (isset($isPage))
-                <h1 class="mb-0">{!! $prompt->name !!}</h1>
+                <h1 class="mb-0">{!! $prompt->name !!} <a href="{{ $idUrl }}" class="world-entry-search text-muted"><i class="fas fa-search"></i></a></h1>
             @else
-                <h2 class="mb-0"><a href="{{ url('prompts/' . $prompt->id) }}">{!! $prompt->name !!}</a></h2>
+                <h2 class="mb-0"><a href="{{ $idUrl }}">{!! $prompt->name !!}</a></h2>
             @endif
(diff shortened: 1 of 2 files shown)


SpeedyD:
Added a minor use by changing the link that's there, that's currently not pulling from the variable.

As well as a looking glass on the page version, cause some folk prefer to copy links rather than from the address bar.

resources/views/prompts/_prompt_entry.blade.php
preimpression, line 1: I suspect this will throw an "undefined variable: idUrl" error since it isn't `$prompt->idUrl`

SpeedyD: And.. you'd be right. I'm kind of not sure why I didn't, it's on my test case. 🤔

ScuffedNewt (approved):

resources/views/prompts/_prompt_entry.blade.php
```diff
@@ -6,9 +6,9 @@
         <x-admin-edit title="Prompt" :object="$prompt" />
         <div class="mb-3">
             @if (isset($isPage))
-                <h1 class="mb-0">{!! $prompt->name !!}</h1>
+                <h1 class="mb-0">{!! $prompt->name !!} <a href="{{ $prompt->$idUrl }}" class="world-entry-search text-muted"><i class="fas fa-search"></i></a></h1>
```
itinerare, line 9: What's the thought with adding the little search indicator here?

ScuffedNewt: pretty sure the same convention is used for items

itinerare: Alright, makes sense enough to me

SpeedyD: Yeah, based it on how items have it, Newt got it in one

itinerare (approved):

PR #1168, opened 2024-12-29
feat(Pages/News): Regenerate Function

SpeedyD:
Takes the text data and re-parses it! Useful for regenerating mentions without updating post time or user or whatnot. :)

### DIFF
--- a/app/Http/Controllers/Admin/PageController.php
+++ b/app/Http/Controllers/Admin/PageController.php
@@ -111,7 +111,7 @@ class PageController extends Controller {
 
         return redirect()->to('admin/pages');
     }
-    
+
     /**
      * Gets the page regeneration modal.
      *
--- a/app/Http/Controllers/Admin/PageController.php
+++ b/app/Http/Controllers/Admin/PageController.php
@@ -111,4 +111,39 @@ class PageController extends Controller {
 
         return redirect()->to('admin/pages');
     }
+    
+    /**
+     * Gets the page regeneration modal.
+     *
+     * @param int $id
+     *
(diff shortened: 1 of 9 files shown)


app/Http/Controllers/Admin/PageController.php
```diff
@@ -111,4 +111,39 @@ public function postDeletePage(Request $request, PageService $service, $id) {
...
+
+    /**
+     * Regenerates a page.
+     *
+     * @param App\Services\PageService $service
+     * @param int                      $id
+     *
+     * @return \Illuminate\Http\RedirectResponse
+     */
+    public function postRegenPage(Request $request, PageService $service, $id) {
```
ScuffedNewt, line 138: Nit-pick but i think this would be better to be more descriptive like `postRegeneratePageText` or something! otherwise looks good

SpeedyD: ..It's just a function name. In an attempt to be consistent with other functions, I use a simple get and post combination, to the point.. What it does is explained in the context box right above it..

I don't see a reason to rename the function?

ScuffedNewt: there's no particular reason aside from stylistic choice, I just think that the extra information in the function name makes sense and follows along some of the others that exist in LK! you don't have to agree or follow that though

itinerare: Seems clear enough to me as-is.

itinerare (approved):

AW0005 (approved):

PR #1177, opened 2025-01-04
feat(design): Remove Manual Thumbnail Image Upload (+minor text alterations)

SpeedyD:
Simple change, but requested apparently. Adds an additional setting to disallow users from uploading a manual thumbnail image, instead forcing them to use the cropper, if automation is disabled.

(Also slightly modifies wording for the automation, as to ensure there is clarity between options.)

### DIFF
--- a/config/lorekeeper/settings.php
+++ b/config/lorekeeper/settings.php
@@ -255,7 +255,7 @@ return [
     | This disables the option for users to manually upload their own
     | thumbnail images in design updates, including use of the cropper.
     | Note that this does not prevent permissioned staff from uploading
-    | custom thumbnail images. 
+    | custom thumbnail images.
     |
     | 0: Allows custom thumbnail uploads.
     | 1: Disallows custom thumbnail uploads.
@@ -274,7 +274,7 @@ return [
     | This disables the option for users to manually upload their own
     | thumbnail images in design updates, requiring use of the cropper.
     | Note that this does not prevent permissioned staff from uploading
-    | custom thumbnail images. 
+    | custom thumbnail images.
     |
     | 0: Allows custom thumbnail uploads.
     | 1: Disallows custom thumbnail uploads.
--- a/config/lorekeeper/settings.php
+++ b/config/lorekeeper/settings.php
@@ -246,39 +246,38 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Masterlist Image Automation Removing Manual Upload For Users
+    | Masterlist Image Automation Hide Manual Thumbnail
     |--------------------------------------------------------------------------
     |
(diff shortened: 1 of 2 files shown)


itinerare:
Yeah, I think this has come up a time or two...

config/lorekeeper/settings.php
```diff
@@ -263,6 +263,26 @@
     */
     'masterlist_image_automation_hide_manual_thumbnail' => 0,
 
+    /*
+    |--------------------------------------------------------------------------
+    | Remove Manual Thumbnail Image Upload Anyway
```
itinerare: Remove the "Anyway"; once again, this is functionally technical writing

config/lorekeeper/settings.php
```diff
@@ -263,6 +263,26 @@
...
+    |--------------------------------------------------------------------------
+    |
+    | NOTE: If the above feature, the Masterlist Image Automation Removing
+    | Manual Upload For Users is enabled, this option is irrelevant.
```
itinerare: See the above note;
`NOTE: If the "Masterlist Image Automation Hide Manual Thumbnail" setting above is enabled, this setting has no effect.`

config/lorekeeper/settings.php
```diff
@@ -263,6 +263,26 @@
...
+    |
+    | The following option is for if you want to disable the manual uploading
+    | of thumbnail images, to ensure users do not attempt to upload their
+    | own images, even despite the lack of automation.
+    | This essentially forces users to use the cropper, instead.
+    | This will remove it purely for users, not administration.
```
itinerare: `This disables the option for users to manually upload their own thumbnail images in design updates, requiring use of the cropper.`
`Note that this does not prevent permissioned staff from uploading custom thumbnail images.`

...or something like that. More concise (and dry). You know the drill.

config/lorekeeper/settings.php
```diff
@@ -263,6 +263,26 @@
...
+    |
+    | 0: Keeps the manual thumbnail image upload for users.
+    | 1: Hides the thumbnail upload for users.
```
itinerare: `0: Allows custom thumbnail uploads.`
`1: Disallows custom thumbnail uploads.`

itinerare:
The functional change is fine, so it's mostly just my nitpicks on the config text and then this should be ready to go 👍

SpeedyD:
> The functional change is fine, so it's mostly just my nitpicks on the config text and then this should be ready to go 👍

Thanks @itinerare! I even changed the text for the preceding setting, as it's functionally similar. :)

itinerare (approved):

ScuffedNewt (approved):
you absolutely beautiful human

PR #1186, opened 2025-01-10
feat(currencies): add currency categories

itinerare:
- add sort, category search to world currencies index
- group home/user bank pages by currency

Currencies grouped by category display a small header on the home/user bank pages; however, unlike items, uncategorized currencies do not display a header so as to match the existing look and feel. Currency display on the profile itself is unimpacted.


ScuffedNewt (approved):

SpeedyD (approved):
I like this, and I approve of this, but I had a minor request:
Could you also add 'is_visible' to the currency _itself_ while you're at it?

This is not a requirement, simply a request, and if you opt not to do so I'll likely work on a PR myself soon enough.

itinerare:
I'd rather not w/in the scope of this PR, but I'm fine with the notion in general.

PR #1188, opened 2025-01-10
feat(currency): Allow 'edit_data' staff to view user banks

SpeedyD:


SpeedyD:
As a note, I pushed the first commit, then realized that the category didn't have a marker. Edited that in, and then found out that, of course, it had been refactored. 🤦 

Anyway, that's why the commits are in such a weird order above.

SpeedyD:
Theeere we go. Thanks to @itinerare for giving me the final hint I needed to fix this up, haha. **Ready for review!**

app/Models/User/User.php
```diff
@@ -495,13 +495,18 @@ public function getCurrencies($showAll = false) {
 
         $owned = UserCurrency::where('user_id', $this->id)->pluck('quantity', 'currency_id')->toArray();
 
-        $currencies = Currency::visible()->where('is_user_owned', 1);
+        $currencies = Currency::where('is_user_owned', 1)
+            ->whereHas('category', function ($query) {
+                $query->visible(Auth::user() ?? null);
```
itinerare: iirc it's policy not to call `Auth` in models. Route the user object in when calling the function instead

SpeedyD: ...And here I just undid that. Fair enough, I'll put it back.

SpeedyD: I assume this is to your liking. :)

app/Models/User/User.php
```diff
@@ -484,24 +484,30 @@ public function canEditRank($rank) {
...
      * Get the user's held currencies.
      *
-     * @param bool $showAll
+     * @param bool       $showAll
+     * @param mixed|null $adminUser
      *
      * @return \Illuminate\Support\Collection
      */
-    public function getCurrencies($showAll = false) {
+    public function getCurrencies($showAll = false, $adminUser = null) {
```
itinerare: Should just be `$user` for clarity, since it's functionally *any* user object (if present).

SpeedyD: Can this not cause issues since this is, in fact, the User model?

itinerare: No; `User::` and `$user` are syntactically different.

SpeedyD: Fair enough. Editing.

itinerare (approved):

AW0005 (approved):

PR #1190, opened 2025-01-12
fix: flat array by default & extension compatibility

ScuffedNewt:
described in dms:

> theres an error on shops due to changes on getCurrencies on User.php in the _sidebar
presumably because its wrapped in the categories now instead of just being the currency
there should probably be a bool option to return it as a flat collection

the reason why im making useCategory a default false is because a lot of extensions use the getCurrencies()


SpeedyD (approved):
Good point, we missed this.

Also, the shop sidebar by default `/resources/views/shops/_sidebar.blade.php` has no categories, either, so I think this is a good call.

itinerare:
Oh, good call.

itinerare (approved):

PR #1191, opened 2025-01-12
fix: criteria loot

ScuffedNewt:
follow up to https://github.com/corowne/lorekeeper/pull/1124 since it has been left open without much activity

### DIFF
--- a/app/Console/Commands/FixEqualsCriteriaLoots.php
+++ b/app/Console/Commands/FixEqualsCriteriaLoots.php
@@ -2,11 +2,10 @@
 
 namespace App\Console\Commands;
 
-use Illuminate\Console\Command;
 use App\Models\Loot\Loot;
+use Illuminate\Console\Command;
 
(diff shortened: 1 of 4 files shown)


app/Console/Commands/FixEqualsCriteriaLoots.php
```diff
@@ -0,0 +1,44 @@
...
+     * @var string
+     */
+    protected $signature = 'fix-equals-criteria-loots';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Command description';
```
itinerare: Minor nit... would like a description for this. Otherwise, looks good, thanks for taking the time to do this!

itinerare (approved):

SpeedyD (approved):
I'm assuming this means PR 1124 would be closed, as this is the same fix but.. uh.. better.

PR #1192, opened 2025-01-12
feat: convert decodes / encodes to casts

ScuffedNewt:
to align existing models with more dev standards


itinerare:
Love this, will give it a thorough look later!

SpeedyD (approved):

app/Models/Character/CharacterDesignUpdate.php
```diff
@@ -332,7 +324,7 @@ public function getUrlAttribute() {
      * @return string
      */
     public function getVoteDataAttribute() {
-        return collect(json_decode($this->attributes['vote_data'], true));
```
itinerare, line 341: Oh, this needs latest develop to be merged in

app/Models/Gallery/GallerySubmission.php
```diff
@@ -35,6 +35,16 @@ class GallerySubmission extends Model {
...
     protected $table = 'gallery_submissions';
 
+    /**
+     * The attributes that should be cast to native types.
+     *
+     * @var array
+     */
+    protected $casts = [
+        'data'      => 'array',
+        'vote_data' => 'array',
```
itinerare, line 45: The decode in `getVoteData()` below is probably safe to remove given this... Guessing also since I changed it from attribute to function it doesn't need to use `->attributes[]` either

itinerare (approved):

PR #1194, opened 2025-01-13
add "reward maker" ext

CH3RVB:
it's basically an ext that allows for easier rewarding for loot like through prompts, also replaces the existing prompt reward system to streamline it

**worth noting that this is a very VERY rough draft** because i wanted to get feedback and see what the general opinion was-- it's a bit of mix of the old [default char rewards PR](https://github.com/corowne/lorekeeper/compare/develop...CH3RVB:lorekeeper:pr/default-character-prompt-rewards) + my [reward maker ext](https://github.com/CH3RVB/lorekeeper/tree/extension/reward-maker) (edited + added reward_key so a model can theoretically have multiple different "reward sets" 

things i believe i will have to edit/add:
- notif for when a character not owned by the prompt's submitter is set as focus and receiving rewards 
- remove migrated prompt rewards from the DB

this was a suggestion i got so i'm definitely feeling like i may have fumbled along the way somewhere, but like i said-- just looking for proper feedback so i can see where to take all of this (do we even want to replace prompt rewards with this or think it's a good idea???????)


preimpression:
Could the tables included in this to be swapped over to a responsive div format?

CH3RVB:
> Could the tables included in this to be swapped over to a responsive div format?

i just woke up so i might be short a few braincells ahahaha but i'm not getting what you're meaning by this, could you explain?

SpeedyD:
> > Could the tables included in this to be swapped over to a responsive div format?
> 
> i just woke up so i might be short a few braincells ahahaha but i'm not getting what you're meaning by this, could you explain?

In your code I saw a block looking like this, in `resources/views/widgets/_reward_display.blade.php`:
```
      <table class="table table-sm">
        <thead>
            <tr>
                <th width="70%">Reward</th>
                <th width="30%">Amount</th>
            </tr>
        </thead>
```
A sort-of equivalent would be this, for example, from `/resources/views/admin/items/items.blade.php`:
```
        <div class="mb-4 logs-table">
            <div class="logs-table-header">
                <div class="row">
                    <div class="col-5 col-md-6">
                        <div class="logs-table-cell">Name</div>
                    </div>
                    <div class="col-5 col-md-5">
                        <div class="logs-table-cell">Category</div>
                    </div>
                </div>
            </div>
```
Basically: Use bootstrap for the rows and columns.

CH3RVB:
> In your code I saw a block looking like this, in `resources/views/widgets/_reward_display.blade.php`:
> 
> ```
>       <table class="table table-sm">
>         <thead>
>             <tr>
>                 <th width="70%">Reward</th>
>                 <th width="30%">Amount</th>
>             </tr>
>         </thead>
> ```
> 
> A sort-of equivalent would be this, for example, from `/resources/views/admin/items/items.blade.php`:
> 
> ```
>         <div class="mb-4 logs-table">
>             <div class="logs-table-header">
>                 <div class="row">
>                     <div class="col-5 col-md-6">
>                         <div class="logs-table-cell">Name</div>
>                     </div>
>                     <div class="col-5 col-md-5">
>                         <div class="logs-table-cell">Category</div>
>                     </div>
>                 </div>
>             </div>
> ```
> 
> Basically: Use bootstrap for the rows and columns.

OH RIGHT yeah i can do that for sure! 🫡

app/Console/Commands/UpdateRewardMaker.php
```diff
@@ -0,0 +1,79 @@
...
+    public function handle() {
+        dd(DB::table('prompt_rewards'));
+        // while this will convert prompt rewards to the new system
+        // the first half if for those who used the existing ext and have existing data
+
+        $rewards = ObjectReward::all();
+
+        foreach ($rewards as $reward) {
+            switch ($reward->object_type) {
+                case 'Questline':
```
ScuffedNewt: definitely don't include questlines in this in the final PR

app/Console/Commands/UpdateRewardMaker.php
```diff
@@ -0,0 +1,79 @@
...
+                    $objmodel = 'App\Models\Questline\Questline';
+                    break;
+                case 'Prompt':
```
ScuffedNewt: You also dont need to do this at all and can instead use an object relation on the ObjectReward model

app/Models/ObjectReward.php
```diff
@@ -0,0 +1,46 @@
...
+
+    /**********************************************************************************************
+    RELATIONS
+     **********************************************************************************************/
+
+    /**
+     * Get the object.
+     */
+    public function object() {
+        return $this->morphTo(__FUNCTION__, 'object_type', 'object_id');
```
ScuffedNewt, line 30: whats the purpose of the `__FUNCTION__` ?

app/Models/ObjectReward.php
ScuffedNewt, line 1: instead of having an object reward model why not just have the stored assets array?

app/Models/Prompt/Prompt.php
```diff
@@ -78,8 +78,23 @@ public function category() {
     /**
      * Get the rewards attached to this prompt.
      */
-    public function rewards() {
-        return $this->hasMany(PromptReward::class, 'prompt_id');
+    public function objectRewards() {
+        return $this->hasMany('App\Models\ObjectReward', 'object_id')->where([
```
ScuffedNewt: ideally this would use the ::class

database/migrations/2024_10_14_165017_add_character_prompt_rewards.php
ScuffedNewt, line 1: since this is pulled from claymores, why not just pull the migration filename so that the migration errors _dont_ happen?

ScuffedNewt (commented):
small review for now will look in more detail later

perappu:
Since I'm putting my eyes on every PR today, just as a clean-up thing, this one should probably be closed due to https://github.com/lk-arpg/lorekeeper/pull/1338 getting merged.

PR #1199, opened 2025-01-20
Feat: fix minor bugs with trades & add tradelistings, trade proposals

### LINKED ISSUE #1150: item_filter in trades does not account for HTML in description or otherwise causing filter to fail
item descriptions that contain `""` will fail the json Parse making it impossible to filter the item inventory select
will fix as part of https://github.com/corowne/lorekeeper/pull/1141

ScuffedNewt:
Adds trade listings to dev, also adds the ability to propose trades to users
proposals use the existing trade system, the only difference being both users can edit both sides of the trades (however it must be accepted by both users to progress like normal trades)

Please pull this locally and test


app/Console/Commands/AddSiteSettings.php
```diff
@@ -75,6 +75,12 @@ public function handle() {
 
         $this->addSiteSetting('is_maintenance_mode', 0, '0: Site is normal, 1: Users without the Has Maintenance Access power will be redirected to the home page.');
 
+        $this->addSiteSetting('trade_listing_duration', 14, 'Number of days a trade listing is displayed for.');
+
+        $this->addSiteSetting('trade_listing_limit', 3, 'Number of trade listings a user can have at once.');
+
+        $this->addSiteSetting('open_trades_queue', 0, '0: Trades do not need mod approval, 1: Trades must be approved by a mod.');
```
ScuffedNewt: I forgot to complete this feature at time of PR, I'll add this in the next couple days

itinerare: Sounds good! I'll just mark this as a draft for now, then.

config/lorekeeper/extension_tracker.php
```diff
@@ -0,0 +1,25 @@
...
+    // 'extension_tracker' => [
+    //     'key' => 'extension_tracker',
+    //     'wiki_key' => 'Extension_Tracker',
+    //     'creators' => json_encode([
+    //         'Uri' => 'https://github.com/preimpression/',
+    //     ]),
+    //     'version' => '1.0.0',
+    // ],
+
+    'trade_listings' => [
```
itinerare: Needs to be removed...

resources/views/widgets/_bank_select_js.blade.php
```diff
@@ -7,6 +7,7 @@
             e.preventDefault();
             var $this = $(this);
             var $clone = $('#' + $this.data('type') + 'Row-' + $this.data('id')).find('.bank-row').clone();
+            console.log($clone);
```
itinerare: Also should be removed...

app/Services/TradeManager.php
```diff
@@ -265,7 +386,7 @@ public function confirmTrade($data, $user) {
             }
 
             if ($trade->is_sender_trade_confirmed && $trade->is_recipient_trade_confirmed) {
-                if (!(Settings::get('open_transfers_queue') && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
+                if (!Settings::get('open_trades_queue') && !(Settings::get('open_transfers_queue') && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
```
ScuffedNewt: needs to be changed to an or

resources/views/pages/credits.blade.php
```diff
@@ -88,6 +88,9 @@
         <p class="mb-0 col-md-4">
             <a href="http://wiki.lorekeeper.me/index.php?title=Extensions:Dynamic_Limits"><strong>Dynamic Limits</strong></a> by <a href="https://github.com/ScuffedNewt">ScuffedNewt</a>
         </p>
+        <p class="mb-0 col-md-4">
+            <a href="http://wiki.lorekeeper.me/index.php?title=Extensions:Trade_Listings"><strong>Trade Listings</strong></a> by <a href="https://github.com/itinerare">itinerare</a> & <a href="https://github.com/ScuffedNewt">ScuffedNewt</a>
+        </p>
         <p class="mb-0 col-md-4">
```
SpeedyD, line 103: This is admittedly a nitpick, but.. I beg you: Please put this in alphabetical like every other extension . . . 😅

config/lorekeeper/extension_tracker.php
```diff
@@ -0,0 +1,15 @@
...
+    //
+    // 'extension_tracker' => [
+    //     'key' => 'extension_tracker',
+    //     'wiki_key' => 'Extension_Tracker',
+    //     'creators' => json_encode([
+    //         'Uri' => 'https://github.com/preimpression/',
+    //     ]),
+    //     'version' => '1.0.0',
+
+];
```
SpeedyD: This file should not exist on v3!!

app/Models/User/User.php
```diff
@@ -385,7 +385,7 @@ public function getDisplayAliasAttribute() {
             return '(Unverified)';
         }
 
-        return $this->primaryAlias->displayAlias;
+        return $this->primaryAlias?->displayAlias ?? '(Unverified)';
```
SpeedyD: What is the initial ? for, after primaryAlias??

SpeedyD (changes requested):
Sorry Newt, got a few Qs and nits for ya.

Also.. is there a way to turn _off_ trade listings and trade proposals?? I didn't see anything regarding that..

We probably also should look into allowing a user to decline trade proposals.. 🤔

ScuffedNewt:
> We probably also should look into allowing a user to decline trade proposals.. 🤔

you can already reject proposals!

SpeedyD:
> > We probably also should look into allowing a user to decline trade proposals.. 🤔
> 
> you can already reject proposals!

Nah, I mean.. decline ahead of time, like a "no thanks" before getting any.

app/Models/User/User.php
ScuffedNewt: there is an edge-case where it can 500

SpeedyD: Shouldn't the double ?? at the end catch those?

ScuffedNewt: no because it's still trying to access an attribute on a null object

ScuffedNewt:
> > > We probably also should look into allowing a user to decline trade proposals.. 🤔
> > 
> > 
> > you can already reject proposals!
> 
> Nah, I mean.. decline ahead of time, like a "no thanks" before getting any.

sorry I don't get what you mean

SpeedyD:
> sorry I don't get what you mean

Basically, what I mean is like a user setting that can ahead of time be set that the user does not want to receive trade proposals.

But.. that may be out of scope for this particular PR, and may actually be a good idea for trades in general, not just proposals. (Which is why I ended it on that we should probably look into it, not actually as note for this PR.. Maybe I'll make it an issue instead.)

ScuffedNewt:
> Maybe I'll make it an issue instead.

good idea, definitely for me if the opt out of proposals they would be unable to create listings

SpeedyD (approved):
Right- with all that said, and having opened issue #1214... there's only one thing left to do.

And that was reviewing the code. I now understand that odd question mark is in fact a feature of PHP 8.. A Nullsafe Operator. Something I will look more into in the future. But for now.. You've addressed all my issues, and as such, I approve this PR.

itinerare:
Right, if that's settled I'll take a look at this when I next have the wherewithal.

SpeedyD (approved):
Merging in develop changes nothing, but re-approving nonetheless.

SpeedyD (approved):

resources/views/home/trades/listings/_listing.blade.php
```diff
@@ -0,0 +1,72 @@
...
+    <div class="card-header">
+        <h3 class="mb-0">
+            {!! $listing->displayName !!}: Posted by {!! $listing->user->displayName !!}
+            <a class="float-right" href="{{ url('reports/new?url=') . $listing->url }}">
+                <i class="fas fa-exclamation-triangle" data-toggle="tooltip" title="Click here to report this trade listing." style="font-size: 75%; opacity: 50%;"></i>
+            </a>
+            @if (isset($showStatus) && $showStatus)
+                <div class="mr-2 badge badge-{{ $listing->expired ? 'danger' : 'success' }} float-right">
+                    {{ $listing->expired ? 'Expired' : 'Active' }}
+                </div>
```
Draconizations, line 11: Needs to use `!$listing->isActive` instead. Expired listings show as active right now.

Draconizations (changes requested):
Just tested locally. Works almost flawlessly, though there is a minor visual bug I could spot.

app/Http/Controllers/Admin/TradeController.php
```diff
@@ -0,0 +1,103 @@
...
+
+        if ($type == 'completed') {
+            $trades->completed();
+        } elseif ($type == 'incoming') {
+            $trades->where('status', 'Pending');
+        } else {
+            abort(404);
+        }
+
+        $openTransfersQueue = Settings::get('open_trades_queue');
```
itinerare: It's probably better to have this fall back to the previous behavior if `open_trades_queue` is 0 (i.e. check `open_transfers_queue`), especially since the former defaults to 0. That way the existing behavior remains an option *and* no one gets caught unawares.

ScuffedNewt: you mean to say have it as,
`$openTransfersQueue = Settings::get('open_trades_queue') ?? Settings::get('open_transfers_queue')` ?

itinerare: I think it'd have to be something like `$openTransfersQueue = Settings::get('open_trades_queue') ? Settings::get('open_trades_queue') : Settings::get('open_transfers_queue')`, if I have my head wrapped around it right? Since it's not that it's necessary to check that the former is set...

app/Services/TradeManager.php
```diff
@@ -265,7 +395,7 @@ public function confirmTrade($data, $user) {
             }
 
             if ($trade->is_sender_trade_confirmed && $trade->is_recipient_trade_confirmed) {
-                if (!(Settings::get('open_transfers_queue') && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
+                if (!Settings::get('open_trades_queue') || !(Settings::get('open_transfers_queue') && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
```
itinerare: (just using the suggestion thing for ease of demonstration)
```suggestion
                if ((!Settings::get('open_trades_queue') || !(Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
```

itinerare: Thought to double-check the logic here, realized that "or" should probably be its own little sub-clause?

app/Http/Controllers/Admin/TradeController.php
ScuffedNewt: these are the same logically!

itinerare: Well, as long as it evaluates the same, that's fine by me.

SpeedyD (approved):
I approve of all changes made.

itinerare (approved):

PR #1203, opened 2025-01-21
feat: misc QoL

ScuffedNewt:
- Adds item search to rarities
- Adds ability for users to disable profile comments (& delete comments on their profile based on site setting)
- Adds require_email config setting

### DIFF
--- a/app/Console/Commands/AddSiteSettings.php
+++ b/app/Console/Commands/AddSiteSettings.php
@@ -93,7 +93,7 @@ class AddSiteSettings extends Command {
 
         $this->addSiteSetting('can_transfer_items_directly', 1, 'Whether or not users can directly transfer items to other users without trading. 0: Users cannot directly transfer items. 1: Direct item transfers are allowed.');
 
-        $this->addSiteSetting('allow_users_to_delete_profile_comments', 1, '0: Users cannot delete profile comments, 1: Users can delete profile comments.');
+        $this->addSiteSetting('allow_users_to_delete_profile_comments', 0, '0: Users cannot delete profile comments, 1: Users can delete profile comments.');
 
         $this->line("\nSite settings up to date!");
     }
--- a/app/Console/Commands/AddSiteSettings.php
+++ b/app/Console/Commands/AddSiteSettings.php
@@ -93,6 +93,8 @@ class AddSiteSettings extends Command {
 
         $this->addSiteSetting('can_transfer_items_directly', 1, 'Whether or not users can directly transfer items to other users without trading. 0: Users cannot directly transfer items. 1: Direct item transfers are allowed.');
 
+        $this->addSiteSetting('allow_users_to_delete_profile_comments', 1, '0: Users cannot delete profile comments, 1: Users can delete profile comments.');
+
         $this->line("\nSite settings up to date!");
     }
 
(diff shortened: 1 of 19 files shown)


app/Console/Commands/AddSiteSettings.php
```diff
@@ -93,6 +93,8 @@ public function handle() {
 
         $this->addSiteSetting('can_transfer_items_directly', 1, 'Whether or not users can directly transfer items to other users without trading. 0: Users cannot directly transfer items. 1: Direct item transfers are allowed.');
 
+        $this->addSiteSetting('allow_users_to_delete_profile_comments', 1, '0: Users cannot delete profile comments, 1: Users can delete profile comments.');
+
```
SpeedyD, line 97: I think this should be 0 by default, not 1, to maintain the fact that it is currently impossible.

resources/views/user/_profile_content.blade.php
```diff
@@ -156,17 +156,17 @@
...
+    @endif
+    <div class="col-md-{{ $user->settings->allow_profile_comments ? 4 : 12 }}">
         <div class="card mb-4">
-            <div class="card-header">
+            <div class="card-header" data-toggle="collapse" data-target="#mentionHelp" aria-expanded="{{ $user->settings->allow_profile_comments ? 'true' : 'false' }}">
                 <h5>Mention This User</h5>
             </div>
-            <div class="card-body">
+            <div class="card-body collapse {{ $user->settings->allow_profile_comments ? 'show' : '' }}" id="mentionHelp">
                 In the rich text editor:
```
SpeedyD, line 170: ...Why would the Mention This User section be collapsible? This is unrelated to comments?

SpeedyD (changes requested):
I left two points up, I think one should be changed, and the other.. well, I guess I'd like an explanation for it.

I have no idea why you seem to consider the 'Mention This User' section is part of comments? Why would it be collapsed? If anything, once I'm done with #1097 it'll be up to administration if users are allowed to see that anyway..

resources/views/user/_profile_content.blade.php
ScuffedNewt: because otherwise it looks ugly if it's col-12 and takes up the whole bottom half of the page (since comments are hidden if the user doesn't enable it)

app/Console/Commands/AddSiteSettings.php
ScuffedNewt: fair

resources/views/user/_profile_content.blade.php
ScuffedNewt: I definitely don't consider it part of comments, it's a visual thing

SpeedyD: Hrm.. I hadn't even considered that angle. Makes sense. Thank you for the explanation.
You can resolve this at your earliest convenience.

SpeedyD (approved):

LostInProgres (approved):

config/lorekeeper/settings.php
```diff
@@ -62,8 +62,13 @@
     | (e.g. ownership checking for characters only associated with an off-site account)
     | will still work provided users link the relevant alias(es).
     |
+    | The email option functiond as a fallback for users who register with an off-site provider.
```
itinerare: Typo here, `functiond`

itinerare:
Good additions on the whole! Just get that typo fixed and I'll be happy to merge.

itinerare (approved):

PR #1205, opened 2025-01-24
feat: mentions

ScuffedNewt:
combines emotes extension & @SpeedyD 's extended mentions into an ajax based system for ease of use & backwards compatibility with updates


SpeedyD (changes requested):
Looks like this is going to majorly clash with my #1097 PR.. Which, y'know, fair. This looks like a cleaner method, and allows for more.

...But I do not like the fact that it effectively removes a TON of backwards compatibility. Regenerating or simply editing older posts/pages/comments/etc. will break this way, as the parsed text is overwritten with the non-parsed text.. and with the code changed to what it is, it will not re-parse the existing code.

I cannot in good conscious approve this PR as it is.

And.. well, most minor note, I guess, is that the `config/lorekeeper/extension_tracker.php` file is re-added for some reason.. 🤔

app/Models/Gallery/GallerySubmission.php
```diff
@@ -376,6 +376,22 @@ public function getDisplayNameAttribute() {
         return '<a href="'.$this->url.'">'.$this->displayTitle.'</a>';
     }
 
+    /**
+     * Displays the submission's name, but with specific classes to identify mentions.
```
SpeedyD, line 370: Nitpick: says submission instead of prompt

SpeedyD: Nvm, this one is right, BELOW it says prompt instead of submission 🤦

SpeedyD (approved):
Having re-reviewed this, I'm not sure if I see any problems.

I'm already thinking of how to properly apply the rest of my plans, especially considering configuration.. Disallowing users to use a certain option, for one. But that'll be a later problem.

LostInProgres (approved):

resources/views/layouts/app.blade.php
```diff
@@ -155,8 +155,140 @@
                 $('[data-toggle="tooltip"]').tooltip({
                     html: true
                 });
-                $('.cp').colorpicker();
+                $('.cp').colorpicker(); <<
+                << << < HEAD
```
SpeedyD: This is a merge conflict left in.

SpeedyD (changes requested):
And this develop merging.. leaves me with no choice but to decline my previous feedback.

I know this is probably the big chunk of the changes that mentions originally was, but in a different PR, all the tinymce initializations were moved to `resources/views/js/_tinymce_wysiwyg.blade.php` - this is where I'd suggest you move this to. :)

SpeedyD (approved):
..and I realize now that I missed you already fixed it. My b, thanks for the conflict fixing commit though :)

resources/views/js/_tinymce_wysiwyg.blade.php
```diff
@@ -9,16 +9,127 @@
             convert_urls: false,
             plugins: [
                 'advlist autolink lists link image charmap print preview anchor',
-                'searchreplace visualblocks fullscreen spoiler',
-                'insertdatetime media table paste {{ config('lorekeeper.extensions.tinymce_code_editor') ? 'codeeditor' : 'code' }} help wordcount'
+                'searchreplace visualblocks code fullscreen spoiler',
+                'insertdatetime media table paste code help wordcount toc mention',
```
SpeedyD: You'll have to re-add `{{ config('lorekeeper.extensions.tinymce_code_editor') ? 'codeeditor' : 'code' }}` in front of `help wordcount`, or the `codeeditor` plugin will break..

ScuffedNewt: thanks for that, missed it

SpeedyD (approved):

SpeedyD (approved):

app/Console/Commands/ConvertExtendedMentions.php
```diff
@@ -0,0 +1,112 @@
...
+            '/\B@([A-Za-z0-9_-]+)/', // matches @mentions
+            '/\B%([A-Za-z0-9_-]+)/', // matches %mentions
+            '/\[user=([^\[\]&<>?"\']+)\]/', // matches [user=username]
+            '/\[userav=([^\[\]&<>?"\']+)\]/', // matches [userav=username]
+            '/\[character=([^\[\]&<>?"\']+)\]/', // matches [character=name]
+            '/\[charthumb=([^\[\]&<>?"\']+)\]/', // matches [charthumb=name]
+            '/\[thumb=([^\[\]&<>?"\']+)\]/', // matches [thumb=name]
+        ];
+
+        $replacements = [
```
SpeedyD, line 96: Oh my god I'm re-reviewing this code and the sudden realization hits me here: all the extensions for images are `.jpg`. I do not know how I missed this prior but that is definitely not going to fly for a _lot_ of images.

SpeedyD (changes requested):
Something needs to be done about the file extension. `.jpg` is not the way to go.

SpeedyD:
@ScuffedNewt: I would like to request you look over this PR once again. :)

PR #1206, opened 2025-01-24
chore(deps): update laravel/fortify

itinerare:
This is essentially a response to some strange (probably edge case) authentication behavior that isn't seen on develop, i.e. a sideways bugfix. Either way, it's probably not a bad idea to update this particular dependency, especially since the more recent version(s) are working fine enough for sites already on develop...


ScuffedNewt (approved):

SpeedyD (approved):
GAH, I thought I approved this! Sorry.

PR #1208, opened 2025-01-25
fix(characters): do JSON encode during character log creation

Draconizations:
Baby's first LK PR (it's a two line change).

Fixes a cast error while trying to update image features (amongst potentially some other edits? This is just the one I ran into). I might look into if using the model instead of directly calling `DB::table` here would work, but that's a task for another day.

### DIFF
--- a/app/Services/CharacterManager.php
+++ b/app/Services/CharacterManager.php
@@ -12,10 +12,12 @@ use App\Models\Character\CharacterDesignUpdate;
 use App\Models\Character\CharacterFeature;
 use App\Models\Character\CharacterImage;
 use App\Models\Character\CharacterImageSubtype;
+use App\Models\Character\CharacterLog;
 use App\Models\Character\CharacterTransfer;
 use App\Models\Sales\SalesCharacter;
 use App\Models\Species\Subtype;
 use App\Models\User\User;
+use App\Models\User\UserCharacterLog;
 use Carbon\Carbon;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Config;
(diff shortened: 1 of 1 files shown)


itinerare (approved):

ScuffedNewt (commented):
I would prefer if the model route was attempted first before merging this as the alternative, since otherwise this is a regression

Draconizations:
> I would prefer if the model route was attempted first before merging this as the alternative, since otherwise this is a regression

I didn't do this initially because it was very late at night for me. Trying it out now, it seems to work fine.

app/Services/CharacterManager.php
```diff
@@ -512,26 +514,37 @@ public function cropThumbnail($points, $characterImage, $isMyo = false) {
...
-                ] : [])
-        );
+                ]
+            );
+        } else {
+            $log = UserCharacterLog::create($shared);
+        }
+
+        if (isset($log)) {
+            return true;
```
ScuffedNewt: nit but you don't need to return true or false, just log, otherwise looks good!

ScuffedNewt (approved):

app/Services/CharacterManager.php
ScuffedNewt: the use of $shared was smart

Draconizations: > nit but you don't need to return true or false, just log, otherwise looks good!

haha I returned true/false because the method returned a boolean already and I wasn't sure if changing it would break anything. I'm fine with changing it if that's preferred

EDIT: went ahead and changed it because `DB::table->insert` doesn't return a boolean either

SpeedyD (approved):
~~Approved provided Newt's nitpick is changed; cause I think it's important to wipe that out.~~

And I make this comment seconds after the changed commit, hah!

app/Services/CharacterManager.php
ScuffedNewt: only if you want! just a nit after all, generally would leave that to authors choice

SpeedyD:
Do want to note.. This PR started as a matter of adding a json_encode into the mix.. And then a commit later, it's removed. That's kinda funny, isn't it.. Sometimes the better solution comes to mind later. :)

app/Services/CharacterManager.php
```diff
@@ -512,26 +514,37 @@ public function cropThumbnail($points, $characterImage, $isMyo = false) {
...
-                'character_id'  => $characterId,
-                'log'           => $type.($data ? ' ('.$data.')' : ''),
-                'log_type'      => $type,
-                'data'          => $data,
-                'created_at'    => Carbon::now(),
-                'updated_at'    => Carbon::now(),
-            ] + ($logType == 'character' ?
-                [
-                    'change_log' => $isUpdate ? json_encode([
+        $log = null;
```
itinerare, line 517: This can probably be removed, since you're already checking if it's set at the end

PR #1209, opened 2025-01-26
feat: allow head section

ScuffedNewt:
Why?
- Allows extensions to add javascript links etc in JUST their location, this is useful since it will prevent cross-site slowdown if multiple JS files are added to the primary view
- (ex for example, dailies has a JS script that is in the app.blade instead of on the dailies page)

### DIFF
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -80,6 +80,8 @@
     @endif
 
     @include('feed::links')
+
+    @yield('head')
 </head>
 
 <body>


SpeedyD (approved):
Ah.. some JavaScript needs to run from the head html.. But running it on every single page when it's only needed on one subsection...

Good point. Easy PR, Easy approval.

itinerare (approved):
Ooh, I actually implemented something like this elsewhere for similar reasons... it works well, in my experience

PR #1198, opened 2025-01-19
Add Tinymce code editor ext

MarskyMessier:
Related to https://github.com/corowne/lorekeeper/issues/1189, merged it into develop, entirely replaces the older code editor.
Edit: forgot to add I didn't add an extension credit given I wasn't sure if I was the one supposed to do so!

PR #1210, opened 2025-01-26
TinyMCE consolidation

SpeedyD:
After noticing #1198 requiring to change SEVERAL versions, I've opted to finally consolidate all instances of TinyMCE initialization, keeping distinct differences as variables.

I've also added an option to use the old code editor instead, but did set the new one as the default.

Finally, I've re-organized the credits; these are meant to be alphabetical, folk- don't tack 'em on to the end!
(And note it's _two_ lists, those without options and then those with.. both are alphabetical, though.)


SpeedyD:
As a sidenote, I've noticed that the old `js/_modal_wysiwyg.blade` was a perfect place to start, and used that as base. I did remove the old file, as it's only in use in two distinct locations..

Which do not seem to be in use. I will be looking through this and will likely be removing it, since it's not used- but that will be in a seperate PR.

SpeedyD:
...Wait a tick.. oh wow, the blade formatter broke the code.

itinerare:
Add it to .bladeignore at the base of the repo if you gotta

SpeedyD:
> Add it to .bladeignore at the base of the repo if you gotta

Thanks! Hopefully this'll fix it.. (**EDIT:** It did!)

LostInProgres (approved):

resources/views/character/_image_js.blade.php
```diff
@@ -7,24 +7,7 @@
...
-                    ],
-                    toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | codeeditor',
-                    content_css: [
-                        '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
-                    ],
-                    spoiler_caption: 'Toggle Spoiler',
-                    target_list: false
-                });
+                @include('js._tinymce_wysiwyg', ['tinymceSelector' => '.imagenoteseditingparse .wysiwyg', 'tinymceHeight' => '500'])
```
itinerare: Bit of a nit, but it feels to me like these variable names are longer than they need to be given that the view they're being handed off to is already fairly specific... But at the same time, that also doesn't really *matter* as long as it *works*,

itinerare:
Also, question: if using the editor in modals was fixed (I suspect so given the existence of the modal-specific setup), does this break it...? (The issue being using the sub-menus, e.g. link editing, while editing something in a modal. This came up again specifically re comments with WYSIWYG enabled for those, as replies and editing are done in a modal.)

resources/views/character/_image_js.blade.php
SpeedyD: Just wanted 'em to be specific. I can abbreviate them if needed..?

SpeedyD:
> Also, question: if using the editor in modals was fixed (I suspect so given the existence of the modal-specific setup), does this break it...? (The issue being using the sub-menus, e.g. link editing, while editing something in a modal. This came up again specifically re comments with WYSIWYG enabled for those, as replies and editing are done in a modal.)

...Good point. I never actually checked this, as.. that issue hasn't arisen in awhile for me, personally. I forgot about it.

First off: No. No, this does not fix it. This is merely reducing code altogether.
Second off: It has in fact not been broken until we got comment replies opened in a modal. The issue arises there again.

However, and this is very important.. This PR has absolutely nothing to do with that issue. This PR isn't even related to modal-specific setups. It merely reduces all the separate TinyMCE instances to a single one.

So.. reply comments out of a modal in a future PR? 🤷

itinerare:
Hmm. I know there was at least some kind of fix floating around, but it may not have made it into core as of yet.

SpeedyD:
> Hmm. I know there was at least some kind of fix floating around, but it may not have made it into core as of yet.

News to me. If anyone knows, I'd gladly see that in a different PR. :shrug:

itinerare:
I'll ask about it, but for the moment hold off on merging this until that's sorted out (in the event that having a separate set-up for modals is more sensible).

SpeedyD:
((Note to self, GOOD THING you _DID_ keep those two files in https://github.com/corowne/lorekeeper/pull/1210/commits/fb9b44edd5513f294f81bfa78bf46ffe2c3b16e4; despite the names containing the words 'modal', they're what replace the description and notes.. I should remember that. Maybe rename them in a future PR.))

SpeedyD:
Will be updating this soon as #1216 gets merged, need it as base. Do not merge in before that.

SpeedyD:
Merged in develop and release (as the latter includes #1216). Will make additional tiny edits, please hold.

SpeedyD:
And now the tinymce editor is no longer being initialized on every single page.

I'm done. Please comment and/or approve. :)

resources/views/account/settings.blade.php
```diff
@@ -223,6 +223,10 @@
 @endsection
 @section('scripts')
     <script>
+        $(function() {
+            @include('js._tinymce_wysiwyg', ['tinymceSelector' => '.wysiwyg', 'tinymceHeight' => '500'])
```
itinerare: I feel like it'd be simpler and more consistent to just have the contents of the include be self-contained (with its own script tag, etc., in line with other similar views used in the project) and included outside the script tag... (Likewise for the many other instances of the include)

SpeedyD: Agreed. Absolutely.. If it weren't for [/resources/views/character/_image_js.blade.php](https://github.com/corowne/lorekeeper/blob/88e8758b52bd5c805ccd2e43a1059aa1487c995d/resources/views/character/_image_js.blade.php).

It's kind of the odd one out that ruins it for everyone, and I am right now not even in the slightest sure how to alter that one.

Suggestions welcome.

itinerare: Oh, *that's* a pain. Fair enough, though.

resources/views/admin/characters/create_edit_character_category.blade.php
```diff
@@ -86,7 +86,9 @@
 @section('scripts')
     @parent
     <script>
-        $(document).ready(function() {
+        $(function() {
```
itinerare: That being said, is there a reason you're using `$(function() {...` in place of the existing `$(document).ready(function() {...`?

SpeedyD: Realized it's shorter, but does the exact same. That's it. No other reason.
(Half-considering another PR to replace other instances..)

itinerare: I'd argue against it for the same reason the [jQuery docs](https://learn.jquery.com/using-jquery-core/document-ready/) recommend using the long form; people with no or limited exposure to JS in general, let alone jQuery, are liable to be looking at it. Alternately put, it may be more concise, but it's less readable.

SpeedyD: Aight, fair enough- I'll do the opposite then.

I do want to point out that we _have_ been using the shorthand on app.blade.php, but I'll make that the long form as well.

itinerare: Sounds good to me.

itinerare (approved):

resources/views/account/settings.blade.php
SpeedyD: I am UNRESOLVING this because I realized something incredibly stupid and I might actually be able to fix this problem.

..YOU SEE- even though it's part of the image_js function?? It's ALSO loaded on the part that GETS loaded!! So I may still want to look into this again.

itinerare (approved):

resources/views/js/_tinymce_wysiwyg.blade.php
```diff
@@ -1,5 +1,7 @@
+@if (!isset($noscript))
```
itinerare: TBH, this was kinda what I was thinking might work as a solution to the earlier issue, haha

That said, I think it might be prudent to *also* check if it's true.

SpeedyD: Yeah, I kinda facepalmed when I realized this myself, but yesterday? Yesterday this idea was nowhere near my mind.

SpeedyD: I sincerely hope this is to your liking. Took your earlier comment in consideration too, and have edited the variable names. :)

resources/views/js/_tinymce_wysiwyg.blade.php
```diff
@@ -0,0 +1,26 @@
+@if (isset($MCEScr) && $MCEScr)
```
itinerare: The variable name *should* also be descriptive/readable and not condensed like this; I'd also recommend going back to the earlier method of having it default to including the script tags unless a given include indicates otherwise, i.e. have it default to the most common use case without needing to specify it in each include.

SpeedyD: Yeah, that makes sense. Fixed.

itinerare: By which I *also* mean *omitting* it from the majority of includes and only setting it when necessary (i.e. when *not* using the default behavior/including the script tag).

SpeedyD: How about I even take it one step further and make sure people don't need to set any values at all?

itinerare (approved):

resources/views/js/_tinymce_wysiwyg.blade.php
itinerare: Y'know, I was kinda thinking about that while puttering around this morning, but couldn't remember if there were sensible defaults for the other two values and didn't want to chase you down with demands anyhow |D so I'm glad you just went for it

SpeedyD: I basically just went with the most common values. Especially since I had to copy paste those values about 30ish times. :)

LostInProgres (approved):

PR #1215, opened 2025-01-28
fix: reply modal sub menus

ScuffedNewt:


itinerare:
This should probably go into the release branch first and foremost, if I'm not mistaken.

SpeedyD:
...Also these are absolutely the wrong files to edit.

Comments has the issue, not these two mislabeled files.

PR #1219, opened 2025-01-29
feat: item purchasing

ScuffedNewt:


config/lorekeeper/ext-tracker/shop_features.php
ScuffedNewt, line 1: this should be deleted after merge since this is the ext branch

SpeedyD (approved):

itinerare (approved):

PR #1222, opened 2025-02-02
feat: Laravel Sanctum API implementation

perappu:
- Add Laravel Sanctum package
- Add api_access staff power
- Add section in admin interface for API token generation
- Add API routes for generating a PAT and retrieving character data (as an example usage)
- Add section in Edit User admin interface for revocation of a user's API tokens

This current implementation only allows for one API token for user. This API token will have all the permissions that the user currently has. I considered adding granular permissions for API tokens, but it seemed like it would get too invasive to current features to add a second sort of power checking.


ScuffedNewt (changes requested):
small feature request, i think this would be most convenient if there was a way for the tokens to be generated on interaction with certain endpoints of the API, if that makes sense, rather than the user having to set their token (in the case of members)

I'm thinking about this from a minigame perspective, but considering (I believe) this feature is used by your arcade centre addon, maybe it already exists and I'm just not quite seeing it?

perappu:
> small feature request, i think this would be most convenient if there was a way for the tokens to be generated on interaction with certain endpoints of the API, if that makes sense, rather than the user having to set their token (in the case of members)
> 
> I'm thinking about this from a minigame perspective, but considering (I believe) this feature is used by your arcade centre addon, maybe it already exists and I'm just not quite seeing it?

Nah you're right, it doesn't exist yet. It makes a lot of sense, I'll see if that's doable -- and this actually helps with wrapping my mind around something I haven't solved yet.

Makes me realize that I don't think the api_access staff permission is necessary, because I can just use checking that we already have (such as, if someone makes an API request to pull data, it can just do a verification against the edit_data power). No reason to stop normal users from having PATs. I'll refactor the feature with this in mind!

perappu:
Just made some updates based on previous discussion:

- Removed api_access power and admin panel stuff (however admins can still revoke all of a user's tokens from edit user panel)
- Allow users to revoke/generate their own tokens from settings
- Non-CSRF route to generate a "hidden" token based on currently authenticated user
- Regenerating the token either via user settings or the non-CSRF route will cause the previous token to be deleted (so a user can theoretically only have two active tokens at a time)

Let me know if anyone thinks the route should live somewhere else ("account" made the most sense to me) or any other feedback!

SpeedyD:
This is a feature for users now? Then I'm going to have to ask if you can add the option to hide it from users. Some administration may not want their users to do this, at all.

perappu:
Due to popular demand, yes, it was added for users!

But that's easy enough. Just added a new site setting where API token generation for users is disabled by default. It does two checks -- on the settings page, and also within the postGenerateToken function.

It's worth noting that, despite the visual warning, the default implementation doesn't allow for the PAT to actually do anything. It was a conscious choice. There's one example route -- allowing people with edit_data to pull data for a specific character -- but that's it.

SpeedyD:
Thank you for adding that. Unfortunately, I do not have the knowledge rn to properly approve your PR, and some of this is going way over my head. I'll try to schedule some time to read up on this properly later on. :)

perappu:
Added another site setting that, in combination with the other site setting, allows for API access/token generation to be blocked entirely.

app/Console/Commands/AddSiteSettings.php
```diff
@@ -95,6 +95,10 @@ public function handle() {
 
         $this->addSiteSetting('allow_users_to_delete_profile_comments', 0, '0: Users cannot delete profile comments, 1: Users can delete profile comments.');
 
+        $this->addSiteSetting('allow_users_to_generate_tokens', 0, '0: Users cannot generate their own API tokens, 1: Users can generate their own API tokens.');
```
ScuffedNewt, line 98: This is better than initial, but I think it would be best if the generation of tokens is entirely outside of non-staff's hands.

When I mean generation of a token on an endpoint, I meant an automatic dispense of the token to either browser session or some other carrying method!

app/Services/UserService.php
```diff
@@ -696,4 +700,71 @@ public function reactivate($user, $staff = null) {
...
+    public function generateToken($user) {
+        try {
+            $user->tokens()->where('personal_access_tokens.name', 'token')->delete();
+
+            $token = $user->createToken('token')->plainTextToken;
+
+            UserUpdateLog::create(['staff_id' => $user->id, 'user_id' => $user->id, 'data' => json_encode([]), 'type' => 'Generated API Token']);
+
+            flash('Token created successfully:')->success();
+            flash($token)->success();
```
ScuffedNewt, line 720: yeah ideally we do not let the users do this at all. most members will have no idea what to do with this and ideally shouldn't

config/sanctum.php
ScuffedNewt, line 1: I think this entire endeavour can be done without Sanctum, is there any benefit to using Sanctum at this point in time?
Just concerned about introducing another dependency if we can do this vanilla

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
```diff
@@ -0,0 +1,30 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void {
+        Schema::create('personal_access_tokens', function (Blueprint $table) {
```
ScuffedNewt, line 12: I would prefer this to be named `user_api_tokens` or similar to follow a more consistent schema

ScuffedNewt (changes requested):

app/Services/UserService.php
perappu: I think I've received conflicting feedback regarding whether or not users should have the feature. I can see circumstances where we might want it for users -- such as giving a discord bot their API key, like I've seen with bots that connect to MMO APIs.

My happy medium would be a feature that can be disabled with a site setting, but if it's strongly desired, I can revert back to the api_access solution. 

Maybe I can have it be disabled in the config rather than as a site setting so it has to be enabled more "deliberately"?

config/sanctum.php
perappu: I chose Sanctum as it's widely supported and also the lightest API solution for Laravel. I understand the concern about adding an additional package, but think it's more secure to go with a frequently updated library rather than something bespoke, which may fall out disuse and have less resources for people building off of it.

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
perappu: Will update -- this is just the default Sanctum migration.

config/sanctum.php
ScuffedNewt: i get that! but what im saying is is there any specific thing that sanctum is providing here that we can't just get with regular endpoints?

routes/lorekeeper/members.php
```diff
@@ -56,6 +56,10 @@
     Route::post('bookmarks/edit/{id}', 'BookmarkController@postCreateEditBookmark');
     Route::get('bookmarks/delete/{id}', 'BookmarkController@getDeleteBookmark');
     Route::post('bookmarks/delete/{id}', 'BookmarkController@postDeleteBookmark');
+
+    Route::post('api/token', 'ApiController@postGenerateToken');
+    Route::get('api/token', 'ApiController@getGenerateToken')->withoutMiddleware([App\Http\Middleware\VerifyCsrfToken::class]);
```
ScuffedNewt: instead of doing without there's an `$except` array in the middleware

app/Console/Commands/AddSiteSettings.php
perappu: See my other comment -- I can definitely remove this code if that's the general consensus, but I'd like to try to standardize as many API use cases as I can (since that's the main purpose of this PR, standardizing an API implementation).

(and also I replied to these in backwards order but thank you for reviewing this!!)

config/sanctum.php
perappu: You'll have to forgive me if I'm getting something wrong but from what I was reading, Fortify isn't designed for API authentication and Sanctum/Passport is preferred. We could use the basic routes and code our own token generation but I'm hesitant to circumvent Fortify in such a way, plus it just feels like reinventing the wheel? Feels more secure to keep the API authentication and session-based authentication (mostly) separate

ScuffedNewt: ah yep i misunderstood! all good on that front

database/migrations/2019_12_14_000001_create_personal_access_tokens_table.php
perappu: On second thought this is probably more trouble than it's worth? I'd have to override the default model, and that could get messy: https://github.com/laravel/sanctum/issues/130

Looks like they never ended up putting it in the config file when I went digging lol

PR #1228, opened 2025-02-13
fix: hide the unlock button on prompts since it gets debited on the submit

ScuffedNewt:
this is something that was fixed in claymores and just needed to be cross ported
all it does is improve UX by hiding the unlock buttons on spots where the unlock should not be pressed


resources/views/widgets/_limits.blade.php
```diff
@@ -50,10 +56,10 @@
                 (Requires {!! implode(
                     ', ',
                     $limits->map(function ($limit) use ($limitTypes) {
-                            return ($limit->quantity ? $limit->quantity . ' ' : '') . $limit->limit->displayName;
+                            return ($limit->quantity ? $limit->quantity . ' ' : '') . $limit->limit->displayName . ($limit->limit_type == 'element' ? ' Element' : '');
```
ScuffedNewt: this line can be reverted

itinerare (approved):

SpeedyD (approved):

PR #1234, opened 2025-02-19
User font accessibility settings

LostInProgres:
Good day, besties of the Lorekeeper community-

This pretty much just adds user settings relating to text accessibility.
Currently included:
- Enable/Disable Site Fonts; forces the font used by core LK (Roboto Condensed, serif)
- Font size; changes font size
- Letter spacing; changes letter spacing, makes letter either further apart or closer together
- Word spacing; changes word spacing, makes letter either further apart or closer together
- Line height; changes how much room is between each line of text

These are all user settings, so these wouldn't affect anyone but the user that has adjusted them. 

Also worth noting that these changes aren't necessarily meant to be pretty; putting the settings on the higher end of the settings can definitely look a little bit messed up in some places, but IMO it's up to the user to decide whether the potential messed-up-ness is worth it.

---

Throwing this here mostly because I remember there being some discussion surrounding this, but honestly can't really... keep up with such discussion over discord.... So here we are :)

This is mostly done in my humble opinion- 
IIRC my To-Do for this was mostly
- turn the values into one of them fancy JSON thingamabobs so that it only needs one database column instead of like- five
- add (stricter) maximums to the values (y'know, so people can't set the font size to 9 and make the site unusable)
- Potentially add a couple more default font options


SpeedyD:
> turn the values into one of them fancy JSON thingamabobs so that it only needs one database column instead of like- five

Allow me to suggest you to _not_ do this. It's fine to keep it in separate columns.

MarskyMessier:
Maybe instead of data being cast into a single column as suggested, if we want to keep it neat backend wise we could have a new table like user_accessibility that's specifically for that purpose (could also be good if people want to add more settings) as not to clog the user_settings table.

SpeedyD:
Honestly, that would work just as well. Just do NOT JSON encode it into a single column.

Actually, I guess Marsky's suggestion would work quite a lot better- it would only add data for users who actually USE the accessibility functions..

itinerare:
I think either way would work, and each have pros and cons; JSON offers ease of adding or altering settings but is a little bit more of a pain to work with, whereas a separate table... adds a separate table, and requires migrations to change the variety of data stored in it... but doesn't need a row for *every* user, just those that engage with the settings.

PR #1246, opened 2025-03-16
fix(notifications): differentiate between MYO and character URLs in admin transfer notifs

itinerare:
This one's probably a fine vintage bug that went undetected for a long while, largely due to the fact that in many cases (when admin transferring full characters), it's been working as intended; however, when admin transferring *MYO slots*, the link given goes nowhere.

This... will technically break existing admin transfer notifications (in the sense that the link(s) given will be incorrect), but given that 1. it doesn't fully error/the notifications load without issue and 2. notifications are (at least intended to be) inherently ephemeral, this is, I think, probably fine.

### DIFF
--- a/app/Services/CharacterManager.php
+++ b/app/Services/CharacterManager.php
@@ -1523,7 +1523,7 @@ class CharacterManager extends Service {
             if ($sender) {
                 Notifications::create('CHARACTER_SENT', $sender, [
                     'character_name' => $character->slug,
-                    'character_slug' => $character->slug,
+                    'character_url'  => $character->is_myo_slot ? 'myo/'.$character->id : 'character/'.$character->slug,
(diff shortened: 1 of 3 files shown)


SpeedyD (approved):
Huh, we've ran into many versions of this particular issue already, surprised there was still another hiding.. Anyway, good catch, approved.

Just one question: Why is there a yarn.lock included in the PR?

itinerare:
Great question! The workflow keeps putting it there .

SpeedyD:
\*squint* Wait, it's the blade formatter?? What the heck?

itinerare:
Presumably the process of installing dependencies via NPM specifically, rather than the formatter itself, but more or less, yes.

ScuffedNewt (approved):

PR #1255, opened 2025-03-31
Fix(footer): Lorekeeper Github link

SpeedyD:
Should technically be done on main as well, but since the old url redirects just fine, I don't think it's that important to rush out a fix.


itinerare:
Yeah, I think it's fine to just handle it here and let that self-resolve when it hits stable.

itinerare (approved):

PR #1263, opened 2025-04-15
Global Component Naming for Debug Purposes

itsjop:
**Edit: This branch is getting retargeted to `development` soon** - leaving open for a moment as to not remove the conversation thus far.

Original Message: Added a simple string flag to the top of every relevant component that passes their component name up to the root `app.blade.php`, where it is appended as `data-` metadata on the `<#app>` div if debug mode is enabled.

This should assist in quickly finding the relevant component for what you're currently looking at, as, to the best of my knowledge, that's not currently immediately accessible through devtools or otherwise. This metadata approach also makes sure that the DOM is not modified through classnames or other methods of tracking information.

Currently this information is being mounted as metadata to the App root element, but if there's a more applicable place to send that information, I welcome feedback as this is my first LK source pull request.

Thanks!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-    "name": "lorekeeper",
+    "name": "lorekeeper_3",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
--- a/resources/views/account/aliases.blade.php
+++ b/resources/views/account/aliases.blade.php
@@ -1,4 +1,4 @@
-@extends('account.layout')
+@extends('account.layout', ['componentName' => 'account/aliases'])
 
 @section('account-title')
     Links
--- a/resources/views/account/banned.blade.php
+++ b/resources/views/account/banned.blade.php
@@ -1,6 +1,5 @@
 @extends('layouts.app', ['pageName' => '/banned'])
 
-
 @section('title')
     Banned
 @endsection
--- a/resources/views/account/banned.blade.php
+++ b/resources/views/account/banned.blade.php
@@ -1,4 +1,5 @@
-@extends('layouts.app')
+@extends('layouts.app', ['pageName' => '/banned'])
+
 
 @section('title')
     Banned
--- a/resources/views/account/bookmarks.blade.php
+++ b/resources/views/account/bookmarks.blade.php
@@ -1,4 +1,4 @@
-@extends('account.layout')
+@extends('account.layout', ['componentName' => 'account/bookmarks'])
 
 @section('account-title')
     Bookmarks
--- a/resources/views/account/deactivate.blade.php
+++ b/resources/views/account/deactivate.blade.php
@@ -1,4 +1,4 @@
-@extends('account.layout')
+@extends('account.layout', ['componentName' => 'account/deactivation'])
 
 @section('account-title')
     Settings
--- a/resources/views/account/deactivated.blade.php
+++ b/resources/views/account/deactivated.blade.php
@@ -1,4 +1,5 @@
-@extends('layouts.app')
+@extends('layouts.app', ['pageName' => '/account/deactivated'])
+
 
 @section('title')
     Deactivated
--- a/resources/views/account/layout.blade.php
+++ b/resources/views/account/layout.blade.php
@@ -1,6 +1,5 @@
 @extends('layouts.app', ['pageName' => '/account'])
 
-
 @section('title')
     Account ::
     @yield('account-title')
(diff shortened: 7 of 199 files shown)


SpeedyD:
I'll admit, this is an interesting one and since it's a relatively small change (despite 199 changed files) I don't really see an issue with accepting it..

..but I have to wonder if this is applicable for release. It is a new function, and as such should be relegated to the develop branch, where there are actually likely _more_ files that require changes.

itinerare:
In addition to the above-- this isn't really a *bugfix*, so it should rightfully go into develop if it's to go anywhere at all-- I'd also like to hear a practical reasoning and/or example use case for it; how does this tangibly improve the project and justify maintenance of this addition, enforcement of it as a standard for the project, etc. long-term? While a simple change in theory, it's very far-reaching, so I'm disinclined to merge it absent a fairly strong argument (especially given as it could exist regardless as an extension-like utility or resource, for instance, that could be available to others to pull as desired without directly impacting the core repo itself).

itsjop:
@itinerare @SpeedyD 
Can definitely retarget this to development upon request! I'm not deadset on merging this into v3 if it's not up to the spec of what's wanted, as I was under the impression the `development` branch was used for active development of current features and this wouldn't be apt for it. This is just an addition I made for my own personal version that was very simple to implement (aside from manually renaming 190 files), which has been helping me (as a novice on the project) jump directly to files that I'd normally have to search for a little bit. 
As far as long-term maintenance, it's a single _optional_ parameter added once on the first line of newly-created `.layout` files, which shouldn't be too much extra strain on future upkeep.  If this would be better as an extension or resource I'd love to move it there as well, pending any extra steps or information that I would need to do to make sure it's done properly. Thanks again!

edit: Might be able to add an enhancement to pull the appropriate code editor from prefs and make a app-link directly to the file 🤔

itinerare:
Currently release/v3.0.0 is only seeing bugfixes pending that becoming the latest stable release, so generally anything that *isn't* fixing a bug already present in that version should go into develop.

That being said, this *does* strike me as something which ought to exist and can potentially be a helpful resource, especially for folks in your position (some experience but not with this project etc.), but which isn't necessarily appropriate for inclusion in the project itself. Regarding maintenance, it looks to impact most every view (albeit minorly), and I'm disinclined to enforce that change for all future additions to the project/have that on the list of inevitable nitpicks for incoming PRs without a very good reason.
Of course, others are welcome to weigh in.

itsjop:
@itinerare Agreed that this shouldn't be mandatory for incoming PRs, as the debug parameter on <#app> won't render at all without the parameter passed, so worst-case it's simply blank there. Additionally, just like this PR has, it's something that can be mass-updated once development features are nearing completion. 
I'll leave this PR open for just a bit more for any further feedback and then re-target it to development for final tweaks/requests. Thanks!

AW0005:
Do you have a screenshot of what this shows you in practice when you're debugging?

I'd be tempted to poke around Laravels docs, there might be a more automated cleaner way to achieve the same result without needing to manually label everything.

itsjop:
@AW0005 Sure thing! Here it is in the firefox devtools:
image
The `data-component-path` is reporting the `user/profile` page, searching the key of which brings you right here:
image

These changes were initially brought about by a lack of devtools that I'd grown accustomed to over in JS land where you can just point at a component and immediately get taken to the exact file and line of the component you pointed at - but I was unable to find anything similar for PHP development. If there's something in the Laravel docs you know of that would be more elegant, I'd be all ears, haha.

itinerare:
I'd personally rather not backload work by mass-updating things later, and don't feel like it's appropriate to merge this without full intention of it being consistent across the board (it it's worth doing, it's worth doing well/thoroughly), which would mean it would be required for incoming PRs.

AW0005:
Thank you! When I said something in laravel I meant more there might be a way to pull the current view paths being used automatically from the framework to output to the data attribute rather than a completely different debug tool. 

That said I do know some people in the server have added this to their LK projects with some success for debugging and it does look like it includes view info:
https://github.com/barryvdh/laravel-debugbar

What you're describing from js isn't going to really exist here in the same form though because the template files are compiled server side, whereas js templates are compiled clientside and that's why the browser has insight on them in that way.

AW0005:
I did a little poking around and something like this:
```
View::composer('*', function ($view) {
            View::share('view_name', $view->getName());
        });
```
or this:
https://laracasts.com/discuss/channels/laravel/how-do-i-get-the-current-file-in-a-blade-template

Might give you a way to get the blade name automagically from laravel instead of having to hard-code it in a bunch of files

PR #1266, opened 2025-04-23
fix(world): Merged item_page into _item_entry and used an include instead

SpeedyD:
De-duplicating more code, yay


AW0005:
So personally I kind of prefer item_entry and item_page being separate because even though they're very similar in core, it can make a lot of sense to want to customize them to be different - primarily to have a more compact list page and a less compact detail page. 

And it's easier to do that if they're already separate. 

Also item_entry has an h3 while item_page had an h1 for the title that you should probably keep if we do still de-dupe them.

SpeedyD:
> So personally I kind of prefer item_entry and item_page being separate because even though they're very similar in core, it can make a lot of sense to want to customize them to be different - primarily to have a more compact list page and a less compact detail page.
> 
> And it's easier to do that if they're already separate.

I can make it an option instead, if that's preferred? It still requires updating regardless.

> Also item_entry has an h3 while item_page had an h1 for the title that you should probably keep if we do still de-dupe them.

I changed it to h3 for consistency, not sure why it was even made h1 in the first place.. 🤔

AW0005:
They should actually be different between the two - item_page has no other titles on the page so it's semantically correct for it to be an h1 whereas the list page has it's own h1 title. Semantically it should probably be updated to h2 for the _item_entry file tbh with an h3 class so it's not huge.

SpeedyD:
...If we're going to be arguing semantic use rather than consistency, I'm inclined to ask if you could make these semantic changes on a different PR.

AW0005:
I mean I'm mostly concerned with the removal of the h1 happening here then making sure it's 100% correct. From a UI perspective having the first title on the page be h1 for item_page is also being consistent with other LK pages - just in the UI of the page rather than striving for DRY code which I'd personally argue is more important. We even do this switch between h levels already on the single prompt pages which has a similar setup to what you're doing for here so it would be more consistent with what we've done elsewhere already as well to have it: https://github.com/lk-arpg/lorekeeper/blob/83d0dbafe51d56766e21f03ba176a1dd99d12fff/resources/views/prompts/_prompt_entry.blade.php#L8

SpeedyD:
> I mean I'm mostly concerned with the removal of the h1 happening here then making sure it's 100% correct. From a UI perspective having the first title on the page be h1 for item_page is also being consistent with other LK pages - just in the UI of the page rather than striving for DRY code which I'd personally argue is more important. We even do this switch between h levels already on the single prompt pages which has a similar setup to what you're doing for here so it would be more consistent with what we've done elsewhere already as well to have it:
> 
> https://github.com/lk-arpg/lorekeeper/blob/83d0dbafe51d56766e21f03ba176a1dd99d12fff/resources/views/prompts/_prompt_entry.blade.php#L8

Hrm.. That.. is actually a good point. Alright, you convinced me. I'll add that in- _tomorrow._

..also noticing that the currently used variable on that is isPage, whereas I have is_page.. _Whoops._ I'll set this PR to draft for now.

SpeedyD:
Got it done.. And with a cherry on top. @AW0005, satisfied?

LostInProgres (approved):

AW0005 (approved):
Thank you for the changes! Sorry it took me so long to re-review

resources/views/world/item_page.blade.php
```diff
@@ -31,122 +30,148 @@
...
-                                            <a href="{!! $item->category->url !!}">
-                                                {!! $item->category->name !!}
-                                            </a>
-                                        </p>
-                                    </div>
-                                @endif
-                                @if (config('lorekeeper.extensions.item_entry_expansion.extra_fields'))
-                                    @if (isset($item->rarity) && $item->rarity)
+                    <?php
+                    $shops = App\Models\Shop\Shop::where(function ($shops) {
```
itinerare: Wait, why is this still included? It should be (and looks like it *is*) using the method on the item model.

SpeedyD: You are absolutely right. I've fixed this now.

SpeedyD:
I've fixed the issues and.. simplified matters across the board, really. Not sure why they were linking separate parts of the item object. 🤔

LostInProgres (approved):

AW0005 (approved):

itinerare:
I'll give it another look over when less brainfogged :+1:

resources/views/world/_item_entry.blade.php
```diff
@@ -1,20 +1,31 @@
...
         <x-admin-edit title="Item" :object="$item" />
-        <h3>
-            @if (!$item->is_released)
-                <i class="fas fa-eye-slash mr-1"></i>
-            @endif
-            {!! $name !!}
-            @if (isset($idUrl) && $idUrl)
-                <a href="{{ $idUrl }}" class="world-entry-search text-muted">
-                    <i class="fas fa-search"></i>
+        @if (isset($isPage))
```
itinerare: Tiny nitpick, but probably safer to also check whether it's also true here.

resources/views/world/_item_entry.blade.php
```diff
@@ -67,14 +78,16 @@
...
-            {!! $description !!}
+            {!! $item->parsed_description !!}
             @if (((isset($item->uses) && $item->uses) || (isset($item->source) && $item->source) || $item->shop_stock_count || (isset($item->data['prompts']) && $item->data['prompts'])) && config('lorekeeper.extensions.item_entry_expansion.extra_fields'))
-                <div class="text-right">
-                    <a data-toggle="collapse" href="#item-{{ $item->id }}" class="text-primary">
-                        <strong>Show details...</strong>
-                    </a>
-                </div>
-                <div class="collapse" id="item-{{ $item->id }}">
+                @if (!isset($isPage))
```
itinerare: As above!

resources/views/world/_item_entry.blade.php
```diff
@@ -67,14 +78,16 @@
...
+                    <div class="text-right">
+                        <a data-toggle="collapse" href="#item-{{ $item->id }}" class="text-primary">
+                            <strong>Show details...</strong>
+                        </a>
+                    </div>
+                @endif
+                <div class="collapse {{ isset($isPage) && $isPage ? 'show' : '' }}" id="item-{{ $item->id }}">
```
itinerare, line 90: Likewise, as above!

resources/views/world/_item_entry.blade.php
SpeedyD: Working on it!

resources/views/world/_item_entry.blade.php
SpeedyD: This should do it!

resources/views/world/_item_entry.blade.php
SpeedyD: Actually, this one should be if FALSE. :)

resources/views/world/_item_entry.blade.php
```diff
@@ -1,20 +1,31 @@
...
         <x-admin-edit title="Item" :object="$item" />
-        <h3>
-            @if (!$item->is_released)
-                <i class="fas fa-eye-slash mr-1"></i>
-            @endif
-            {!! $name !!}
-            @if (isset($idUrl) && $idUrl)
-                <a href="{{ $idUrl }}" class="world-entry-search text-muted">
-                    <i class="fas fa-search"></i>
+        @if (isset($isPage) && $isPage == true)
```
itinerare: Most minor of minor nitpicks at this point, but this can just be `isset($isPage) && $isPage` (and likewise for if false, albeit `!$isPage`)... it's a little more concise and (more importantly) consistent with the rest of the project.

SpeedyD: ..wait, then why your complaint at line 90's previously?
afbeelding
It clearly HAD THAT.. I added '== true' because I thought you wanted it more specific XD

SpeedyD: Well, this should do it then.

itinerare: Ah, might've autopiloted through em then, my bad!

itinerare (approved):

PR #1275, opened 2025-05-16
feat: add description to item in inventory

MarskyMessier:
Drove me a bit insane that shop has the item description but not inventory (which means you can't quick check the item desc from your inventory) so I added it back o7

### DIFF
--- a/resources/views/home/_inventory_stack.blade.php
+++ b/resources/views/home/_inventory_stack.blade.php
@@ -18,6 +18,17 @@
         @endif
     </div>
 
+    @if ($item->parsed_description)
+        <div class="mb-2">
+            <a data-toggle="collapse" href="#itemDescription" class="h5">Description <i class="fas fa-caret-down"></i></a>
+            <div class="card collapse mt-1" id="itemDescription">
+                <div class="card-body">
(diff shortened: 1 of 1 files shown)


SpeedyD (changes requested):
I'll be requesting a minor change before approving this- can you remove the show class?

Yes, it will add a click to users wanting to read the description, but in all honesty, the item being accessed from inventory usually means someone wants to use it..
And if it's pushed down in the modal by potentially long descriptions from the get-go, that's more likely frustrating than the lack of descriptions.

Of course, it would also be a nice if this could have a setting, but I'm not entirely sure if THAT is necessary here.

MarskyMessier:
Right I honestly didn't think much about it as I copy-pasted the snippet from shop, thanks for pointing it out tho! Makes a lot of sense.

SpeedyD (approved):
Yeah, the difference I guess is.. in the shop you WANT to immediately know what the item is for, otherwise why purchase it, right? :P

LostInProgres (approved):

Draconizations (approved):

itinerare (approved):

PR #1279, opened 2025-05-24
fix(ranks & gallery submissions): parsed rank description on setup + gallery character notifications

liwoyadan:
Two tiny fixes...

1. Setting a parsed rank description when the Admin and Member roles are created upon initial setup and running the setup command for the admin user. I noticed that for sites setting up for the first time, the help icon on user profiles does not actually display a tooltip when hovered over; the tooltip uses the contents of `parsed_description` but when doing initial setup commands that value is null and only `description` is set...which now that I look at commit history on the main branch as I type this I think this might be the case for v2/main as well. (Main branch went from using `description` to `parsed_description` way back in April of 2019, so...) This just sets a value for it during setup so the tooltip works from the get-go. <sub>(The parse function the string is wrapped in can just be removed though...if that's more ideal I can edit that real quick.)</sub>
2. Was realizing that users & myself weren't reliably getting notifications for owned characters being included in other users' gallery submissions. Taking a peek at the code chunk I noticed it checks `$character->user` when the GalleryCharacter model does not have a user relation. Like other instances in that code chunk it should be going `$character->character->user` instead.

### DIFF
--- a/app/Console/Commands/SetupAdminUser.php
+++ b/app/Console/Commands/SetupAdminUser.php
@@ -46,16 +46,16 @@ class SetupAdminUser extends Command {
         if (!Rank::count()) {
             // These need to be created even if the seeder isn't run for the site to work correctly.
             $adminRank = Rank::create([
-                'name'        => 'Admin',
-                'description' => 'The site admin. Has the ability to view/edit any data on the site.',
+                'name'               => 'Admin',
+                'description'        => 'The site admin. Has the ability to view/edit any data on the site.',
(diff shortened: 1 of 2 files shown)


itinerare:
Could I trouble you to separate these fixes out into two separate PRs? Thank you in advance!

liwoyadan:
Yeah of course! I'll close this one and set up two separate PRs. :thumbsup:

itinerare:
Thanks for understanding!

PR #1282, opened 2025-05-26
fix: misc shop

ScuffedNewt:


itinerare (dismissed):

app/Http/Controllers/Admin/Data/ShopController.php
```diff
@@ -149,7 +149,11 @@ public function getEditShopStock($id) {
                 $type.'Category' => $categories->toArray(),
             ];
         } else {
-            $items = $model::orderBy('name')->pluck('name', 'id')->toArray();
+            if ($model == '\App\Models\Raffle\Raffle') {
```
itinerare, line 152: Keep the PR focused; move these changes to a separate PR.

itinerare (changes requested):

app/Http/Controllers/Admin/Data/ShopController.php
itinerare: (Either that or change the title etc, but in general I would rather PRs be as focused as is possible.)

SpeedyD (approved):

app/Http/Controllers/Admin/Data/ShopController.php
SpeedyD: Title's changed, seems good to me 🤷

itinerare (approved):

PR #1283, opened 2025-05-26
New Core Ext: All Traits Index

SpeedyD:
The All Traits Index, unofficially known as the Kitchen Sink Index (from the phrase "everything AND the kitchen sink").

It is an extension I've made a while ago, based off of the existing indexes. I very recently made a version for v3 and another for current dev for kicks- even adding proper settings to the latter cause why not.

..Which made me realize- why not add it to core? It's out of the way, very little code, and folk can just.. turn it off if they don't want it.


LostInProgres (approved):

Draconizations (approved):

itinerare:
I'd rather the route not *literally* be `kitchensink`, but otherwise, looks good.

SpeedyD:
> I'd rather the route not _literally_ be `kitchensink`, but otherwise, looks good.

The route is no longer literally `kitchensink`, but instead `all-traits-index`. All other references to Kitchensink remain, but are not publicly visible. :)

resources/views/world/_sidebar.blade.php
```diff
@@ -10,6 +10,9 @@
         @if (config('lorekeeper.extensions.visual_trait_index.enable_universal_index'))
             <div class="sidebar-item"><a href="{{ url('world/universaltraits') }}" class="{{ set_active('world/universaltraits*') }}">Universal Trait Index</a></div>
         @endif
+        @if (config('lorekeeper.extensions.visual_trait_index.enable_all_trait_index'))
+            <div class="sidebar-item"><a href="{{ url('world/all-traits-index') }}" class="{{ set_active('world/kitchensink*') }}">All Traits Index</a></div>
```
itinerare: The `set_active()` here needs to be updated!

SpeedyD: ...Dangit. Push in progress.

itinerare (approved):

PR #1284, opened 2025-05-27
feat(image): Adding header images to News/Sales/Pages

SpeedyD:
See title. :)

### DIFF
--- a/resources/views/admin/news/create_edit_news.blade.php
+++ b/resources/views/admin/news/create_edit_news.blade.php
@@ -32,6 +32,21 @@
     </div>
 
     <div class="form-group">
+        {!! Form::label('Header Image (Optional)') !!} {!! add_help('This image will show up above the news and on the meta-image.') !!}
+        <div class="custom-file">
+            {!! Form::label('image', 'Choose file...', ['class' => 'custom-file-label']) !!}
+            {!! Form::file('image', ['class' => 'custom-file-input']) !!}
+        </div>
+        <div class="text-muted">Recommended size: None (Choose a standard size for all news headers)</div>
+        @if ($news->has_image)
+            <div class="form-check">
+                {!! Form::checkbox('remove_image', 1, false, ['class' => 'form-check-input']) !!}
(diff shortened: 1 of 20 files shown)


SpeedyD:
..this was meant to be a draft. Why was it not a draft. 🤔

itinerare:
Prompt page meta tag images probably *should* be its own PR...

SpeedyD:
> Prompt page meta tag images probably _should_ be its own PR...

Yeah, figured. Done and done (#1285) and I believe this is NOW actually ready for review.

resources/views/admin/news/create_edit_news.blade.php
```diff
@@ -31,6 +31,21 @@
         </div>
     </div>
 
+    <div class="form-group">
+        {!! Form::label('Header Image (Optional)') !!} {!! add_help('This image will show up above the news content and on the meta-image.') !!}
+        <div class="custom-file">
+            {!! Form::label('image', 'Choose file...', ['class' => 'custom-file-label']) !!}
+            {!! Form::file('image', ['class' => 'custom-file-input']) !!}
+        </div>
+        <div class="text-muted">Recommended size: None (Choose a standard size for all news headers)</div>
```
itinerare: I suspect the "standard size" recommendation isn't fully necessary for these use cases...

SpeedyD: Based it off of the shop headers recommendation, to remain consistent...
afbeelding
(From: [`resources/views/admin/shops/create_edit_shop.blade.php`](https://github.com/lk-arpg/lorekeeper/blob/develop/resources/views/admin/shops/create_edit_shop.blade.php#L32))

I can remove it if you want to, though?

itinerare: Well, the thing about that is that those images are displayed in the shop index. So the recommendation is towards having a shop index that isn't kind of a mess... but no such concern exists here.

SpeedyD: Aight, fair enough 👍

LostInProgres (approved):

itinerare (approved):

PR #1292, opened 2025-06-14
Limits if check for manual unlocking

ScuffedNewt:
Would like someone to double check my logic here but:

line 53 in `LimitManager.php` was 
```php
if (!$is_unlock && $limits->first()->is_unlocked && !$limits->first()->is_auto_unlocked) {
```

this is wrong because `$limits->first()->is_unlocked` is already checked prior, & conflicts with the `is_auto_unlocked`

Updated to 
```php
if (!$is_unlock && !$limits->first()->is_auto_unlocked) {
```

### DIFF
--- a/app/Helpers/Helpers.php
+++ b/app/Helpers/Helpers.php
@@ -450,17 +450,6 @@ function prettyProfileName($url) {
 }
 
 /**
- * Returns the given objects limits, if any.
- *
- * @param mixed $object
- *
- * @return bool
- */
-function getLimits($object) {
-    return App\Models\Limit\Limit::where('object_model', get_class($object))->where('object_id', $object->id)->get();
-}
-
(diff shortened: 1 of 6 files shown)


SpeedyD:
> this is wrong because `$limits->first()->is_unlocked` is already checked prior, & conflicts with the `is_auto_unlocked`

Long story short, it IS checked prior AND if it conflicts, it needs to be fixed. I think your logic is solid. 👍

-----

..Anyway.. I would like to review this proper at some point, but I do not have the brain power right now =_=;..

What I can't figure out is where CheckLimits is called, I need to know what SETS `$is_unlock` to true, since it's false by default..

Draconizations (approved):

LostInProgres (approved):

PR #1301, opened 2025-07-01
add docker and docker-compose for easy setup

castaway2000:
as the title says i have added framework and implementation for an easy local deployment of lorekeeper entirely in docker-compose along with readme updates to describe how to make it work for anyone easily.

implementation was tested on my macbook air with the current commited dockerfile and compose yaml i would expect this to work for others on any OS with docker and compose.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -1,0 +1,251 @@
+# Running Lorekeeper with Docker
+
+This guide will help you set up and run Lorekeeper using Docker and Docker Compose, making it easy to get started without manual installation steps.
+
+## Prerequisites
+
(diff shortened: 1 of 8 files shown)


itinerare:
First and foremost, while it's in theory useful to be able to leverage docker like this, that will not be true for the majority of the project's userbase-- who are primarily non-technical in background, and for whom docker is liable to be an unnecessary additional layer of complexity. Consequently, this is a poor fit for inclusion in the project itself.

Second, it's recommended to create an issue or similar discussing your intent before opening a PR, especially without prior history with the project.

PR #1313, opened 2025-07-25
Fix: Lack of alias set as primary causing error 500

MarskyMessier:
I noticed this earlier trying to add social auth to another LK, if someone sets one alias and doesn't change it to their primary and the "require alias" setting isn't on it will error out on all pages involving the display alias. The fix should paliate that issue (and display "No alias" instead).

### DIFF
--- a/app/Models/User/User.php
+++ b/app/Models/User/User.php
@@ -398,7 +398,7 @@ class User extends Authenticatable implements MustVerifyEmail {
             return '(Unverified)';
         }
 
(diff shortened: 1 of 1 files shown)


itinerare:
If I'm not mistaken, this should go to the release branch.

MarskyMessier:
I'm good to close this pr and just reopen one on release!

MarskyMessier:
Closed pull request - ported to release.
See #1314

PR #1322, opened 2025-08-15
fix(user profiles): Rank info help does not appear if description empty

SpeedyD:
Noticed the little ? just.. does nothing if there's no text. Might as well remove it so that it doesn't confuse users. :)

### DIFF
--- a/resources/views/user/_profile_content.blade.php
+++ b/resources/views/user/_profile_content.blade.php
@@ -56,7 +56,7 @@
                 <div class="col-lg-3 col-md-3 col-4">
                     <h5>Rank</h5>
                 </div>
-                <div class="col-lg-9 col-md-9 col-8">{!! $user->rank->displayName !!} {!! !$user->rank->parsed_description ? '' : add_help($user->rank->parsed_description) !!}</div>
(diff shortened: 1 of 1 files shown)


liwoyadan (commented):
While this does nicely remove the help icon when the parsed description string is null, when there's content set it returns a 1 instead of the help icon:

SpeedyD:
> While this does nicely remove the help icon when the parsed description string is null, when there's content set it returns a 1 instead of the help icon

And this is why you shouldn't code five things at once. I did not think my use of the ternary operator through. 🤦

Fixed now, thank you for your comment @liwoyadan. :)

itinerare:
I'd wager it'd be simpler to just check if `parsed_description` is null/just have `!$user->rank->parsed_description ? ...`

SpeedyD:
> I'd wager it'd be simpler to just check if `parsed_description` is null/just have `!$user->rank->parsed_description ? ...`

Not sure if this is much of an improvement.. But I changed it :o

itinerare:
.... Oh yeah, checking if it's set (`$user->rank->parsed_description ? ...`) would be more sensible.

SpeedyD:
> .... Oh yeah, checking if it's set (`$user->rank->parsed_description ? ...`) would be more sensible.

...done and done. I'm going to bed now. XD

itinerare (approved):

liwoyadan (approved):

Draconizations (approved):

PR #1324, opened 2025-08-18
fix(meta tags): no more extra spaces

liwoyadan:
Since the new blade formatting there has been extra space rendered in meta tag titles when yielded. This trims off the extra space and also only renders the trailing `-` after the site's title (as well as the `::`) if there is an additional title yielded. Below are just preview screenshots displaying the before and after for Discord embeds when linking a user's characters page.
Previously generated Discord embed
Discord embed generation now

Meta descriptions have tags stripped because...I don't think those actually render? (thinking emoji) Previewing the opengraph/facebook/twitter/etc embeds the tags show up along with the text as `<p><strong>` so I figured it could just be taken out.

### DIFF
--- a/resources/views/account/layout.blade.php
+++ b/resources/views/account/layout.blade.php
@@ -1,8 +1,7 @@
 @extends('layouts.app')
 
 @section('title')
-    Account ::
-    @yield('account-title')
+    Account{!! View::hasSection('account-title') ? ' :: ' . trim(View::getSection('account-title')) : '' !!}
 @endsection
 
 @section('sidebar')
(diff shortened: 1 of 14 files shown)


SpeedyD (approved):
This PR let me to research 'hasSection' and 'getSection'.

...It seems I was unaware of an interesting feature, that potentially opens new opportunities for development.

I highly approve of this PR.

itinerare:
Yeah, there are a few such tools that Laravel provides and they're *great*.

Anyhow, thanks for taking the time to do this!

itinerare (approved):

PR #1328, opened 2025-08-31
feat: clarify and make easier for users to understand the effects of …

ScuffedNewt:
…manual unlocking and reducing user error in action workflows that expect certain unlock types

currently prompts expect `is_auto_unlocked` to be set to true so that the submission workflow can process correctly.
current workflow is:
user fills out submission -> acknowledges they cannot return to draft prior to staff cancelling -> limits are processed -> submitted

with `is_auto_unlocked` set to false, the limits could not be processed since the user has not "manually" unlocked. Manually unlocking is good for one time debits (ex unlocking a shop to enter it, so that a user accidentally clicking on it doesn't debit them) or debits that aren't implicitly part of a workflow

would love some comments on this in regards to making clear to users how these options interact or if it is already clear, what should default behaviours be, etc. considering it's part of dev now and should be documented according to the knowledge of the average user

### DIFF
--- a/resources/views/widgets/_limits.blade.php
+++ b/resources/views/widgets/_limits.blade.php
@@ -14,10 +14,14 @@
 @if ($limits)
     @if (!isset($compact) || !$compact)
         <h4 class="my-3">{!! $object->displayName !!}'s Requirements</h4>
-
         <p>
             You must obtain or complete all of the following in order to access this {{ $object->assetType ? (substr($object->assetType, -1) === 's' ? substr($object->assetType, 0, -1) : $object->assetType) : '' }}.
         </p>
+        @if ($limits->first()->is_unlocked && $limits->first()->isUnlocked(Auth::user() ?? null))
+            <div class="alert alert-success">
+                You previously have met these requirements and have unlocked access to this {{ $object->assetType ? (substr($object->assetType, -1) === 's' ? substr($object->assetType, 0, -1) : $object->assetType) : '' }}.
+            </div>
+        @endif
         <div class="mb-2 logs-table">
             <div class="logs-table-header">
                 <div class="row no-gutters">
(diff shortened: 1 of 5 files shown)


Draconizations:
I am not too familiar with limits, so I cannot make a judgement based on what I do know. But I _can_ say that these changes made things a lot clearer for me. To the point I don't really have questions anymore.

Take my stamp of approval with a grain of salt since again, I'm not knowledgeable here and have no good input on default behaviors. I just think that this is definitely better than before.

Draconizations (approved):

itinerare:
Yeah, seconding that this is all *great* clarification; on that front it's all good by me, but maybe like another set of eyes on it for the functional bits (if that makes sense). Good work on the whole, though!

LostInProgres (approved):

resources/views/widgets/_limits.blade.php
```diff
@@ -75,28 +84,29 @@
...
             @if (Auth::check() && !$limits->first()->isUnlocked(Auth::user() ?? null) && !$limits->first()->is_auto_unlocked)
-                <div class="alert alert-secondary p-0 mt-2 mb-0">
+                <div class="alert alert-secondary text-center px-0 py-1 mb-0">
                     {!! Form::open(['url' => 'limits/unlock/' . $limits->first()->id]) !!}
                     {!! Form::submit('Unlock', ['class' => 'btn btn-sm btn-secondary']) !!}
                     {!! Form::close() !!}
                 </div>
             @else
-                <div class="alert alert-secondary p-0 mt-2 mb-0">
+                <div class="alert alert-secondary text-center px-0 p-1 mb-0 font-weight-bold">
```
SpeedyD, line 93: Should this class not be p*y*-1? :)

I believe it just being p-1 negates the px-0 🤔

SpeedyD (approved):
Honestly, I think this is a great improvement, but I had ONE little thing I'm wondering if needs to be fixed...

resources/views/widgets/_limits.blade.php
ScuffedNewt: I don't think so, but feel free to correct me if I'm wrong

PR #1330, opened 2025-09-04
Add File Manager

Alliionart:
Integrates [alexusmai's](https://github.com/alexusmai/laravel-file-manager) file explorer into LK for staff to easily pick through the desired disks/folders. The file manager is accessible through:

- sitename.com/filemanager
- File manager on admin settings
- Tinymce Image Source Button

This updates Tinymce to v5.4.2 - so far I have not seen any conflicts with anything core in 3.0.0. But maybe need to watch out for any other extensions with ties to older versions. This should be compatible with future tinymce versions.

All methods require the user be staff AND have the "manage_files" power. The default accessible folder is the /files folder and nothing else. Though two public examples exist in the commented code in the config.

Security tested on all 3 with both staff without the power, staff with the power, normal members and guests. 
fm-gallery-view
file-manager-main


.gitignore
```diff
@@ -9,6 +9,5 @@ Homestead.json
 Homestead.yaml
 npm-debug.log
 yarn-error.log
-*.env
```
itinerare, line 12: This shouldn't be changed...

routes/web.php
```diff
@@ -11,6 +11,10 @@
 |
 */
 
+use App\Http\Controllers\FileManagerController;
+
+Route::get('filemanager', [FileManagerController::class, 'index']);
```
itinerare: Until/unless the rest of the routes are updated to the new format, this should use the same format as the others.

app/Http/Kernel.php
```diff
@@ -64,6 +64,7 @@ class Kernel extends HttpKernel {
         'signed'        => \Illuminate\Routing\Middleware\ValidateSignature::class,
         'throttle'      => \Illuminate\Routing\Middleware\ThrottleRequests::class,
         'verified'      => \Illuminate\Auth\Middleware\EnsureEmailIsVerified::class,
+        'manage_files'  => \App\Http\Middleware\CheckPower::class.':manage_files',
```
itinerare: Is there a reason this is set up like this rather than using the existing "power" middleware above?

composer.json
```diff
@@ -28,7 +29,8 @@
         "socialiteproviders/tumblr": "^4.1",
         "socialiteproviders/twitch": "^5.3",
         "spatie/laravel-feed": "^4.1",
-        "spatie/laravel-honeypot": "^4.1"
+        "spatie/laravel-honeypot": "^4.1",
+        "tinymce/tinymce": "5.4.2"
```
itinerare, line 33: It might be better to handle this via npm and include it in what mix handles... Probably out of scope for this PR, though.

itinerare: All told, it might be easier to update tinyMCE *that* way in a separate PR, and *then* apply these changes on top... as it is, this is a tough PR to review because the signal/noise ratio is a bit dire due to the update changes.

Alliionart: Would it be worth it to even update tinymce to a higher version like v7 for the [enhanced code editor](https://www.tiny.cloud/docs/tinymce/7/advcode/) to merge that into core?  Adding the [reference](https://www.tiny.cloud/docs/tinymce/latest/npm-projects/) here

itinerare: I've had issues with tinyMCE versions past... 6.7.3, it looks like? where the toolbar likes to vanish. though it may be a quirk of vite or something. But, assuming that doesn't kick in, potentially?

preimpression: @itinerare Did it load without the toolbar and take way too long to have the toolbar appear? I had that issue and seemed to resolve it by removing addons/plugins I didn't necessarily use (spoiler, emojis, etc) and now it loads much faster. Potentially a similar cause?

itinerare: The toolbar just never loaded-- but this is, again, only on versions after the stated one, so not something that should impact LK unless you've gone out of your way to update it.

SpeedyD:
I'm having trouble understanding why this seems entirely necessary.. is there a way to disable this function entirely?

itinerare:
Yeah, part of the notion is that it be behind a config toggle.

SpeedyD:
Ah, I assume that'll be coming soon, then? Cause so far the only thing I've seen restricting it is rank powers.

PR #1337, opened 2025-09-30
Allows items to be a cost in shops

ScuffedNewt:


config/lorekeeper/ext-tracker/shop_features.php
SpeedyD, line 1: This file should not be here, since it's a core ext now?

ScuffedNewt: It's there because this is the extension branch, it would be preferable for it to be removed after the merge Vs being removed in the source branch and readded

app/Services/ShopManager.php
```diff
@@ -166,6 +167,47 @@ public function buyStock($data, $user) {
...
+
+                            $stackQuantity = $data['stack_quantity'][$userItemStackId] ?? $stack->count;
+                            $requiredQuantity -= $stackQuantity;
+                            $selected[] = [
+                                'stack'    => $stack,
+                                'quantity' => $stackQuantity,
+                            ];
+                        }
+                    } else {
+                        $stacks = UserItem::where('user_id', $user->id)->where('item_id', $cost->item->id)->where('count', '>', '0')->get();
```
itinerare: This broadly looks great! My only worry re automatically debiting items from relevant stacks is that there's a possibility that portions of stacks that aren't available (tied up in a trade or similar) get automatically selected here... I would expect that wouldn't be an issue if they're manually selected (since the inventory select widget shouldn't allow selecting unavailable quantitie(s) from stacks), at least.

ScuffedNewt: that is true, i think the solution here would be to filter based on the available quantity
I think crafting does something similar so could reference from that either

what is the preferred solution, manual selection ONLY, or follow crafting's implementation, or use a different one based on the `availableQuantity` attribute?

https://github.com/Draginraptor/lorekeeper/blob/extension/crafting_system/app/Services/RecipeManager.php#L129

itinerare: I think having the option of automatic selection is a good idea if it's practical, since it reduces friction on what could potentially be a high-frequency task depending on how sites use this particular feature. I think if you can manage a fairly compact check using `availableQuantity` or something that'd be ideal? I imagine it'd be possible to filter using that...

itinerare (approved):

Draconizations (approved):

PR #1339, opened 2025-10-09
fix: blade typo

MarskyMessier:
All in the title, just a fix in a blade typo.


AW0005 (approved):

LostInProgres (approved):

itinerare:
Oh, I just realized that this issue is also present on the release branch-- so this should target that instead.

itinerare (approved):

PR #1344, opened 2025-10-19
fix(shops): Inactive shop refreshes as Active

SpeedyD:
When editing a shop that was previously set as Inactive, the shop would mysteriously be toggled to Active (even if still Inactive). This causes admins undue stress from having to recheck every single time they edit an inactive shop.

Turns out the `isActive` check bounces against `is_active`. So to fix this, all references to `isActive` have been renamed to `isActiveCheck` for shops, and an additional line was added for actual inactive shops to also show as inactive.

Additionally, a minor text change was added as Timed Shop settings override the is_active option.


ScuffedNewt (approved):
👏🛳️

itinerare (approved):

Draconizations (approved):

PR #1345, opened 2025-10-20
feat(comments): add class definition in css for 'btn-faded' 

MarskyMessier:
This is a proposition to remove an unused class (`btn-faded`) in a blade file pertaining to comments - I am unsure as to why or how that class wriggled its way in, but it's not used elsewhere nor referenced in the css files. (Best guess being ToyHouse code being referenced and forgotten at some point)

Given it's not an "official" bootstrap color either, I replaced it by the secondary color and added back the outline to all buttons (reply, edit & feature). If it feels too crowded we could just make them links instead of buttons and tie text colors to them. Either way just a small fix from me yelling at clouds again.

### DIFF
--- a/resources/views/comments/_actions.blade.php
+++ b/resources/views/comments/_actions.blade.php
@@ -3,16 +3,16 @@
     <div class="my-1 row justify-content-between no-gutters">
         <div class="col-auto">
             @can('reply-to-comment', $comment)
-                <button data-toggle="modal" data-target="#reply-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-uppercase"><i class="fas fa-comment"></i><span
-                        class="ml-2 d-none d-sm-inline-block">Reply</span></button>
+                <a data-toggle="modal" href="#reply-modal-{{ $comment->getKey() }}" class=" text-secondary px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-comment"></i><span
+                        class="ml-2 d-none d-sm-inline-block">Reply</span></a>
             @endcan
             @can('edit-comment', $comment)
-                <button data-toggle="modal" data-target="#comment-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-uppercase"><i class="fas fa-edit"></i><span
-                        class="ml-2 d-none d-sm-inline-block">Edit</span></button>
+                <a data-toggle="modal" href="#comment-modal-{{ $comment->getKey() }}" class="text-secondary px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-edit"></i><span
+                        class="ml-2 d-none d-sm-inline-block">Edit</span></a>
             @endcan
             @if (((Auth::user()->id == $comment->commentable_id && $comment->commentable_type == 'App\Models\User\UserProfile') || Auth::user()->isStaff) && (isset($compact) && !$compact))
-                <button data-toggle="modal" data-target="#feature-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-success text-uppercase"><i class="fas fa-star"></i><span
-                        class="ml-2 d-none d-sm-inline-block">{{ $comment->is_featured ? 'Unf' : 'F' }}eature Comment</span></button>
+                <a data-toggle="modal" href="#feature-modal-{{ $comment->getKey() }}" class="text-success px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-star"></i><span
+                        class="ml-2 d-none d-sm-inline-block">{{ $comment->is_featured ? 'Unf' : 'F' }}eature Comment</span></a>
             @endif
             @can('delete-comment', $comment)
                 <button data-toggle="modal" data-target="#delete-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-outline-danger text-uppercase"><i class="fas fa-minus-circle"></i><span
@@ -22,10 +22,10 @@
         <div class="col-auto text-right">
             {{-- Likes Section --}}
             <a href="#" data-toggle="modal" data-target="#show-likes-{{ $comment->id }}">
-                <button href="#" data-toggle="tooltip" title="Click to View" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded">
+                <a href="#" data-toggle="tooltip" title="Click to View" class="px-3 py-2 px-sm-2 py-sm-1">
                     {{ $comment->likes()->where('is_like', 1)->count() -$comment->likes()->where('is_like', 0)->count() }}
                     {{ $comment->likes()->where('is_like', 1)->count() -$comment->likes()->where('is_like', 0)->count() !=1? 'Likes': 'Like' }}
-                </button>
+                </a>
             </a>
             {!! Form::open(['url' => 'comments/' . $comment->id . '/like/1', 'class' => 'd-inline-block']) !!}
             {!! Form::button('<i class="fas fa-thumbs-up"></i>', [
--- a/public/css/lorekeeper.css
+++ b/public/css/lorekeeper.css
@@ -1,3 +1,7 @@
+:root {
+    --btn-faded : transparent;
+}
+
 /**************************************************************************************************
 
     Font definitions
@@ -376,6 +380,35 @@ main > .row {
 .btn.disabled {
     opacity: 0.5;
     cursor: not-allowed;
+}
+
+.btn-faded,
+.btn-faded.disabled,
+.btn-faded:disabled {
+    color: inherit;
+	background-color: var(--btn-faded);
+	border-color: color-mix(in srgb,var(--btn-faded),#000 15%);
+}
+
+.btn-faded:hover,
+.btn-faded:not(:disabled):not(.disabled):active,
+.btn-faded:not(:disabled):not(.disabled).active,
+.show>.btn-faded.dropdown-toggle  {
+	background-color: color-mix(in srgb,var(--btn-faded),#000 10%);
+	border-color: color-mix(in srgb,var(--btn-faded),#000 10%);
+}
+
+.btn-faded:focus,
+.btn-faded.focus {
+	border-color: color-mix(in srgb,var(--btn-faded),#000 20%);
+}
+
+.btn-faded:not(:disabled):not(.disabled):active:focus,
+.btn-faded:not(:disabled):not(.disabled).active:focus,
+.show>.btn-faded.dropdown-toggle:focus,
+.btn-faded:focus,
+.btn-faded.focus  {
+	box-shadow: 0 0 0 0.2rem color-mix(in srgb,var(--btn-faded),#000 10%);
 }
 
 .pagination {


preimpression:
Could you provide a screenshot of the differences before/after, please? iirc the outlines ended up making things a bit busy visually, but it has been some time since I last saw vanilla comments.

resources/views/comments/_actions.blade.php
```diff
@@ -3,15 +3,15 @@
     <div class="my-1 row justify-content-between no-gutters">
         <div class="col-auto">
             @can('reply-to-comment', $comment)
-                <button data-toggle="modal" data-target="#reply-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-uppercase"><i class="fas fa-comment"></i><span
+                <button data-toggle="modal" data-target="#reply-modal-{{ $comment->getKey() }}" class="btn btn-outline-secondary btn-sm px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-comment"></i><span
```
SpeedyD: I don't think setting `btn-outline-secondary` is that good an idea. May I suggest the following instead?
```suggestion
                <button data-toggle="modal" data-target="#reply-modal-{{ $comment->getKey() }}" class="btn text-secondary btn-sm px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-comment"></i><span
```

resources/views/comments/_actions.blade.php
```diff
@@ -3,15 +3,15 @@
...
                         class="ml-2 d-none d-sm-inline-block">Reply</span></button>
             @endcan
             @can('edit-comment', $comment)
-                <button data-toggle="modal" data-target="#comment-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-uppercase"><i class="fas fa-edit"></i><span
+                <button data-toggle="modal" data-target="#comment-modal-{{ $comment->getKey() }}" class="btn btn-outline-secondary btn-sm px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-edit"></i><span
```
SpeedyD: Same here, frankly, but to have Edit stand out from Reply, how about this..
```suggestion
                <button data-toggle="modal" data-target="#comment-modal-{{ $comment->getKey() }}" class="btn text-info btn-sm px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-edit"></i><span
```
Info is a lighter blue, it just feels softer :)

resources/views/comments/_actions.blade.php
```diff
@@ -3,15 +3,15 @@
...
                         class="ml-2 d-none d-sm-inline-block">Edit</span></button>
             @endcan
             @if (((Auth::user()->id == $comment->commentable_id && $comment->commentable_type == 'App\Models\User\UserProfile') || Auth::user()->isStaff) && (isset($compact) && !$compact))
-                <button data-toggle="modal" data-target="#feature-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-success text-uppercase"><i class="fas fa-star"></i><span
+                <button data-toggle="modal" data-target="#feature-modal-{{ $comment->getKey() }}" class="btn btn-sm btn-outline-success px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-star"></i><span
```
SpeedyD: And this was already handled well, just remove `btn-faded`.
```suggestion
                <button data-toggle="modal" data-target="#feature-modal-{{ $comment->getKey() }}" class="btn btn-sm text-success px-3 py-2 px-sm-2 py-sm-1 text-uppercase"><i class="fas fa-star"></i><span
```

SpeedyD (changes requested):
I admit having no idea what `btn-faded` is either and can't even find anything related to Bootstrap, but I disagree with adding outlines. It makes it feel busier, more crowded.

So I've added suggestions below, basically instead of outline, just make it `text-secondary`, `text-info` and `text-success` respectively, to make them stand out from one another without being busier. :)

MarskyMessier:
I think no outline is fine, I personally added them as it was how the delete button was, and I think it might look a bit awkward that only delete & likes have the outline but not the rest of the buttons. I do like the decluttering however, I'm in for the text itself, just asking if maybe it would be more judicious to remove the outline from the delete button ? 
Asking for more input before I review Speedy's changes!

The screenshot below show both the outline changes (top) and Speedy's proposal (bottom).
Screenshot 2025-10-27 013230
Screenshot 2025-10-27 013344

SpeedyD:
> I think no outline is fine, I personally added them as it was how the delete button was, and I think it might look a bit awkward that only delete & likes have the outline but not the rest of the buttons. I do like the decluttering however, I'm in for the text itself, just asking if maybe it would be more judicious to remove the outline from the delete button ? Asking for more input before I review Speedy's changes!

Personal reason is: The Delete button **should** stand out because out of the four buttons, it's the one that is the dangerous one, and the others technically just.. function and read as links instead.

As for likes, it's.. detached from the rest. It does not perceive similar to the other buttons, so I feel like that one is fine by itself.

Ironically, the '# Likes' text is _also_ a button, but it doesn't stand out like the rest.. 🤷

preimpression:
This conundrum has given me thoughts on the overall UI of comments and trying to make them less cluttered/etc, so I'm going to fiddle around with ui/ux for comments when I have free time, beyond what you've suggested here. 

Thank you for the push to look into this!

MarskyMessier:
I discovered while editing this that the v3 version of bootstrap is slightly older than dev, meaning that the `btn` class does not have a background defined in the app.css while develop has that. This causes the button to default to a grey color (assumedly done via the browser and not the site itself). With the mindset of decluttering the view a bit, I decided to change the buttons for straight-up links instead so they don't get that grey coloring on v3. This does not change anything functionality-wise and is purely aesthetic. 

> Personal reason is: The Delete button should stand out because out of the four buttons, it's the one that is the dangerous one, and the others technically just.. function and read as links instead.
> As for likes, it's.. detached from the rest. It does not perceive similar to the other buttons, so I feel like that one is fine by itself.

That's honestly fine ! After some thoughts I do think keeping the delete button as a button is better than anything, best to be cautious than anything.

SpeedyD:
Let's see.. main is on v4.1.3, and so is release.. and develop is on v4.6.2..

..From what I can see, both of those versions of Bootstrap should allow this, and.. I guess I see no problem with this? Should visually remain the same as the screenshot you did of my proposal?

I'm currently not in the right mental state to properly test this, but from what I can tell it _should_ work. I'll review in the morning.


<details><summary>Wait..</summary>
<p>Why does my brain do this to me? Able to quickly research which versions we're on and whether or not Bootstrap allows this, but when it comes to just.. copy pasting a bit of code and testing, I'm suddenly too exhausted mentally.. what the heck..</p>
</details>

resources/views/comments/_actions.blade.php
```diff
@@ -3,16 +3,16 @@
     <div class="my-1 row justify-content-between no-gutters">
         <div class="col-auto">
             @can('reply-to-comment', $comment)
-                <button data-toggle="modal" data-target="#reply-modal-{{ $comment->getKey() }}" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded text-uppercase"><i class="fas fa-comment"></i><span
-                        class="ml-2 d-none d-sm-inline-block">Reply</span></button>
+                <a data-toggle="modal" href="#reply-modal-{{ $comment->getKey() }}" class=" text-secondary px-3 py-2 px-sm-2 py-sm-1 text-uppercase btn-faded"><i class="fas fa-comment"></i><span
```
SpeedyD: Missed a btn-faded here.

SpeedyD (approved):
Alright, got to testing, there's only ONE minor thing you overlooked.

Call this an approve once that's finished. :)

MarskyMessier:
> Alright, got to testing, there's only ONE minor thing you overlooked.
> 
> Call this an approve once that's finished. :)

That should now be fixed !

SpeedyD (approved):
100% approve now

AW0005:
I think for the sake of keeping this PR scoped to it's original changes as closely as possible, I'm going to advocate for putting the "Reply" button back to grey.

Unfortunately the text-info color isn't accessible:
image

text-primary is slightly better but still doesn't meet the 4.5 ratio recommendation for keeping text accessible. To fix that properly any suggestions I would have, would be a massive amount of scope creep on this PR's original intentions and since it sounds like Uri is going to be looking into improving this UI in general anyway might be best to leave any unnecessary style adjustments for that endeavor.

Draconizations:
> I think for the sake of keeping this PR scoped to it's original changes as closely as possible, I'm going to advocate for putting the "Reply" button back to grey.

I agree with this, also for the sake of keeping this PR to its original scope. Can also confirm `text-info` is really hard for me to read personally.

MarskyMessier:
I reverted the changes from info back to secondary, so it's back to how the colors were originally

Draconizations (approved):

SpeedyD (approved):

LostInProgres (approved):

perappu (approved):

ScuffedNewt (approved):

resources/views/comments/_actions.blade.php
```diff
@@ -22,10 +22,10 @@ class="ml-2 d-none d-sm-inline-block">Delete</span></button>
         <div class="col-auto text-right">
             {{-- Likes Section --}}
             <a href="#" data-toggle="modal" data-target="#show-likes-{{ $comment->id }}">
-                <button href="#" data-toggle="tooltip" title="Click to View" class="btn btn-sm px-3 py-2 px-sm-2 py-sm-1 btn-faded">
+                <a href="#" data-toggle="tooltip" title="Click to View" class="px-3 py-2 px-sm-2 py-sm-1">
```
MarskyMessier: Noted that I've just noticed an issue with the like counter button where you cannot access it if the wrapped link is an `a` class instead of `button`. 
Any suggestion to make that and the tooltip work is appreciated! Atm I haven't found a quick way to make it work with the tooltip.

SpeedyD: @MarskyMessier At this point I'd suggest closing this PR and remaking one to just remove the `btn-faded` class.

The scope of this PR has been to hell and back and even though the other changes are viable, they're also somewhat unnecessary.

If you do find them necessary, I'd suggest making new PRs for additional changes.

SpeedyD (changes requested):
Repeating my strong suggestion to recreate this PR, and thus removing my approval.

MarskyMessier:
> I discovered while editing this that the v3 version of bootstrap is slightly older than dev, meaning that the `btn` class does not have a background defined in the app.css while develop has that. This causes the button to default to a grey color (assumedly done via the browser and not the site itself). With the mindset of decluttering the view a bit, I decided to change the buttons for straight-up links instead so they don't get that grey coloring on v3. This does not change anything functionality-wise and is purely aesthetic.

@SpeedyD I mentionned earlier in the replies why the scope went to changing it to another class. I'm happy to redo a PR, but this might cause issues down the line with `btn` on its own. I'll test it again, but I don't want to ship something I know is half broken from the get-go.

SpeedyD:
Yes, I'll agree with that point: At Bootstrap 4.1.3 it's only outlines that got transparent backgrounds, and somewhere between that version and 4.6.2 they made it so that all btn classes have transparent backgrounds (unless a specific color is added)...

But.. if the button transparency is the issue, why not just add the `.bg-transparent` class?

All you'd need to do then is.. effectively replace all the `.btn-faded` classes into `.bg-transparent` classes. Completely different scope, no need to swap html code.. Easier solution?

SpeedyD:
@MarskyMessier: Have you seen and considered my previous comment? :)

MarskyMessier:
Sorry for the delay, I had issues with an extension prior to the comment that made me unable to properly check back my PRs on my dev repo, I've reverted the changes back to buttons and swapped the btn-faded for bg-transparent which should be better as you suggested

itinerare:
Got to taking a look at this finally (thanks for your patience/the nudge)... I agree with addressing the non-functional class/actually specifying behavior there, given that currently the styling of them is just falling back to browser defaults, which is... not terribly desirable.

In Vivaldi, for example:
image

I'm just not sure I agree with removing the background entirely. Having at least something there is useful for conveying that it's a button with some utility associated, I think, since it keeps it consistent with the rest of the UI.

If anything, it may be simpler to just... add some functionality for `.btn-faded`?

MarskyMessier:
> I'm just not sure I agree with removing the background entirely. Having at least something there is useful for conveying that it's a button with some utility associated, I think, since it keeps it consistent with the rest of the UI.

> If anything, it may be simpler to just... add some functionality for .btn-faded?

That's a good question, however I'm not sure how to convey that? The outline proposition wasn't working, and the default grey background isn't ideal on visibility either. I doubt having full colored buttons (secondary and success) would be better either. Besides making them look like how any buttons do for bootstrap, I'm at a loss as to how convey it better that it's a button...

perappu:
Sorry if this is a goofy/clueless answer, but couldn't you just put `.bg-white` on the buttons to solve the gray/default background thing and call it a day? It's not the perfect solution, but it gets the job done.

MarskyMessier:
> Sorry if this is a goofy/clueless answer, but couldn't you just put `.bg-white` on the buttons to solve the gray/default background thing and call it a day? It's not the perfect solution, but it gets the job done.

I'd (personally) rather have the faded button be declared white in the css itself mostly for compatibility on LKs. I like to keep in mind people who will edit the style and have to edit yet another class meant for light mode the likes of `bg-light` or light classes in general (hi I'm people), which can also get confusing for a class meant to be white.
Also it's visually the same as doing `bg-transparent` unless the body background is changed itself because LK's body is white by default. 

Best I can think of is a white low opacity background for `btn-faded ` say like 0.5 that would show as white anyway for default LK yet still retain a button background for any other ARPGs that have their background color and/or css edited. Let me know if that's sensical !

perappu:
> Best I can think of is a white low opacity background for `btn-faded ` say like 0.5 that would show as white anyway for default LK yet still retain a button background for any other ARPGs that have their background color and/or css edited.

Yeah I think that'd be fine. I forgot about how some people editing their theme might not have access to the code in a way that they can just remove the `bg-white` class...

SpeedyD:
...I'm kind of wondering at this point that _perhaps_ we should just remove btn-faded and leave it there. Let site admin sort of their css changes and all..

Since this entire PR thread is just opinions going back and forth and clashing slightly. I gave bg-transparent as alternative suggestion to remove the.. 'default' problem.. and apparently it's still not great.

So.. I'm strongly considering just.. letting it sit. This may be something we're not going to be agreeing on.. 🤔

perappu:
I mean I do think we reached a consensus finally, which is to just actually implement a btn-faded into lorekeeper.css in a way that's simple to override with custom styling.

SpeedyD:
> I mean I do think we reached a consensus finally, which is to just actually implement a btn-faded into lorekeeper.css in a way that's simple to override with custom styling.

So in the end, hilariously enough; the PR to `remove unused "btn-faded" class from buttons`... will in the end actually _retain_ the btn-faded class and instead add it to the CSS :P

MarskyMessier:
I think this is the best move I can do : I've added a root variable in the css, alongside the `btn-faded` class so it displays correctly and has a purpose! 
The background is transparent, and has a slight "shadow" overlay on hover like general bootstrap buttons. I've added an outline too that isn't massively overwhelming but should be enough to indicate that they're buttons. The color is inherited from the body to avoid adding too much to edit. People can easily override it via the declared root variable if they need to.

Here's how it looks :
image
(with the normal, hovered, focused and disabled state to showcase above)
<br>
image

perappu (approved):

itinerare (approved):

PR #1346, opened 2025-10-20
feat: extra dropdown options

ScuffedNewt:


SpeedyD (commented):
Looks good to me- only reason I'm not approving is because my brain is foggy and I can't give a proper in-depth review, sorry.

I'll try to look through it again when I feel better. ^^;

Draconizations (approved):
LGTM. If/when this gets merged I probably can make the push to update and PR [multiple trait subtypes](https://github.com/Draconizations/lorekeeper/tree/extension/multiple-trait-subtypes-v3.1) as well.

config/lorekeeper/extensions.php
```diff
@@ -83,8 +83,16 @@
         'currency_id' => 1,
     ],
 
-    // Organised Traits Dropdown - Draginraptor
-    'organised_traits_dropdown' => 0,
+    // Organised Traits Dropdown - Draginraptor, ScuffedNewt
+    'organised_traits_dropdown' => [
+        'enable'           => 1,
+        'display_species'  => 0, // displays species
+        'display_subtype'  => 0, // displays subtype, SPECIES DISPLAY MUST ALSO BE ENABLED FOR THIS TO WORK.
```
itinerare:
```suggestion
        'display_subtype'  => 0, // displays subtype. NOTE: species display must also be enabled for this to work.
```
nit: whole note should probably not be caps (though some emphasis is probably useful)

ScuffedNewt: Good point!

app/Models/Feature/Feature.php
```diff
@@ -307,9 +310,41 @@ public static function getDropdownItems($withHidden = 0) {
...
                 foreach ($features as $id  => $feature) {
-                    $grouped[$category][$id] = $feature['name'];
+                    $grouped[$category][$id] = $feature['name'].
+                    (
+                        config('lorekeeper.extensions.organised_traits_dropdown.display_species') && $feature['species_id'] ?
+                        ' <span class="text-muted"><small>'.$feature['species']['name'].'</small></span>'
+                        : ''
+                    ).
+                    (
+                        config('lorekeeper.extensions.organised_traits_dropdown.display_subtype') && $feature['subtype_id'] ?
```
itinerare, line 339: Also, as is the note in the config file is not true... this works fine without species display being enabled. Probably either fix that or remove the note?

app/Models/Feature/Feature.php
```diff
@@ -307,9 +310,41 @@ public static function getDropdownItems($withHidden = 0) {
                 return in_array($value, array_keys($grouped), true);
             });
 
+            // Sort by rarity if enabled
+            if (config('lorekeeper.extensions.organised_traits_dropdown.rarity.enable') && config('lorekeeper.extensions.organised_traits_dropdown.rarity.sort_by_rarity')) {
+                foreach ($grouped as $category => &$features) { // &$features to modify the array in place
+                    uasort($features, function ($a, $b) {
```
itinerare, line 316: Having played around with it... I think the UX would be better it sorted, essentially, common > rare (reverse from how it is currently)

config/lorekeeper/extensions.php
```diff
@@ -83,8 +83,16 @@
...
-    // Organised Traits Dropdown - Draginraptor
-    'organised_traits_dropdown' => 0,
+    // Organised Traits Dropdown - Draginraptor, ScuffedNewt
+    'organised_traits_dropdown' => [
+        'enable'           => 1,
+        'display_species'  => 0, // displays species
+        'display_subtype'  => 0, // displays subtype. NOTE: species display must also be enabled for this to work.
+        'rarity'           => [
+            'enable'         => 1, // If enabled, displays trait rarity in the dropdown.
+            'sort_by_rarity' => 1, // If enabled, sorts traits by rarity in the dropdown.
```
itinerare: *This* could use a note that it does not function without rarity display being enabled.

itinerare:
Having played around with it, this is *really good* QoL. Just the few issues to hammer out, and then it should be set!

app/Models/Feature/Feature.php
ScuffedNewt: ah yes i remember i did play around with this, would make sense that I didn't update it

app/Models/Feature/Feature.php
Draconizations: Oh, yeah agreed. I missed that when looking over the code initially (this is what I get for reviewing PRs when sick, good thing we require two).

ScuffedNewt: makes sense!

config/lorekeeper/extensions.php
ScuffedNewt: do you think it is better to either:
- make it not function unless display rarity is enabled
or
- make it function even if it isn't enabled?

itinerare: I think it would be fine either way? It's not very obvious when rarity display isn't enabled, but folk might like if it functioned without it... so maybe the latter option just for the extra flexibility.

app/Models/Feature/Feature.php
itinerare: Haha, yeah, fair. I'm not too fussed as long as the note matches the functionality, here

config/lorekeeper/extensions.php
Draconizations: I know I'd personally prefer the latter. I don't think allowing them both to function separately would hugely increase confusion?

itinerare (approved):

PR #1348, opened 2025-10-22
Fix character sale value throwing SQL error instead of validating input

perappu:
A small addition of some validation rules that prevents an error like this from happening when a non-decimal value is entered in for character sale_value:
image

And replaces it with this:
image


itinerare (approved):

SpeedyD (approved):
I'll approve of this, cause it makes sense, but I'm starting to wonder if setting the input to a type number with a step of 0.01 wouldn't also be useful..

And perhaps add an additional Javascript validation, like `.valueAsNumber.toFixed(2)` ? Well, the step should be enough..

ScuffedNewt (approved):

PR #1352, opened 2025-11-01
Feature: admin notifications

SpeedyD:
Got @preimpression's permission to port this, and so far that is purely that. Still gonna look coding through to see if things are necessary.. like, I think "UserCharacterLog" is not necessary for this? Gonna have to test that..

Also, need to still add toggles for users and overall, since this is somewhat obtrusive.

Also thinking of just making the circles badges? I think the huge circle is a bit much and bootstrap's badges are perfect for this.. Opinions on that welcome.


SpeedyD:
I've opted to do other changes later. Currently I'm just.. This needs to get out. I don't know why I left it hanging.

ScuffedNewt (approved):

MarskyMessier (approved):

itinerare:
This should have toggles before being merged, imo.

SpeedyD:
Okay, so, updates.

1. In the extensions.php config file, the entirety of Admin Notifications can be disabled. It's enabled by default.
2. An admin can, for themselves, enable/disable the number, or make it slightly more compact.
3. I've added a check if classes exist for Affiliates and the Donation Shop- both extensions still require to be added as 'use' case in the User model, but at least the notification side should be automated now.

I hope this is satisfactory. I've been intending on doing this since I stated 'draft for now' in the intro message, but simply never got to them.. until now.

SpeedyD:
aaand minor merge conflict fixed again.

app/Models/User/User.php
```diff
@@ -18,8 +20,10 @@
 use App\Models\Notification;
 use App\Models\Rank\Rank;
 use App\Models\Rank\RankPower;
+use App\Models\Report\Report;
 use App\Models\Shop\ShopLog;
 use App\Models\Submission\Submission;
+use App\Models\Trade;
```
perappu:
```suggestion
use App\Models\Trade\Trade;
```

Needs to be updated to this due to https://github.com/lk-arpg/lorekeeper/pull/1199

app/Models/User/User.php
```diff
@@ -722,4 +726,37 @@ public function getSubmissions($user = null) {
...
     }
+
+    /**
+     * Check if there are any Admin Notifications.
+     *
+     * @param mixed $user
+     *
+     * @return int
+     */
+    public function hasAdminNotification($user) {
```
perappu, line 761: I hate to be like this, but this function creates an _extremely_ gnarly amount of queries if you're on a user that has a good amount of permissions (and this is only a sampling, AND it's on the admin account so it's not making all of the rank-related DB calls that an ordinary user would have). It's a problem I've always had with the original extension. And since this appears in the navbar, it's on _every_ page:

Image

Is there.... any way to make this better? Like at all? I feel like this is starting to get into some of the real flaws we have with hasPower() at the moment...

perappu (changes requested):

app/Models/User/User.php
```diff
@@ -722,4 +726,37 @@ public function getSubmissions($user = null) {
...
+        $count[] = $this->hasPower('manage_characters') ? CharacterDesignUpdate::characters()->where('status', 'Pending')->count() : 0; // designCount
+        $count[] = $this->hasPower('manage_characters') ? CharacterDesignUpdate::myos()->where('status', 'Pending')->count() : 0; // myoCount
+        $count[] = $this->hasPower('manage_characters') ? CharacterTransfer::active()->where('is_approved', 0)->count() : 0; // transferCount
+        $count[] = $this->hasPower('manage_characters') ? Trade::where('status', 'Pending')->count() : 0; // tradeCount
+        $count[] = $this->hasPower('manage_characters') ? Trade::where('status', 'Pending')->count() : 0; // tradeCount
+        $count[] = $this->hasPower('manage_submissions') ? GallerySubmission::pending()->collaboratorApproved()->count() : 0; // galleryCount
+        $count[] = $this->hasPower('manage_reports') ? Report::where('status', 'Pending')->count() : 0; // reportCount
+        $count[] = $this->hasPower('manage_reports') ? Report::assignedToMe($this)->count() : 0; // assignedReportCount
+
+        // If Adoption Center is installed:
```
perappu: I'll be honest, I'm not 100% sure how I feel about adding accommodations for specific extensions in core, even if it was in the original extension. I think this is the only time in the entire codebase that we'd be doing so, so it feels a bit odd... I don't really foresee either of the ones here making it into core either.

itinerare: Not per se; the command for migrating subtypes to the new format that supports multiple has handling for the couple of multiple-subtypes commands. That's not so much "standing support" as "support to migrate away from", though....

perappu: That's true, and when I thought about it more we do have that 'criteria' in ParsePostRequestFields, but that extension is imminently coming into core (and has been a candidate for core for a very long time), so...

itinerare: Yeah... that one's been a notable exception on the grounds of intended inclusion in core for a while.

app/Models/User/User.php
SpeedyD: Right, wasn't aware. Thanks.

app/Models/User/User.php
SpeedyD: Boy howdy, that's a good question. Since this is straight up a port from the ext, I haven't actually thought about many changes.. But you are right.. perhaps something needs to be done.

..what if it saves the numbers somewhere, like in a new database table, pull that instead.. and.. Hrm.. it would be updated if.. anyone submits something? Might mean a lot of new functions spread around..

...I'd have to think about it. Suggestions can always be PR'd to the https://github.com/SpeedyD/lorekeeper/tree/feature/admin-notifications branch itself, haha.. 😅

app/Models/User/User.php
SpeedyD: Mrh, something to consider, I suppose, but until we can sort out the new query handling, I think I'll just let this one sit.

app/Models/User/User.php
perappu: If you get a chance, can you update this branch to latest dev to grab the hasPower() changes from https://github.com/lk-arpg/lorekeeper/pull/1479? It won't solve everything, but I'd be curious to pull it again and see if the queries were at all improved from that alone.

SpeedyD: Updated to latest dev. Only change I had to enforce was removing
`use App\Models\Rank\RankPower;`
`use App\Models\Report\Report;`

I don't exactly have the time or mental space rn to fix up the queries. I once again would like to ask/suggest perhaps submitting a PR to the above branch. ^^;

app/Models/User/User.php
SpeedyD: ..thinking about it now, I feel like once this is actually _in_ core, it would the _extension developer's_ responsibility to add this, not ours. So I am going to remove this.

app/Models/User/User.php
perappu: No no that's exactly why I asked, I'd like to look at it myself & see if it improved just from the latest core updates alone. Thanks for grabbing that!

SpeedyD: In that case, no problem at all, haha. Thanks for looking into it. :)

SpeedyD:
Thanks to @liwoyadan, the queries are heavily improved, and are cached.

Thanks again, Min!

PR #1356, opened 2025-11-10
chore(workflows): use self-contained workflows

itinerare:
I'm not liable to maintain the external workflows at this point (and would probably set things up differently now if I did), so I figure it makes most sense to just incorporate the full workflows back into the repo.

An accompanying PR targeting develop will be along shortly!

### DIFF
--- a/.github/workflows/Lint.yml
+++ b/.github/workflows/Lint.yml
@@ -9,19 +9,87 @@ on:
       - '**'
     types:
       - opened
-      - edited
       - synchronize
       - reopened
 
 jobs:
   pint:
-    uses: itinerare/github-actions/.github/workflows/pint.yml@main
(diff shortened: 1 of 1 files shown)


SpeedyD (approved):
Don't see a problem here. I'm kinda glad I looked into workflows for a different project of mine, haha, cause now I understand what I'm seeing. :)

Draconizations (approved):
Yeah this makes sense to me. Much easier for the LK community to contribute to the workflows as well if set up this way (and doesn't put most of the burden on you).

itinerare:
Yeah, at the time that change was made, no one else in the community had any idea what was going on in there-- so making it easier for myself was the best way about it. I'm happy that's been changing!

PR #1361, opened 2025-11-17
feat(shops): create takeUserAssets function

perappu:
Resolves the bug where, when deducting multiple items in a single cost group, it tries to deduct stacks multiple times. (Issue https://github.com/lk-arpg/lorekeeper/issues/1351)
(so this is both a fix _and_ a feat in a way? haha)

Some thoughts:
- Reverts fillUserAssets() to no longer have the deduction code, so $selected is no longer needed as a function parameter.
- I thought about adding an accompanying takeCharacterAssets, but since we only remove currency from characters at the moment, it felt unnecessary at this time.
- Changed the exception messages for failing to deduct costs to be a little more clear.


app/Helpers/AssetHelpers.php
```diff
@@ -419,6 +390,56 @@ function fillUserAssets($assets, $sender, $recipient, $logType, $data, $selected
...
+        if ($key == 'items' && count($contents)) {
+            $service = new App\Services\InventoryManager;
+            // do not loop the assets here, just the stackdata. otherwise it will deduct stacks multiple times
+            if (!$selected) {
+                flash('No selected item found for debiting.')->error();
+
+                return false;
+            }
+
+            foreach ($selected as $stackData) {
```
ScuffedNewt, line 438: haven't confirmed, but we're certain there is no validation required against $contents here? ex that the quantity of selected stacks == contents=>assets['quantity']?

should we have this as extra validation?

ScuffedNewt (commented):
fit for purpose, will follow up with proposed changes from discord at a later date--but let's get this out there so it's not affecting people

only one comment that i would like to be double checked / discussed but after that LGTM

app/Helpers/AssetHelpers.php
perappu: I think we should trust that the stacks add up to the quantity provided (otherwise we'd be filtering and re-counting the stacks... feels like an unnecessarily expensive operation, but I might be having a failure of imagination here) but a validation that all selected items are present within $contents would be sufficient I think

perappu: Yeah okay I did some experiments on different ways we could compare the quantities, namely by simplifying them both to [item_id => quantity] and seeing if the arrays are equal.

I put in a solution that works but I don't know how I feel about it?

Simplifying $contents is (relatively...) straightforward, it's just `array_combine(array_keys($contents), array_map("abs", array_column($contents, 'quantity')))`. We do array_map("abs") because the quantities passed can be negative.

Simplifying $selected is less so. You have to pull the item_id from the object out of the array, map it to the quantity, and then sum up values for the identical keys. I don't think there's a way to do this without storing variables, nested for loops, and/or turning it into a collection and doing map()/mapWithKeys() shenaniganry.

So I took the route of tallying up the $selected as we deduct them to build the [item_id => quantity] array, and then comparing at the end instead. 

**My main concern** is that if there is a discrepancy, if takeUserAssets() is not called from a try-catch-rollback, **it may throw the error but still deduct the quantities**. I'm not 100% sure this is how it will behave though. If it is, I think that could be damning enough to just not do this validation.

Definitely would be open to hearing more opinions about this from multiple people and if the validation is worthwhile at all!

ScuffedNewt: > My main concern is that if there is a discrepancy, if takeUserAssets() is not called from a try-catch-rollback, it may throw the error but still deduct the quantities.

I think we should expect it to be called in the safety of a transaction, that is something we shouldn't have to validate against (we don't anywhere else...)

app/Helpers/AssetHelpers.php
```diff
@@ -419,6 +390,74 @@ function fillUserAssets($assets, $sender, $recipient, $logType, $data, $selected
...
+                    foreach ($service->errors()->getMessages()['error'] as $error) {
+                        flash($error)->error();
+                    }
+
+                    return false;
+                }
+            }
+
+            // Validate deducted quantities against $contents
+            if ($selectedQuantities != array_combine(array_keys($contents), array_map('abs', array_column($contents, 'quantity')))) {
```
ScuffedNewt: I do think there is a way we can do this in one pass before the foreach, let me have a look, but if this gets approval from another user prior to me being able to do that this has my ack to merge

app/Helpers/AssetHelpers.php
perappu: After some [Discord discussion](https://discord.com/channels/678909186683437056/678909186683437059/1443247605793558569) it was decided that it's not worth doing the quantity validation. I added a note in the docblock regarding the verdict & a future warning to other users to do the validation in the containing function.

perappu: > > My main concern is that if there is a discrepancy, if takeUserAssets() is not called from a try-catch-rollback, it may throw the error but still deduct the quantities.
> 
> I think we should expect it to be called in the safety of a transaction, that is something we shouldn't have to validate against (we don't anywhere else...)

Right now it's not strictly mandatory for the companion function fillUserAssets() to be called inside of a transaction, so I don't want to make takeUserAssets() strictly mandatory either. I just think it's way too risky to create a scenario where we know it could explode catastrophically, given that LK is oriented towards a beginner programmer

Draconizations (approved):
LGTM, I figure that validation can always be added later if someone does figure out a good way to do it.

app/Helpers/AssetHelpers.php
perappu: It got approval, so resolving the comment to let this be merged

PR #1363, opened 2025-11-23
chore(deps-dev): update tinyMCE

itinerare:
Updates tinyMCE to 5.10, partly to fix an odd issue some folks have been reporting; this was mentioned this as being a known good version, so.

Tried updating via NPM + mix, found it fussy and not worth potentially breaking folks' setups over. That said, I did get the files themselves via NPM, so that seems to be a viable method in the event that we need to repeat this process down the line.

Seems to work as normal from some local testing, but do probably take it for a spin and make sure all's well.


ScuffedNewt (approved):
comments on discord but generally LGTM

ScuffedNewt (approved):

SpeedyD (approved):

PR #1365, opened 2025-11-25
feat: multiple trait subtypes

Draconizations:
Finally opening a PR for [this ext of mine](https://wiki.lorekeeper.me/index.php?title=Extensions:Multiple_Trait_Subtypes). Similar to multiple character subtypes, but for traits instead. It *should* replace every reference to a `subtype_id` in the Feature model to use the `subtypes` relation instead, but there's always a chance I missed some. The enhanced trait dropdown and trait indexes are supported as well.

Run `php artisan convert-trait-subtype` after migrations. This does drop the `subtype_id` column, so any extensions that reference it will become incompatible.


resources/views/world/_features_index.blade.php
```diff
@@ -22,8 +22,8 @@
                                         <i class="fas fa-eye-slash mr-1"></i>
                                     @endif
                                     {!! $feature->first()->displayName !!}
-                                    @if ($showSubtype && $feature->first()->subtype)
-                                        <br />({!! $feature->first()->subtype->displayName !!} Subtype)
+                                    @if ($showSubtype && count($feature->first()->subtypes))
```
SpeedyD: I had a question regarding this part. If a trait has one or more subtypes that are in fact invisible, would the subtype count still show up?

For instance, if a trait has two subtypes and both are invisible, will it still say `(Subtypes:` ? It looks like it'll say 'None' based on the displaySubtypes function, while it should actually.. show nothing?

This is just what I'm assuming based on the code I'm seeing. May I perhaps suggest replacing the `count($feature->first()->subtypes))` with a new function in the trait model that checks if it has one or more _(specific to the user) visible_ traits? :)

Draconizations: Good catch, I'll get to that shortly, energy depending

app/Http/Controllers/Admin/Data/FeatureController.php
```diff
@@ -350,11 +352,17 @@ public function postDeleteFeature(Request $request, FeatureService $service, $id
      */
     public function getCreateEditFeatureSubtype(Request $request) {
         $species = $request->input('species');
-        $subtype_id = $request->input('subtype_id');
+        $subtype_ids = $request->input('subtype_ids');
+
+        if ($subtype_ids != 'null') {
+            $subtype_ids = explode(',', $subtype_ids);
```
ScuffedNewt: will check but this has a bad smell, don't think we need to explode this

resources/views/world/_features_index.blade.php
Draconizations: Hoping I didn't miss any spot where trait subtypes are displayed to regular users, if not, this should be resolved now. Though instead of making the function a check on whether it should be displayed, it returns the relevant subtypes instead since I do need to check the actual count in one (1) spot.

app/Http/Controllers/Admin/Data/FeatureController.php
Draconizations: No clue what the alternative is, feel free to enlighten me haha

config/lorekeeper/extensions.php
```diff
@@ -34,7 +34,7 @@
         'enable_species_index'   => 0, // Enables the Species Trait Index
         'enable_subtype_index'   => 0, // Enables the Subtype Trait Index
         'enable_universal_index' => 0, // Enables the Universal Trait Index
-        'enable_all_trait_index' => 0, // Enables the All Traits Index
+        'enable_all_trait_index' => 1, // Enables the All Traits Index
```
SpeedyD: This should remain 0 by default. ^^;

(I assume it was turned on for testing?)

app/Http/Controllers/Admin/Data/FeatureController.php
SpeedyD: Couldn't you just pass it as-is and do work with the results on the blade?

resources/views/admin/features/create_edit_feature.blade.php
```diff
@@ -101,16 +101,19 @@
 
         function refreshSubtype() {
             var species = $('#species').val();
-            var subtype_id = {{ $feature->subtype_id ?: 'null' }};
+            var subtype_ids = "{{ !$feature->subtypes->isEmpty() ? implode(',', $feature->subtypes->pluck('id')->toArray()) : 'null' }}";
```
SpeedyD: Following my other comment, if you don't explode you don't need to implode? At least that's what I'm thinking.. @ScuffedNewt can you look this over, perhaps?

ScuffedNewt: Yeah I think you're right & was also my initial thoughts, can confirm though

config/lorekeeper/extensions.php
Draconizations: ...yes. I will turn that back off

resources/views/admin/features/create_edit_feature.blade.php
Draconizations: OH yeah that makes more sense. Thank you both

Draconizations:
so sorry for the commit mess everyone, turns out having two editors open for different LK repos confused me greatly

SpeedyD:
You may also want to resolve conversations you've fixed :)

SpeedyD (commented):
Aight, so, real review time: My brain is exhausted, but I _think_ it looks good? I however cannot in good conscience approve it until I look it over and test properly when NOT exhausted..

Sorry for that. ^^;;

SpeedyD (approved):

ScuffedNewt (approved):

itinerare:
Just needs conflict resolution!

Draconizations:
Will get to that tomorrow! I'm about to head to bed o7

ScuffedNewt (approved):

PR #1366, opened 2025-11-26
fix: SQL error on universal/all trait index if no rarities are present

Draconizations:
Noticed this while messing around on my local which has no rarities. Figured I might as well fix it while I'm at it.

### DIFF
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -444,8 +444,11 @@ class WorldController extends Controller {
         $features = count($categories) ?
             $features->orderByRaw('FIELD(feature_category_id,'.implode(',', $categories->pluck('id')->toArray()).')') :
             $features;
-        $features = $features->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')')
-            ->orderBy('has_image', 'DESC')
+        $features = count($rarities) ?
+            $features->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')') :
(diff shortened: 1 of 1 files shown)


app/Http/Controllers/WorldController.php
```diff
@@ -484,15 +487,17 @@ public function getKitchenSinkFeatures(Request $request) {
...
+        $features = Feature::visible(Auth::user() ?? null);
+
         $features = count($categories) ?
-        $query = Feature::visible(Auth::user() ?? null)->orderByRaw('FIELD(feature_category_id,'.implode(',', $categories->pluck('id')->toArray()).')')
-            ->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')')
-            ->orderBy('has_image', 'DESC')
-            ->orderBy('name')
-            ->get()
-            ->groupBy(['feature_category_id', 'id']) :
-        $query = Feature::visible(Auth::user() ?? null)->orderByRaw('FIELD(rarity_id,'.implode(',', $rarities->pluck('id')->toArray()).')')
```
itinerare, line 494: Just from a quick glance... was it intentional that the visibility checks vanished?

Draconizations: definitely not

Draconizations: upon further inspection, it's still there! but moved to line 490, before count() checks.

itinerare: All good then!

itinerare (approved):

ScuffedNewt (approved):

PR #1367, opened 2025-11-27
Feat(characters): Show only traits belonging to the character's species

SpeedyD:
As well as non-species specific traits.

Can be turned off with new option in extensions.php.

Second commit adds a refresh on species change- also edited the part for subtypes to be more consistent with other code. :)


resources/views/admin/prompts/create_edit_prompt.blade.php
SpeedyD, line 1: ...what the heck, blade formatter? I didn't touch this file XD

app/Http/Controllers/Characters/DesignController.php
```diff
@@ -254,6 +254,19 @@ public function getFeaturesSubtype(Request $request) {
         ]);
     }
 
+    /**
+     * Shows the edit image trait portion of the modal.
+     *
+     * @return \Illuminate\Contracts\Support\Renderable
+     */
+    public function getFeaturesFeature(Request $request) {
```
ScuffedNewt: nit, feel free to ignore
would prefer if this and blade where named a bit more clearly, ex. `getSpeciesFeatures` and `_species_features`

app/Models/Feature/Feature.php
```diff
@@ -353,7 +365,21 @@ public static function getDropdownItems($withHidden = 0) {
 
             return $features_by_category;
         } else {
-            return self::where('is_visible', '>=', $visibleOnly)->orderBy('name')->pluck('name', 'id')->toArray();
+            if (config('show_species-only_traits_in_dropdown') && $withSpecies) {
```
ScuffedNewt: should `species-only` be underscored?

config/lorekeeper/extensions.php
```diff
@@ -140,4 +140,7 @@
 
     // Unmerge Item Page and Item Entry - Speedy
     'unmerge_item_page_and_entry' => 0, // If enabled, uses the html on world/item_page.blade.php instead of the include that links to world/_item_entry.blade.php
+
+    // Show Species-only traits in dropdown - Speedy
+    'show_species-only_traits_in_dropdown' => 0, // If enabled, will only show traits from the associated species as well as traits that aren't species-limited in the dropdown menus.
```
ScuffedNewt: nit: instead of mixing dash and "_" maybe name something like `show_only_species_traits_in_dropdown`

ScuffedNewt (approved):
Only nits on my end, LGTM and huge QoL!

itinerare:
Yeah, I don't have the wherewithal to give this a meaningful look atm, but it's *really* nice QoL.

app/Http/Controllers/Characters/DesignController.php
SpeedyD: Thing is, this refreshes the _new trait_ selection in the features modal. Traits are called 'features' internally. So.. it's.. Features Feature. 😅

Understandable nit, but at the same time _THIS_ is actually consistency..

SpeedyD:
(Note, final commit was automated by Github, all I did was click 'update branch'. It contains no changes and should not count for review.)

app/Models/Feature/Feature.php
perappu: +1 to this should be underscored rather than a dash

app/Http/Controllers/Characters/DesignController.php
perappu: I agree it should be something a little more clear, because admittedly it wasn't 100% clear to me what was going on when I initially read it. maybe `getFeaturesBySpeciesFeature` if you really want to keep it?

SpeedyD: `getFeaturesTrait` it is

app/Models/Feature/Feature.php
SpeedyD: And `show_exlusively_species_traits_in_dropdown` it is

config/lorekeeper/extensions.php
SpeedyD: nitfix: `show_exlusively_species_traits_in_dropdown`

ScuffedNewt (approved):

LostInProgres (approved):

app/Http/Controllers/Characters/DesignController.php
```diff
@@ -235,7 +235,7 @@ public function getFeatures($id) {
             'specieses' => ['0' => 'Select Species'] + Species::visible()->orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
             'subtypes'  => Subtype::visible()->where('species_id', '=', $r->species_id)->orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
             'rarities'  => ['0' => 'Select Rarity'] + Rarity::orderBy('sort', 'DESC')->pluck('name', 'id')->toArray(),
-            'features'  => Feature::getDropdownItems(),
+            'features'  => Feature::getDropdownItems(Auth::user()->hasPower('edit_data'), $r->species_id),
```
itinerare: For what it's worth... I still think it makes more sense to hide hidden traits in design updates *regardless* of staff powers, because that isn't an area that you're interacting with *as staff*.

SpeedyD: Gonna be honest, this kinda feels like a nitpick.. But alright.

itinerare: imo, not a nit. It was the way it was for a reason.

itinerare (approved):

PR #1371, opened 2025-11-27
Fix(blade-formatter): Hopefully?

SpeedyD:
Honestly sick of this file as blade formatter keeps editing it, so I decided to do this workaround.


itinerare:
It'd probably be better just to add it to the ignore list, at this point...

SpeedyD:
> It'd probably be better just to add it to the ignore list, at this point...

Disagree, rest of the file still has worth being formatted..

Draconizations (approved):
I agree with Speedy, the rest of the file is worth formatting

SpeedyD:
ALTERNATE SOLUTION I FORGOT ABOUT:
`{{-- blade-formatter-disable --}}` and `{{-- blade-formatter-enable --}}`

100% works, which is proven by the fact that the above has no blade formatting commit. :)

itinerare (approved):
Yeah, that's a much better/more precise solution.

Draconizations (approved):
Yeah this is probably the cleaner solution LOL

PR #1372, opened 2025-11-29
fix: dynamic rewards rewardable_id and item_id error

perappu:
- Certain very popular extensions and other custom coded features have asset types that do not actually exist as a model (things such as EXP or points), and therefore do not have a rewardable_id. This migration changes the rewardable_id column to be nullable. Otherwise, it will cause the rewards to silently fail to save, because it can't create a null value in that column. I figured it was better to handle this in core than rely on various slightly different extensions or custom code solutions.
- The item_id column does not exist and throws a SQL error if you tried to refer to it, so I changed it to object_id instead (this should serve the same purpose of creating a "null" relationship).

Edit:
- Also now actually shows errors from RewardService, instead of just silently eating them.

### DIFF
--- a/app/Models/Reward/Reward.php
+++ b/app/Models/Reward/Reward.php
@@ -30,6 +30,36 @@ class Reward extends Model {
         'data' => 'array',
     ];
 
+    /**
+     * Validation rules for reward creation.
+     *
+     * @var array
+     */
+    public static $createRules = [
+        'object_id'            => 'required',
+        'object_model'         => 'required',
+        'rewardable_recipient' => 'required',
+        'rewardable_id'        => 'nullable',
+        'rewardable_type'      => 'required',
+        'quantity'             => 'required',
+        'data'                 => 'nullable',
+    ];
+
+    /**
+     * Validation rules for reward updating.
+     *
+     * @var array
+     */
+    public static $updateRules = [
+        'object_id'            => 'required',
+        'object_model'         => 'required',
+        'rewardable_recipient' => 'required',
+        'rewardable_id'        => 'nullable',
+        'rewardable_type'      => 'required',
+        'quantity'             => 'required',
+        'data'                 => 'nullable',
+    ];
+
     /**********************************************************************************************
 
         RELATIONS
--- a/app/Models/Reward/Reward.php
+++ b/app/Models/Reward/Reward.php
@@ -51,7 +51,7 @@ class Reward extends Model {
 
         if (!class_exists($model)) {
             // Laravel requires a relationship instance to be returned (cannot return null), so returning one that doesn't exist here.
-            return $this->belongsTo(self::class, 'id', 'item_id')->whereNull('item_id');
+            return $this->belongsTo(self::class, 'id', 'object_id')->whereNull('object_id');
         }
 
         return $this->belongsTo($model, 'rewardable_id');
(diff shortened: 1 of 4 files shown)


SpeedyD (approved):

SpeedyD:
Honestly, approved if only for the second edit- I think that's a bug to begin with, as 'item_id' is referenced NOWHERE else but ni that line.

ScuffedNewt (changes requested):
Would prefer this have some level of validation against null `object_id` for things like Items etc, it's unlikely to happen but I've seen people manage to do worse

Something as simple as a helper that returns if the `object_id` is allowed to be null or not on the reward model or the `AssetsHelper`

SpeedyD:
> Would prefer this have some level of validation against null `object_id` for things like Items etc, it's unlikely to happen but I've seen people manage to do worse
> 
> Something as simple as a helper that returns if the `object_id` is allowed to be null or not on the reward model or the `AssetsHelper`

I am of strong believe that in Lorekeeper development, it is never a bad idea to PR a fix or give code suggestion :)

If you know how to fix this in a more efficient way, I doubt @perappu would mind you PRing it for their `perappu:pr/dynamic-rewards-fixes` branch :)

perappu:
> Would prefer this have some level of validation against null `object_id` for things like Items etc, it's unlikely to happen but I've seen people manage to do worse
> 
> Something as simple as a helper that returns if the `object_id` is allowed to be null or not on the reward model or the `AssetsHelper`

I don't think object_id is nullable? So it shouldn't be at all possible for it to turn out null. It'll be "0" at worst, which to my knowledge, when it comes to SQL comparisons is a non-null value (vs PHP ones). But we could certainly add validation rules on Reward which would prevent that from happening at all!

perappu:
So while looking into adding validation rules I also discovered that any and all errors from RewardService are just getting silently eaten, so I also made it that errors from RewardService will actually flash at the user instead of just silently failing.

Should be ready to go now!

ScuffedNewt:
Another brain worm I had, but if this is to allow null ids it should probably update the frontend to explicitly handle that (ex. Hide the ID selection when a non-Id object is selected)
Otherwise this wouldn't handle full support and would still be up for extensions to handle

perappu:
> Another brain worm I had, but if this is to allow null ids it should probably update the frontend to explicitly handle that (ex. Hide the ID selection when a non-Id object is selected) Otherwise this wouldn't handle full support and would still be up for extensions to handle

I think it's really hard to identify a "non-ID" object in this case. Most things are going to have handling for the ID selection already -- your own C&C with Exp and Points is a great example.

I think any potentially complicated AJAX required to make the frontend understand when a model should and shouldn't have an ID associated is something that should be on the extension creator when adding the rows in loot_select blades, loot js, etc. The most we can reasonably do is make the field nullable IMHO

Full transparency, given there's now some pretty huge bugfixes in this PR, I'm also interested in just getting it out the door.

SpeedyD:
Isn't the only thing extensions need to do is add to AssetHelpers and the two Loot_selects? I think we should just approve this and get it going, we can finetune as we get closer to _releasing_ this.

SpeedyD (approved):

ScuffedNewt (approved):

PR #1376, opened 2025-12-04
fix universal/all traits index

KingAKamina:
Fix to make Universal Traits Index and All Traits Index show


LostInProgres (approved):

SpeedyD (commented):
...I'm unsure why this is thought necessary?

KingAKamina:
> ...I'm unsure why this is thought necessary?

Cus the editted indexes don't show at all

SpeedyD:
Oh, right, I see now. PR #1366 seems to have broken this.. Odd.

I'm approving of this change, but I'm thinking of looking into maybe reducing reused code even further..

SpeedyD (approved):

PR #1379, opened 2025-12-16
feat(loot): overhaul loot blades and consolidate duplicated code

perappu:
This is an approach for modernizing the code found in the loot related blades.

- The long inline php code has been moved into AssetHelpers, and consolidated as much as possible.
- `getRewardTypes()` returns the data in the `rewardable_type[]` fields, based on an array of "showXYZ" data and an `$isCharacter` value.
- `getRewardData()` fetches the data that will be used to populate `rewardable_id[]` select fields, and replaces the code in _loot_select and _loot_select_row.
- `rewardable_id[]` fields will now be programmatically generated based on the results of getRewardTypes as well as valid recipient types, reducing the need for many select fields in _loot_select and _loot_select_row.
- Leveraging Dynamic Rewards, we are also able to finally cleanly differentiate between "character" reward types and "user" reward types. (No more characters being able to be given raffle tickets and recipes!)
  - When you switch between recipient types, it will now make an AJAX request to automatically change what reward types are available for selection.
  - This has been tested and found compatible with trade listings.
  - When a character recipient type is selected, only character valid currency and items will be shown.

Please review the code comments as well, as I tried to make things as clear as possible for people who may be adding their own asset types or porting those from extensions.

Thank you!


resources/views/js/_loot_js.blade.php
```diff
@@ -53,17 +60,45 @@
...
+                type: "POST",
+                headers: {
+                    'X-CSRF-TOKEN': '{{ csrf_token() }}'
+                },
+                url: "{{ url('rewards/types') }}",
+                data: {
+                    recipient: $recipient,
+                    prefix: '{{ $prefix }}',
+                    type: '{{ $type }}',
+                    showData: JSON.parse('{!! json_encode($showData) !!}'),
```
ScuffedNewt, line 66: only a nit / for knowledge, I think you can use `@json` for this (don't make a new revision just for this, though)

ScuffedNewt (approved):
for once in my LK existence I humbly ask this be merged ASAP, this will be so helpful 😭

resources/views/js/_loot_js.blade.php
perappu: good to know! tbh I just used the method I knew had worked from my past finagling for getting json to go through lol

SpeedyD (approved):

PR #1385, opened 2025-12-20
fix(users): fix hardcoded member rank id in create user fortify action

perappu:
Discovered this when my galera cluster created the member rank with a default ID of 5! yay!

Just pulls in the rank selection logic from the v2/non-Fortify user creation in UserService and applies it to Fortify.


itinerare (approved):

SpeedyD (approved):
...buh.. why'd this change.. O_o;

PR #1390, opened 2025-12-29
feat: user and character reward grants

ScuffedNewt:
image
image
image
image

### DIFF
--- a/app/Services/RewardManager.php
+++ b/app/Services/RewardManager.php
@@ -1,0 +1,143 @@
+<?php
+
+namespace App\Services;
+
+use App\Facades\Notifications;
+use App\Models\Character\Character;
+use App\Models\Currency\Currency;
+use App\Models\Item\Item;
+use App\Models\Loot\LootTable;
+use App\Models\Raffle\Raffle;
+use App\Models\User\User;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\DB;
+
+class RewardManager extends Service {
+    /*
+    |--------------------------------------------------------------------------
+    | Reward Service
+    |--------------------------------------------------------------------------
+    |
+    | Handles the grants of rewards on objects.
+    |
+    */
+
+    /**********************************************************************************************
+
+        REWARD GRANTS
+
+    **********************************************************************************************/
+
+    /**
+     * Grants rewards to users or characters.
+     *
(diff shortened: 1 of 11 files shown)


SpeedyD:
I'm.. a little bit confused. Isn't this just ... item and currency grants but in a single page? 🤔

..Ah, and loot table grants, since that's also a recent addition..

SpeedyD (approved):
I've been told that this is, in fact, a condensed rewarder, potentially replacing Currency Grant, Item Grant, Loot Table Grant.. (Award Grant? for those with the ext anyway)

It seems like a decent feature to me, and so I'm approving it. We can eventually remove the other grant features, I reckon.

itinerare:
Nah, we had a lengthy discussion to the effect of "that would cause problems elsewhere".

SpeedyD:
> Nah, we had a lengthy discussion to the effect of "that would cause problems elsewhere".

_Would_ it? I can't really imagine.. 🤔 Well, I also haven't looked into it that closely, so I guess I'll take your word for it..

app/Services/RewardManager.php
```diff
@@ -0,0 +1,143 @@
...
+                if ($q <= 0) {
+                    throw new \Exception('All quantities must be at least 1.');
+                }
+            }
+
+            // add assets first
+            if (isset($data['rewardable_type'])) {
+                foreach ($data['rewardable_type'] as $key => $assetType) {
+                    $reward = null;
+                    switch ($assetType) {
```
perappu: I think you could technically simplify this entire switch statement to something like like `$reward = getAssetModelString(strtolower($data['rewardable_type'][$key]))::find($data['rewardable_id'][$key]);`

kinda like we do in Reward.php: https://github.com/lk-arpg/lorekeeper/blob/7744b30d0758ba721933733ea4f24e7519dd8e2a/app/Models/Reward/Reward.php#L80

Doing it that way means people can automatically grant whatever goofy extension stuff they have from this page, too. (since it'll automatically show up in the _loot_select, but as it stands will just explode when it hits this function)

perappu (changes requested):

app/Services/RewardManager.php
ScuffedNewt: True true

perappu (approved):

PR #1394, opened 2026-01-02
fix(trait-indexes): 500 error due to 'subtype' missing

SpeedyD:
Fixes a 500 error on opening modals from any of the trait indexes.

I'm not entirely sure why eager loading is required here when all it does is indirectly load a trait page? Maybe I'm missing something. Would love an explanation and would not mind fixing if required.

### DIFF
--- a/app/Http/Controllers/WorldController.php
+++ b/app/Http/Controllers/WorldController.php
@@ -477,7 +477,7 @@ class WorldController extends Controller {
      * @return \Illuminate\Contracts\Support\Renderable
      */
     public function getFeatureDetail($id) {
-        $feature = Feature::visible(Auth::user() ?? null)->where('id', $id)->with('species', 'subtype', 'rarity')->first();
+        $feature = Feature::visible(Auth::user() ?? null)->where('id', $id)->first();
 
         if (!$feature) {
             abort(404);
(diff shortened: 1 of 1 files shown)


itinerare:
It's not terribly necessary, but it's still going to be marginally more efficient-- the point of eager loading is, as always, taking fewer trips to the database. That's going to be true whether you have twenty objects on the page or just one.

SpeedyD:
> It's not terribly necessary, but it's still going to be marginally more efficient-- the point of eager loading is, as always, taking fewer trips to the database. That's going to be true whether you have twenty objects on the page or just one.

So it might be better to _just_ remove subtype? Or perhaps re-add the use in the Feature model.. If that'd fix it..?

itinerare:
Well, there's still a relation in there... but yeah, I'd say that if the issue is that that relation isn't working for eager loading somehow, it would be worth investigating why that is first and foremost.

This is in part because ultimately, the goal will be to enable automatic eager loading (once Laravel 12 is set up, so this is a bit down the line)-- which will remove the need to manually specify eager loading, but *will* demand that all the eager loading *works*. So merely removing it for now is just kicking the can down the road.

SpeedyD:
...That means that once Laravel 12 hits, we might find ourselves with a LOT of cans to pick up. Got it.

itinerare:
Yeeeeeeep. You're getting the picture.

SpeedyD:
Hrm, might actually not be a problem here. I realized it's RELATIONSHIP loading, so it's less 'kicking the can down' and more 'kicking the can towards a recycling center'... you get my drift.

..which also brings me to the following solution.

itinerare (approved):
There you go!

Draconizations (approved):
can't believe I missed this. either way, let's get this merged

PR #1399, opened 2026-01-11
Attachments

ScuffedNewt:
Allows a drop in widget, similar to rewards and limits, to be added to link objects together.
The links can have custom values and displays (created by the site owner) to allow for easy linking of prompts or other drop-in type displays

attachments can be display only or used as references in updates / other logic


SpeedyD:
...I feel like I'm missing something, what precisely does this add??

ScuffedNewt:
> ...I feel like I'm missing something, what precisely does this add??

I realised after I forgot some features, namely being able to reference attachments through parsed text a-la dialogues, but the idea is that attachments are a catch all for referencing and creating relationships between objects, so that extensions or otherwise have a common ground to use and can apply custom logic to, if needed

it's basically the generic of rewards or limits

PR #1404, opened 2026-01-18
fix(character): Minor button renaming

SpeedyD:
This is admittedly a nitpick, but I've changed this on some sites ages ago. Having a bunch of buttons all just be 'edit' when they're strictly for specific purposes seems a bit odd.

I did keep 'Edit' for the visibility update button, because that one directly updates. It does not open a modal, and as such it's actually more consistent with other administrative Edit buttons.

### DIFF
--- a/resources/views/character/_image_info.blade.php
+++ b/resources/views/character/_image_info.blade.php
@@ -120,7 +120,7 @@
 
                 @if (Auth::check() && Auth::user()->hasPower('manage_characters'))
                     <div class="mt-3">
-                        <a href="#" class="btn btn-outline-info btn-sm edit-features" data-id="{{ $image->id }}"><i class="fas fa-cog"></i> Edit Features</a>
+                        <a href="#" class="btn btn-outline-info btn-sm edit-features" data-id="{{ $image->id }}"><i class="fas fa-cog"></i> Edit Traits</a>
                     </div>
                 @endif
             </div>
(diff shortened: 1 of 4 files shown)


SpeedyD:
..don't mind that. I figured I'd fix the typo in the branch name but apparently instead of updating it on a PR, Github just assumes it is deleted and closes the PR. Whoops.

resources/views/character/_image_info.blade.php
```diff
@@ -120,7 +120,7 @@
 
                 @if (Auth::check() && Auth::user()->hasPower('manage_characters'))
                     <div class="mt-3">
-                        <a href="#" class="btn btn-outline-info btn-sm edit-features" data-id="{{ $image->id }}"><i class="fas fa-cog"></i> Edit</a>
+                        <a href="#" class="btn btn-outline-info btn-sm edit-features" data-id="{{ $image->id }}"><i class="fas fa-cog"></i> Edit Features</a>
```
itinerare: The label on this should probably be traits...

SpeedyD: The modal include species, subtype, rarity.. and in some ext cases, even more. So I figured 'features' in this case can mean actual features of the design. :)

itinerare: Yes, but *features* in this case refers to *traits*-- which will be more self-explanatory as a label than something vague but all-encompassing.

SpeedyD: ...I suppose that in the end 'traits' is technically also all-encompassing.

itinerare (approved):

MarskyMessier (approved):
I think that's a fine PR to add better accessibility to what people edit front end wise!

Issue #1393, opened 2025-12-31
Adding an email redirects to /home, causing a 404


ScuffedNewt:
After adding an email to an alias created account, url redirect is /home but should be /

ScuffedNewt:
Closed by https://github.com/lk-arpg/lorekeeper/pull/1409

PR #1409, opened 2026-01-21
Allow unverified users to edit their emails

ScuffedNewt:
- ~Allow non admin ranks to do admin things~
  - ~nice if you don't want to have to give everyone admin to do admin things, especially since you need to remove the role in the DB~
Moved to https://github.com/lk-arpg/lorekeeper/pull/1462
---

- Allow unverified users to edit their emails 
image
image

### DIFF
--- a/routes/web.php
+++ b/routes/web.php
@@ -27,6 +27,12 @@ require_once __DIR__.'/lorekeeper/browse.php';
 
 Route::feeds('feeds');
 
+// Updating email from verification notice
+Route::middleware('auth')->group(function () {
+    Route::get('/email/update', 'HomeController@getEmail');
+    Route::post('/email/update', 'HomeController@postEmail');
+});
+
 /**************************************************************************************************
     Routes that require login
 **************************************************************************************************/
--- a/app/Http/Controllers/HomeController.php
+++ b/app/Http/Controllers/HomeController.php
@@ -48,7 +48,7 @@ class HomeController extends Controller {
     public function getLink(Request $request) {
         // If the user already has an alias associated with their account, redirect them
         if (Auth::check() && Auth::user()->hasAlias) {
-            redirect()->to('home');
+            return redirect()->route('home');
         }
 
         // Display the login link
@@ -111,7 +111,11 @@ class HomeController extends Controller {
     public function getEmail(Request $request) {
         // If the user already has an email associated with their account, redirect them
         if (Auth::check() && Auth::user()->hasEmail) {
-            return redirect()->to('home');
+            return redirect()->route('home');
+        }
+
+        if (config('lorekeeper.settings.allow_unverified_users_to_modify_emails') && $request->is('email/update')) {
+            return view('auth.update_email');
         }
 
         // Step 1: display a login email
@@ -128,7 +132,7 @@ class HomeController extends Controller {
         if ($service->updateEmail(['email' => $data], Auth::user())) {
             flash('Email added successfully!');
 
-            return redirect()->to('home');
+            return redirect()->route('home');
         } else {
             foreach ($service->errors()->getMessages()['error'] as $error) {
                 flash($error)->error();
--- a/config/lorekeeper/settings.php
+++ b/config/lorekeeper/settings.php
@@ -429,4 +429,14 @@ return [
     |
     */
     'enable_character_content_warnings'                 => 1,
+
+    /*
+    |--------------------------------------------------------------------------
+    | Allow Unverified Users to Modify their Email Addresses
+    |--------------------------------------------------------------------------
+    |
+    | Allows users who have not yet verified their email addresses
+    | to modify their email address, in order to correct mistakes.
+    */
+    'allow_unverified_users_to_modify_emails'           => 1,
 ];
--- a/resources/views/auth/update_email.blade.php
+++ b/resources/views/auth/update_email.blade.php
@@ -1,0 +1,41 @@
+@extends('layouts.app')
+
+@section('title')
+    Update Email Address
+@endsection
+
+@section('content')
+    <h1>Update Email Address</h1>
+    <p class="mb-1">
+        The currently linked email to your account is:
+    </p>
+    <div class="alert alert-secondary mb-2" role="alert">
+        {{ Auth::user()->email }}
+    </div>
+    <p>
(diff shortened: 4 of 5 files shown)


itinerare:
It'd be better if these were separate PRs, though tbh I don't think allowing admins to edit user emails is the way to resolve e.g. "user typo'd their email address and isn't getting the verification email" to begin with/I'm disinclined to accept that change at all. One idea that came up in the server was just having a separate view where unverified users can check/change it, which would probably be preferable...

SpeedyD:
I can see the benefit of giving non-admins admin rights; means that some of the sites I'm on can just give me my own role, haha..

But I'm not sure about the email editing either..

ScuffedNewt:
> I'm disinclined to accept that change at all.

any particular reason why?


> having a separate view where unverified users can check/change it, which would probably be preferable...

I can raise another CR with this change, would like to hear the reasoning for the above though, I don't think having both is bad necessarily

itinerare:
It's essentially still a security thing that's currently treated as an inconvenience because sometimes it *is* a genuine mistake... but e.g. I don't particularly enjoy opening the door to someone social engineering their way into getting an admin to change an account's email address. And while, yes, it will always be possible to edit directly via the DB, there's a difference between it being possible and being convenient.

perappu:
I'm in agreeance with Mercury here and was just about to come in and say the same thing. It's absolutely a vector for social engineering -- I've heard of people doing similar things with Toyhouse or Discord accounts, as well as at my job. I think the suggestion of having a separate view where unverified users can check/change it would be the right call to solving the "I typo'd my email and I'm not getting verification emails" problem. Maybe make it visible on the verify email page that people get redirected to?

ScuffedNewt:
Happy to make the change for it to be user-initiated then! Hadn't thought about the social engineering aspect of it, I guess I have too much trust in the thought process of people with admin privileges 😭

SpeedyD (approved):
I like this fix- as long as a user has no verified alias, they can fix the email address themselves. Means they have to be logged in already- and that occurs right after registration.

Good call! :)

routes/web.php
```diff
@@ -27,6 +27,13 @@
 
 Route::feeds('feeds');
 
+// Updating email from verification notice
+Route::middleware('auth')->group(function () {
+    Route::get('/email/update', 'HomeController@getEmail');
+
```
SpeedyD: This is 100% a nitpick- wondering why there's an empty line in between is all :)

SpeedyD:
Oh yeah, does this also fix #1393 ?

LostInProgres (approved):

app/Models/Rank/Rank.php
```diff
@@ -73,7 +73,7 @@ public function getDisplayNameAttribute() {
      * @return bool
      */
     public function getIsAdminAttribute() {
-        return $this->attributes['is_admin'];
+        return $this->attributes['is_admin'] || $this->powers()->where('power', 'admin')->exists();
```
itinerare: See, I understand why you've done this. But you've put back the query that's going to be *everywhere*...

itinerare:
Once again, I would like to voice that these should really be separate PRs.

app/Models/Rank/Rank.php
ScuffedNewt: do you prefer another attribute?

SpeedyD (changes requested):
I believe I'll swap my approval into an RFC. Merc isn't going to drop the 'split into two PRs' point, and after thinking it through, I agree with it.

I strongly suggest moving the 'user can change own email address if unverified' one to a new PR, since the most recent comments are about the faux admin ability. ;)

app/Models/Rank/Rank.php
itinerare: Unsure what the ideal solution is (brainfog's been bad lately), beyond wanting to avoid a query if possible for performance reasons.

SpeedyD: Might be a bit redundant and roundabout, but why not add a column 'fauxAdmin' with a bool.. that way you only need to pull that attribute..?

Does mean you'd have to slightly rewrite the rank controller, so that it actually gets set or unset depending on whether the power is there..

perappu: I was actually thinking on this morning and I agree that's probably the best approach, even if it's technically data duplication. Alternatively, would always eager loading powers alongside Rank solve the issue? aka setting the $with on the Rank model. To some degree I'm surprised it doesn't do that already

itinerare: Possibly? re the latter

ScuffedNewt:
Resolves https://github.com/lk-arpg/lorekeeper/issues/1393

SpeedyD:
I'm guessing this is still in progress, since the secondary admin code is still in here? Perhaps set this to draft until it's fixed. :)

app/Models/Rank/Rank.php
```diff
@@ -95,6 +95,15 @@ public function canEditRank($rank) {
         }
         if ($this->hasPower('edit_ranks')) {
             if ($this->isAdmin) {
+                // editing a false admin rank
```
perappu: another old section

app/Models/Rank/Rank.php
perappu: Looks like this is a section that got moved to the other PR and should be removed

config/lorekeeper/powers.php
```diff
@@ -10,7 +10,10 @@
     | The list of staff powers that can be attached to a user rank.
     |
     */
-
+    'admin' => [
```
perappu: another section that should be removed

resources/views/admin/users/_create_edit_rank.blade.php
perappu, line 1: all the changes in this file need to be discarded

perappu (changes requested):
This is more or less just directly poking at the things that need to get removed to draw attention to it, as the PR otherwise looks fine and I'm happy with the changes made.

SpeedyD:
Yeah, you basically pointed out the exact things I found issue with, @perappu.

perappu (approved):
ty for making the necessary updates o>

itinerare (approved):

PR #1411, opened 2026-01-24
feat(deps): bootstrap v5 update and other relevant dependancies

liwoyadan:
I wrote out a draft PR much longer in May of last year but lost it when my browser crashed (tear rolls down my face) so I'll just leave a shorter draft PR here for now so I don't forget again.

Already did a pass with the gulp migration tool as well as manual edits for any straggling classes left behind. Bootstrap v5 also written with a new `_variables.scss` to style it to look like how default Lorekeeper did with Bootstrap v4 (as to not totally change the look of sites that update) but of course site owners and etc are free to modify.

Other things I still need to do:
- Update jQuery in a number of places to replace deprecated functions
- There's something odd about the bootstrap5-toggle package specifically with toggles that are contained in tabs like the ones on character pages that staff can see for toggling visibility, etc. I noticed that when flipping between tabs, the toggle would no longer render, and so I tried to do a workaround (you can see the listener in the scripts section of `resources/views/layouts/app.blade.php`) but although that *did* re-render the toggles, ones that were toggled on would re-render as off and vice versa...it's been some months now so I'll go squint at it again soon


liwoyadan:
Hate to impose, but someone will...likely have to pick this one up for me. |D... (I believe I told Toko about it but if divvying up the workload that remains works out better, do feel free!! Remember to take breaks!!)

itinerare:
No worries! It's definitely a big task regardless....

PR #1418, opened 2026-01-26
QoL(LootTables): Added sorting and ability to search for DisplayName

SpeedyD:
Admittedly this one is a bit rougher. Please look over this one with discerning eyes.


SpeedyD:
Aight, workflows went cleanly now. Sorry 'bout that.

ScuffedNewt (approved):

perappu (changes requested):
trying to get github to let me just actually request the changes, please hold

perappu (changes requested):
I think it's working this time... sorry if this sent you a billion notifications.

edit: nevermind, no it didn't. I'll try this again later today

app/Http/Controllers/Admin/Data/LootTableController.php
```diff
@@ -29,8 +29,40 @@ class LootTableController extends Controller {
     public function getIndex(Request $request) {
         $query = LootTable::query();
 
-        if ($request->get('name')) {
-            $query->where('name', 'LIKE', '%'.$request->get('name').'%');
+        $data = $request->only(['name', 'namespec', 'sort']);
+        if (isset($data['name'])) {
+            if (isset($data['namespec']) && $data['namespec'] == 'display-name') {
+                $query->where('display_name', 'LIKE', '%'.$request->get('name').'%');
```
perappu:
```suggestion
                $query->where('display_name', 'LIKE', '%'.$data['name'].'%');
```

teeny tiny optimization because you already have the value from $data

app/Http/Controllers/Admin/Data/LootTableController.php
```diff
@@ -29,8 +29,40 @@ class LootTableController extends Controller {
...
+            } elseif (isset($data['namespec']) && $data['namespec'] == 'both') {
+                $query->where('name', 'LIKE', '%'.$request->get('name').'%')->orWhere('display_name', 'LIKE', '%'.$request->get('name').'%');
```
perappu:
```suggestion
                $query->where(function ($q) use ($data) {
                    $q->where('name', 'LIKE', '%'.$data['name'].'%')
                    ->orWhere('display_name', 'LIKE', '%'.$data['name'].'%');
                });
```

So, I've observed this on a lot of log queries -- stacking "orWheres" and "wheres" can result in a lot of strange behaviors, because Laravel doesn't know by default how to group them intelligently at all. Example: https://github.com/lk-arpg/lorekeeper/pull/1440

While not presently a problem because this is the only where() called, this will cause a bug if additional where() conditions are added onto the loot table search. I think it's a nice pro-active measure to stick it in its own group, but YMMV of course

also same optimization as above

app/Http/Controllers/Admin/Data/LootTableController.php
```diff
@@ -29,8 +29,40 @@ class LootTableController extends Controller {
...
+            } else {
+                $query->where('name', 'LIKE', '%'.$request->get('name').'%');
```
perappu:
```suggestion
                $query->where('name', 'LIKE', '%'.$data['name'].'%');
```

and once more, with feeling!

perappu (changes requested):

perappu (approved):
thank you for looking at those changes!

SpeedyD:
> thank you for looking at those changes!

No, thank YOU for suggesting them! :)

PR #1425, opened 2026-01-30
fix(style): remove thead-light to uniformize tables

MarskyMessier:
This is mostly an aesthetic fix for tables as there were some tables with thead-light when it's not the case for most tables anymore. I believe develop has some additional ones with trade listing added. 
My reasoning is that it doesn't change that much for the visual and helps being more consistent accross theme as *-light can be inconsistent in darker themes + people will generally style the head with a color. This is basically just to uniformize tables accross the site.

I do know some might like the darker shade *-light gives for light mode bootstrap so feel free to give other suggestions or more cross-theme friendly options!

### DIFF
--- a/resources/views/admin/submissions/submission.blade.php
+++ b/resources/views/admin/submissions/submission.blade.php
@@ -101,7 +101,7 @@
             <h2>Add-Ons</h2>
             <p>These items have been removed from the {{ $submission->prompt_id ? 'submitter' : 'claimant' }}'s inventory and will be refunded if the request is rejected or consumed if it is approved.</p>
             <table class="table table-sm">
-                <thead class="thead-light">
+                <thead>
                     <tr class="d-flex">
                         <th class="col-2">Item</th>
                         <th class="col-4">Source</th>
--- a/resources/views/character/design/addons.blade.php
+++ b/resources/views/character/design/addons.blade.php
@@ -33,7 +33,7 @@
             <div class="card mb-3">
                 <div class="card-body">
                     <table class="table table-sm">
-                        <thead class="thead-light">
+                        <thead>
                             <tr class="d-flex">
                                 <th class="col-2">Item</th>
                                 <th class="col-4">Source</th>
--- a/resources/views/home/_submission_content.blade.php
+++ b/resources/views/home/_submission_content.blade.php
@@ -74,7 +74,7 @@
         <div class="card-header h2">Rewards</div>
         <div class="card-body">
             <table class="table table-sm">
-                <thead class="thead-light">
+                <thead>
                     <tr>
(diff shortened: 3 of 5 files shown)


SpeedyD:
Perhaps it is just me, but I feel like having a stronger head color makes it stand out from the rest of the table..

I'd suggest doing the opposite- instead of REMOVING thead-light everywhere, I'd suggest ADDING it everywhere.

MarskyMessier:
Then could I suggest thead-dark instead? *-light properties have been a bit of a troublemaker in UI editing as in most dark theme the light property can be hell to deal with, it also makes the table head stand out more than light. It's the same shade as the navbar I believe.

SpeedyD:
Dark.. Yeah, that could work. :)

ScuffedNewt:
Small request, can screenshots be included?

MarskyMessier:
Okay so I'm revisiting this PR and I'm now curious on one thing : is there a reason why LK has both table element and custom table styling ? Wouldn't it be smart to then convert all table elements into the one LK uses for uniformity?
This of course would be done on a new PR as it is out of scope of this one, technically.

itinerare:
So as far as I recall, the situation is:
- HTML tables aren't very responsive with regards to mobile
- Some cases we have tables for don't play nice with the bootstrap pseudo-tables for... reasons I don't remember? I recall it being a deliberate thing, though.

Now granted, this is a few years distant at this point, so my recollection is certainly not infallible.

SpeedyD:
Now that you mention that, I also remember that a WHIIIIILE back, almost all tables got REPLACED with Bootstrap Pseudo table. So I'm kind of thinking maybe we should replace all of them.

perappu:
It doesn't apply to the ones in this PR, but I believe sortable has issues with the nesting that happens with the bootstrap tables. I said it on discord but I'm pretty sure these tables aren't bootstrap ones due to the row sizing behavior being weird on mobile/not wanting the table to wrap within the already cramped space with how some of these divs are nested. I'd be happy to be wrong though.

Either way, I wouuuuuld argue that a major refactor of various tables across the site is something that should live in a develop PR, rather than this v3 one.

itinerare:
Yeah, it being related to the manual sort setup sounds about right/matches my vague recollection...

SpeedyD:
I have none of this in my memory, but I'll take your words for it. Either way: Yes, not in this PR. Sorry.

Whiiiich brings us back to: What is actually gonna happen with this PR? The last suggestion was to make it thead-dark rather than removing or adding thead-light everywhere, as well as a request for screenshots..

perappu:
Yeah, @MarskyMessier if you could provide screenshots as requested and make the change to thead-dark then I think this is fine.

PR #1430, opened 2026-02-15
fix: allow utilisation of $type

ScuffedNewt:
this is already used in the child blades but was never actually being passed down correctly


ScuffedNewt:
need to fix a few things on this looking at it in practice but the idea is there

SpeedyD:
Gonna need some context here.. Also, perhaps change this to draft if still unfinished? ^^;

ScuffedNewt:
> Also, perhaps change this to draft if still unfinished? ^^;

yes, good point, will do!

ScuffedNewt:
> need to fix a few things on this looking at it in practice but the idea is there

i dont remember what this comment was referring to and as far as i can see this just works...

perappu (approved):

SpeedyD (approved):

PR #1432, opened 2026-02-22
fix: character default assets use character asset keys

ScuffedNewt:


ScuffedNewt:
This also requires `mergeAssetsArrays` to be updated, will cut another commit in a min

perappu (approved):

perappu (approved):

SpeedyD (approved):

PR #1433, opened 2026-02-25
Raffle Options

ScuffedNewt:
adds a lot of functionality to raffles including:
- FTO / Non-Owner raffles
- Entry rewards
- Winner rewards
  - Optional positional rewards
- Ability to reward characters (also adds a general `$showCharacter` option to rewards)
- User self-entry (pretty common demand / sought-after feature)
- Unordered results in the event of less than winner # entrants
image
image
image


SpeedyD (approved):

SpeedyD:
@ScuffedNewt: There are conflicts, could you fix them? :)

ScuffedNewt:
I will resolve them when approvals are given since it makes more sense to do all necessary changes in one

app/Console/Commands/ConvertRewards.php
perappu, line 1: Just leaving a comment for @itinerare 's attention that, once this gets merged, we'll want people to re-run `php artisan convert-rewards`

app/Services/RaffleService.php
```diff
@@ -121,11 +155,128 @@ public function updateRaffleGroup($data, $group) {
...
+        if (!$rewardService->populateRewards(
+            get_class($raffle),
+            $raffle->id,
+            $rewardData,
+            true
+        )) {
+            foreach ($rewardService->errors()->getMessages()['error'] as $error) {
+                flash($error)->error();
+            }
+            throw new \Exception('Failed to create rewdddards.');
```
perappu, line 278:
```suggestion
            throw new \Exception('Failed to create rewards.');
```

this made me laugh LOL but it should _probably_ be spelled correctly.......

database/migrations/2026_02_24_223014_add_description_to_raffles.php
```diff
@@ -0,0 +1,29 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class extends Migration {
+    /**
+     * Run the migrations.
+     */
+    public function up(): void {
+        Schema::table('raffles', function (Blueprint $table) {
+            //
+            $table->text('description')->nullable()->default(null);
```
perappu, line 14:
```suggestion
            if (!Schema::hasColumn('raffles', 'description'))
            {
                $table->text('description')->nullable()->default(null);
            }
```

Adding a description field is a suuuuuper common custom code change (I've been asked for it a few times) so I think it'd just be a nice QoL to wrap this in a hasColumn check. Avoids the issue we had with sales images lmao

database/migrations/2026_02_24_223014_add_description_to_raffles.php
```diff
@@ -0,0 +1,29 @@
...
+            $table->text('parsed_description')->nullable()->default(null);
```
perappu, line 15:
```suggestion
            if (!Schema::hasColumn('raffles', 'parsed_description'))
            {
                $table->text('parsed_description')->nullable()->default(null);
            }
```

less common but just for good measure I'd put it here too

config/lorekeeper/ext-tracker/raffle_options.php
perappu, line 1: Since you added the credits in credits.blade.php, I think this file can get deleted. It'll just create duplicate credits

composer.json
```diff
@@ -43,7 +43,10 @@
     "config": {
         "optimize-autoloader": true,
         "preferred-install": "dist",
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "php-http/discovery": true
+        }
```
perappu, line 49: what's the purpose of this change? It's not immediately evident to me

app/Http/Controllers/RaffleController.php
```diff
@@ -29,11 +29,27 @@ public function getRaffleIndex() {
         } else {
             $raffles->where('is_active', '=', 1);
         }
-        $raffles = $raffles->orderBy('group_id')->orderBy('order');
+        $query = $raffles->orderBy('group_id')
+            ->orderBy('order');
+
+        if (!Auth::check() || Auth::check() && !Auth::user()->hasPower('manage_raffles')) {
```
perappu, line 35:
```suggestion
        if (!Auth::check() || (Auth::check() && !Auth::user()->hasPower('manage_raffles'))) {
```

needs an extra set of parantheses to make the evaluation correctly

perappu (changes requested):

perappu (changes requested):
github is making me leave a comment to request changes, here are some requested changes

app/Http/Controllers/RaffleController.php
```diff
@@ -29,11 +29,27 @@ public function getRaffleIndex() {
         } else {
             $raffles->where('is_active', '=', 1);
         }
-        $raffles = $raffles->orderBy('group_id')->orderBy('order');
+        $query = $raffles->orderBy('group_id')
```
perappu, line 32:
```suggestion
        $query = $raffles->orderBy('group_id', 'DESC')
```

oh additional thing I noticed when pulling is that the oldest raffles are at the top, this will sort it newest first

app/Http/Controllers/RaffleController.php
```diff
@@ -29,11 +29,27 @@ public function getRaffleIndex() {
...
+                    return $q->group->is_active;
+                }
+
+                return $q->is_active;
+            });
+        } else {
+            $query = $query->get();
+        }
+
+        $grouped = $query->groupBy(function ($item) {
```
perappu, line 47:
```suggestion
        $grouped = $query->groupBy(function ($item) {
            return $item->group ? $item->group->name : null;
        })->sortBy(function ($item, $key) {
            return $key !== '';
        });
```

and this will put ungrouped raffles at the top.... ymmv on that but I think it's better. idk there might be a better way to display raffles lol some sites have a lot of them, but i think we can do that later

also I think using null (which gets keyed as an empty string, hence the sortBy callback) is better than "ungrouped"? cuz like. technically someone could make a raffle group called "Ungrouped" but they can't make a raffle group with an empty name

resources/views/raffles/index.blade.php
```diff
@@ -14,29 +14,96 @@
...
-        @if ($prevGroup)
-            </div>
-        @endif
-        <div class="card mb-3">
-            <div class="card-header h3">{{ $groups[$raffle->group_id]->name }}</div>
-            <ul class="list-group list-group-flush">
-    @endif
+        @foreach ($raffles as $key => $raffle)
+            <div class="card mb-3">
+                @if ($key != 'Ungrouped')
```
perappu, line 19:
```suggestion
                @if ($key != '')
```

a change in line with my previous change to the user-side RaffleController.php

app/Http/Controllers/RaffleController.php
SpeedyD: I'd like to suggest this one doesn't get merged in. For raffles, I feel strongly that the OLDER ones should remain at the top, because they should also technically be ROLLED sooner. I'm aware that YMMV, but in general you want attention on the ones expiring sooner.

perappu: It's actually a common complaint I get from users to change it so newer raffles are first lmao, so I kinda had that in mind when I suggested it. It might make more sense to you to have it opposite, but generally I've found people want it newest first.

The complaint is usually about the completed tab though so I guess we could just change it to sort the completed tab by newest first, but there's a consistency issue then... Site owners just really don't like raffles from 4 years ago showing up over raffles from 1 week ago lol

PR #1434, opened 2026-02-26
fix(prompts) consolidate all count arrays for prompt limits into one

notecardPasta:
Just moves a few of the repeating count arrays into the prompt model to cut down on code/easier editing. There's potential for consolidating the limit count even further but I'm not sure what the best way would be so this is something simple!

### DIFF
--- a/app/Models/Prompt/Prompt.php
+++ b/app/Models/Prompt/Prompt.php
@@ -6,7 +6,6 @@ use App\Models\Model;
 use App\Models\Reward\Reward;
 use App\Models\Submission\Submission;
 use Carbon\Carbon;
-use Illuminate\Support\Facades\Auth;
 
 class Prompt extends Model {
     /**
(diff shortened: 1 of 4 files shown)


app/Models/Prompt/Prompt.php
```diff
@@ -84,6 +85,27 @@ public function rewards() {
...
     }
 
+    /**
+     * Get an array of how many prompts the user has completed in general.
+     *
+     * @param mixed $user
+     *
+     * @return array
+     */
+    public function filterCount($user) {
```
ScuffedNewt: nit, but would prefer if this was getCount or similar since filter implies a specific operation

ScuffedNewt (approved):

SpeedyD (approved):
Agree with the nit, but otherwise 🤷

MarskyMessier (approved):
+1 with the nit so it's a bit clearer on the backend but otherwise I approve

itinerare (approved):

PR #1438, opened 2026-03-09
fix: show hidden character ownership logs to staff

perappu:
Went about it this way because it prevents having to either a) use `Auth::user()->hasPower()` on every single character row or b) fix something across multiple locations. (I love global scopes and whereRelation...) 

Also removed some checks that weren't needed given we already check for if character is deleted (aka null).

### DIFF
--- a/app/Models/User/UserCharacterLog.php
+++ b/app/Models/User/UserCharacterLog.php
@@ -4,6 +4,8 @@ namespace App\Models\User;
 
 use App\Models\Character\Character;
 use App\Models\Model;
+use Illuminate\Database\Eloquent\Builder;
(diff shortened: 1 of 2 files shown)


itinerare (approved):

ScuffedNewt (approved):

SpeedyD:
I'm clearly a bit late approving this, but I love this- I was thinking about how to pull this off myself recently, but due to (still?!?) being sick, haven't gotten to it.. glad to see it merged!

I also am sure I would've never figured out to used the booted function.. since I still don't know what it does. XD
(Note to self: research more Laravel functions)

PR #1455, opened 2026-04-26
feat(images): update image handling to prevent dangling images, make delete calls non destructive, and other minor cleanups

ScuffedNewt:
example of dangling images with items
image

also fixes image operations occurring on edit when there is no new image uploaded

no regression with characters or galleries
image
image
image

removed the wrap on the delete call since it now doesn't throw if file doesn't exist


MarskyMessier (approved):

Draconizations (approved):

app/Services/CharacterManager.php
```diff
@@ -196,142 +195,85 @@ public function createCharacter($data, $user, $isMyo = false) {
...
-            if ($imageWidth > $imageHeight) {
-                // Landscape
-                $canvas = Image::canvas($image->width(), $image->width());
-                $image = $canvas->insert($image, 'center');
-            } else {
-                // Portrait
-                $canvas = Image::canvas($image->height(), $image->height());
-                $image = $canvas->insert($image, 'center');
-            }
+            $image = $this->makeImageSquare($image);
```
itinerare, line 209: Just from a quick glance... I'm not spotting where this function is?

itinerare:
Really good cleanup on the whole!

app/Services/CharacterManager.php
ScuffedNewt: ah darn, I didn't commit it 😔

itinerare: Happens to the best of us!

itinerare (approved):

PR #1456, opened 2026-04-26
Add debugbar

### LINKED ISSUE #1274: Laravel debug bar for dependency
Using the debug bar as dev is really nice and I think it would be a good addition to dependencies even if it's just a local tool for devs, I know it's another dep to maintain but I think a lot of folks (especially less experimented devs) would appreciate it.
It'd also help reduce friction when helping others in debug, where the toolbar would give info quicker than having to rummage through folders and settings for sure.
[https://laraveldebugbar.com/](url)

ScuffedNewt:


composer.json
```diff
@@ -22,7 +22,7 @@
         "laravel/socialite": "^5.2",
         "laravel/tinker": "^2.0",
         "laravelcollective/html": "^6.0",
-        "socialiteproviders/deviantart": "^4.2",
```
SpeedyD, line 25: Pretty sure this one is meant to keep the ^

ScuffedNewt: it isn't, otherwise it causes an error with dependencies

perappu: what is the error that gets caused? I've never seen that, so this feels like a weird thing to delete as part of this PR

ScuffedNewt: correction, it was because the lock file was not updated on develop, it's fine with the caret after the lock is updated

perappu: Yeah that makes sense, the PR was to v2 and pulled into v3/dev so it didn't have the lockfile

perappu (approved):

itinerare (approved):

PR #1460, opened 2026-05-04
fix(aliases): Enable 'remember me' option on user login

ScuffedNewt:
prevent alias login from expiring after the default cache expiration

made a v3 pr at request from toko


ScuffedNewt:
original pr: https://github.com/lk-arpg/lorekeeper/pull/1459

perappu (approved):

SpeedyD (approved):
So I read through https://laravel.com/docs/13.x/authentication .. it really is just one word, huh

itinerare (approved):
I can't believe it's just that easy,

PR #1461, opened 2026-05-04
fix(emails): update email send failures to except gracefully

ScuffedNewt:
image
image
image

### DIFF
--- a/app/Actions/Fortify/UpdateUserProfileInformation.php
+++ b/app/Actions/Fortify/UpdateUserProfileInformation.php
@@ -52,7 +52,7 @@ class UpdateUserProfileInformation implements UpdatesUserProfileInformation {
         try {
             $user->sendEmailVerificationNotification();
         } catch (\Throwable $e) {
-            flash('Profile updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.')->warning();
+            flash('Profile updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator.')->warning();
 
             return;
         }
--- a/app/Actions/Fortify/UpdateUserProfileInformation.php
+++ b/app/Actions/Fortify/UpdateUserProfileInformation.php
@@ -49,6 +49,12 @@ class UpdateUserProfileInformation implements UpdatesUserProfileInformation {
             'email_verified_at' => null,
         ])->save();
 
-        $user->sendEmailVerificationNotification();
+        try {
+            $user->sendEmailVerificationNotification();
+        } catch (\Throwable $e) {
+            flash('Profile updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.')->warning();
+
+            return;
+        }
     }
 }
--- a/app/Listeners/SendEmailVerificationNotification.php
+++ b/app/Listeners/SendEmailVerificationNotification.php
@@ -15,7 +15,7 @@ class SendEmailVerificationNotification {
             try {
                 $event->user->sendEmailVerificationNotification();
             } catch (Throwable $e) {
-                flash('Account created successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.')->warning();
+                flash('Account created successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator.')->warning();
             }
         }
     }
--- a/app/Listeners/SendEmailVerificationNotification.php
+++ b/app/Listeners/SendEmailVerificationNotification.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace App\Listeners;
+
+use Illuminate\Auth\Events\Registered;
+use Illuminate\Contracts\Auth\MustVerifyEmail;
+use Throwable;
+
+class SendEmailVerificationNotification {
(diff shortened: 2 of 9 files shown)


SpeedyD (approved):

app/Actions/Fortify/UpdateUserProfileInformation.php
```diff
@@ -49,6 +49,12 @@ protected function updateVerifiedUser($user, array $input) {
             'email_verified_at' => null,
         ])->save();
 
-        $user->sendEmailVerificationNotification();
+        try {
+            $user->sendEmailVerificationNotification();
+        } catch (\Throwable $e) {
+            flash('Profile updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.')->warning();
```
perappu:
```suggestion
            flash('Profile updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator.')->warning();
```

reaaaally nitpicky thing, but we 100% do not want to encourage people to manually verify someone's email in the database. I heckle people about this all the time lol

If someone doesn't get verification emails, they can't get password resets either, which can result in them easily getting locked out of the account. (which can then only be remedied by someone who claims the account is theirs telling an admin to manually change the email... which is ripe for social engineering...)

app/Listeners/SendEmailVerificationNotification.php
```diff
@@ -0,0 +1,22 @@
...
+class SendEmailVerificationNotification {
+    /**
+     * Handle the event.
+     */
+    public function handle(Registered $event) {
+        if ($event->user instanceof MustVerifyEmail && !$event->user->hasVerifiedEmail()) {
+            try {
+                $event->user->sendEmailVerificationNotification();
+            } catch (Throwable $e) {
+                flash('Account created successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.')->warning();
```
perappu:
```suggestion
                flash('Account created successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator.')->warning();
```

same issue noted previously

app/Services/UserService.php
```diff
@@ -168,7 +168,13 @@ public function updateEmail($data, $user) {
         $user->email_verified_at = null;
         $user->save();
 
-        $user->sendEmailVerificationNotification();
+        try {
+            $user->sendEmailVerificationNotification();
+        } catch (\Exception $e) {
+            $this->setError('error', 'Email updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator to verify your email manually.');
```
perappu:
```suggestion
            $this->setError('error', 'Email updated successfully! However, we couldn\'t send the verification email due to email configuration issues. Please contact an administrator.');
```

aaaaand again

perappu (changes requested):

perappu:
Other than the wording stuff I suggested, this looks fine and I'll approve once those changes are accepted.

SpeedyD (changes requested):
I am likewise changing my approval to an RFC until @perappu's suggestions are taken. I did not think about it from that angle, and I fully agree with their line of thinking.

app/Actions/Fortify/UpdateUserProfileInformation.php
ScuffedNewt: I am happy to make the change, however i have a followup pr that adds the ability to verify emails in the edit user UI

yes i agree with password resets etc, however the issue already exists if they type the email wrong etc.
PR https://github.com/lk-arpg/lorekeeper/pull/1409 will cover this case

so going to disagree with this

perappu: I thought all that PR does was add the ability for users to change their own emails? I didn't see a manual verification of emails in that PR? I would be vehemently against it if I _did_ see it there, which I didn't, so...

Even all that aside, I just think all around a simple "please contact an administrator" is way better than suggesting a specific course of action -- because the issue might be misconfigured .env file rather than someone's specific email etc. It's more future-proof and covers more edge cases

ScuffedNewt: > i have a followup pr

ScuffedNewt: I don't see the reason to be against it, I understand the consequences of social engineering but at the same time a lot of sites dealing with these type of issues do not have capacity to resolve them or developers to deal with it. I think it is better UX to allow these features and make consequences known than make it purposefully difficult

itinerare: I'm also pretty strongly against anything more direct than "please contact an administrator"-- even all else being equal, I'd rather error messages be no more involved than "this shouldn't happen, please let someone know."

perappu: > > i have a followup pr

Ah okay I misunderstood, I thought you meant that _was_ the follow up PR. I'm still strongly against that feature and I have a lot of reasons for it, but that's something we can duke out in that specific PR. More pressing issue is just what Merc mentioned

itinerare (approved):

perappu (approved):

itinerare:
That aside, big fan of having this fail more gracefully! Thanks for the work here.

PR #1462, opened 2026-05-04
feat(staff): add secondary admin power

ScuffedNewt:
broken up from original pr https://github.com/lk-arpg/lorekeeper/pull/1409
- Allow non admin ranks to do admin things
   - nice if you don't want to have to give everyone admin to do admin things, especially since you need to remove the role in the DB

secondary admins can't edit their own powers
image

### DIFF
--- a/app/Services/RankService.php
+++ b/app/Services/RankService.php
@@ -39,6 +39,10 @@ class RankService extends Service {
                     if (!config('lorekeeper.powers.'.$power)) {
                         throw new \Exception('Invalid power selected.');
                     }
+
+                    if ($power == 'admin') {
+                        $data['is_secondary_admin'] = true;
+                    }
                 }
 
                 $powers = array_unique($data['powers']);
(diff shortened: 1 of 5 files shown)


app/Services/RankService.php
```diff
@@ -99,6 +103,10 @@ public function updateRank($rank, $data, $user) {
                     if (!config('lorekeeper.powers.'.$power)) {
                         throw new \Exception('Invalid power selected.');
                     }
+
+                    if ($power == 'admin') {
+                        $data['is_secondary_admin'] = true;
+                    }
```
SpeedyD, line 109: Correct me if I'm wrong, but doesn't this also need an opposite? If power is NOT admin, set to false?

SpeedyD (commented):
Above question is my only question. If answered or fixed, you got my approval. 👍

itinerare:
Thanks for separating this out!

app/Services/RankService.php
ScuffedNewt: i believe the default behaviour is that if it is undefined it sets to false, but i can test it later

SpeedyD: Please do. :)

perappu:
+1 what Speedy said, I think adding an explicit else-then-false condition could be a good call for something as hefty as an admin powers check (even if the default behavior is false). 

other than that, looks good to me!

perappu (approved):

SpeedyD (approved):

itinerare (approved):

PR #1465, opened 2026-05-04
Criteria Rewards

ScuffedNewt:
# Adds the following new features from the extension:
- Ability to edit and add criteria to drafts / gallery submission / prompt submissions (admin side)
- Remove added criteria (admin side)
- Character criteria thanks to @perappu 

image

## Character Criteria
image

## Collaborator Rewards
image

image
- Provides same as core functionality utilising criteria

## Participant Rewards
- Provides equivalent functionality to core

image

image

image


app/Services/SubmissionManager.php
```diff
@@ -467,6 +488,24 @@ public function approveSubmission($data, $user) {
...
+            if (isset($data['criterion'])) {
+                foreach ($data['criterion'] as $key => $criterionData) {
+                    $criterion = Criterion::where('id', $criterionData['id'])->first();
+                    if (isset($criterionData['criterion_currency_id'])) {
+                        $criterion_currency = Currency::find($criterionData['criterion_currency_id']);
+                    } else {
+                        $criterion_currency = $criterion->currency;
+                    }
+
+                    if (!$service->creditCurrency($user, $submission->user, $promptLogType, $promptData['data'], $criterion_currency, $criterion->calculateReward($criterionData))) {
```
ScuffedNewt: this should actually add to the asset array (current is original implementation but it shows as two logs instead of one)

Camyza:
Found a issue from LK vanilla that limits the data being sent to the database for those with longer criterion data for galleries originally on v3.  The data column in the gallery_submissions table needs to be adjusted and I'm not sure what is the best way. I changed mine from varchar with a length limit of 1024 to text with no limit since I was testing on my site to see if it worked to show Gallery's reward info in the detailed log link before being submitted.  It took a while but finding this out when it's been an issue since last year back in April was a huge sigh of relief.

The exact line is $table->string('data', 1024)->nullable()->default(null);  in the mysql-schema.dump file.

(Hope this is okay to put here.)

perappu:
> Found a issue from LK vanilla that limits the data being sent to the database for those with longer criterion data for galleries originally on v3. The data column in the gallery_submissions table needs to be adjusted and I'm not sure what is the best way. I changed mine from varchar with a length limit of 1024 to text with no limit since I was testing on my site to see if it worked to show Gallery's reward info in the detailed log link before being submitted. It took a while but finding this out when it's been an issue since last year back in April was a huge sigh of relief.
> 
> The exact line is $table->string('data', 1024)->nullable()->default(null); in the mysql-schema.dump file.

This is a good call out. I think for this PR, changing it to a `$table->json()` column would be slightly more ideal (json column essentially builds on longtext() but has DB level validation that the data needs to be in a json format)

Issue #1424, opened 2026-01-30
Item search view gets overloaded on big queries


MarskyMessier:
Turns out if you're one of the LK georg, having too many users / items being held will crash the item search page when it gets too much to render.
This should be fixed by a paginated query, and two tabs for both characters & users (Ideally in separate sub-files so it's not one big same mix).

PR #1470, opened 2026-05-07
Fix: use tab and paginate the item search page

MarskyMessier:
There was an issue with item search where having an item owned by a lot of people would completely crash the search because of the amount of queries at once ( #1424 ) , so I paginated it and made it so users and characters are in separate tab.


ScuffedNewt (approved):

SpeedyD (approved):

itinerare:
I want to take a look at this to make sure it's not running afoul of a quirk of pagination I've run into prior, actually...

MarskyMessier:
@itinerare Following our conversation on discord (tl;dr - pagination being quirky when synchronised on different queries on the same page, sometimes) would you rather have this being moved to an asynchronised pagination or do you think it's fine to let it as such for now?

itinerare:
To be fair, I think it might be more of a UX issue than a "correctness" one; if it's paginating in sync successfully at the moment, I'm not too troubled by it... it just remains to be seen whether that's *expected* behavior.

That said, having mulled it over, I'm fine leaving that for a subsequent PR if folk feel it should be otherwise.

PR #1477, opened 2026-05-12
fix: make fontawesome persist and viewable in tinyMCE

strabii:
this is a small fix/enhancement to make fontawesome work live within the tinyMCE editor!

ps. I realized I put this on the wrong branch earlier, I only properly tested this on release v3.0.0 so it should be going towards the correct location now :')

**current tinyMCE behavior:**
- fontawesome code (`<i class="fas fa-heart"></i>`) gets automatically stripped in the tinyMCE editor because it's considered an empty element.
- workarounds to get it to persist must be done manually every time (the comment method or typing `&nbsp;`)
- even using workarounds to make sure the code isn't stripped, FA will not display a live preview in the editor (it will render properly on site pages/profiles/comments though)

**testing outcome:**
- FA will now display in tinyMCE's live preview while editing it.
- this will prevent the autoconversion to `<em>` tags when `<i>` is used.
- only `<i>` and `<em>` **"empty"** code will always automatically pad with a `&nbsp;` so it'll never get eaten again by accident (`<i class="fas fa-heart">&nbsp;</i>`). non-empty/normal use of italics code is not affected ([test shown here](https://github.com/lk-arpg/lorekeeper/pull/1477#discussion_r3227848193))

> **notes:**
> - the fix to make FA show up in tinyMCE was simply adding in the css in the content_css lol
> - the `extended_valid_elements` is set to only target the class attribute when used in `<i>` and `<em>` attributes, so hopefully it shouldn't be affecting anything else. both of these are included to account for the editor's previous behavior with the autoconverting tags to `<em>`, and the autopadding should only target empty elements too (so anyone using the noted workarounds should also be unaffected).
>   - (to be technical, we're not limited to `<i>` to use FA, folks can legalize other elements in the editor if they want, but for the sake of brevity, the average user is most likely not going to know alternatives are possible because of the default copy+paste code the FA website gives.)
> - the autopadding behavior is set with the `#` in front (in `extended_valid_elements: '#i[class]',`); it's set as a default, since this behavior probably won't show up outside the use of FA and accidentally losing tags is more of the unwanted behavior. individuals _can_ remove the `#` if desired though, but they'll be on their own with understanding what doesn't trigger the eating behavior.
>   - while it would be nice for the autopadding to spawn with a comment (`<!-- -->`) so there won't be a visible space, I don't think this is possible without some more complicated code or a plugin at that point. however, users can still manually edit the autopadded `&nbsp;` to a comment if they prefer instead.

### DIFF
--- a/resources/views/character/_image_js.blade.php
+++ b/resources/views/character/_image_js.blade.php
@@ -20,9 +20,11 @@
                     toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                     content_css: [
                         '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
+                        '{{ asset('css/lorekeeper.css') }}',
+                        '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
+                    extended_valid_elements: '#i[class],#em[class]',
                     target_list: false
                 });
             });
@@ -64,7 +66,8 @@
                     toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                     content_css: [
                         '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
+                        '{{ asset('css/lorekeeper.css') }}',
+                        '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
                     target_list: false
--- a/resources/views/character/_image_js.blade.php
+++ b/resources/views/character/_image_js.blade.php
@@ -70,6 +70,7 @@
                         '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
+                    extended_valid_elements: '#i[class],#em[class]',
                     target_list: false
                 });
             });
--- a/resources/views/comments/comments.blade.php
+++ b/resources/views/comments/comments.blade.php
@@ -98,9 +98,11 @@
                 toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                 content_css: [
                     '{{ asset('css/app.css') }}',
-                    '{{ asset('css/lorekeeper.css') }}'
+                    '{{ asset('css/lorekeeper.css') }}',
+                    '{{ asset('css/all.min.css') }}', // fontawesome
                 ],
                 spoiler_caption: 'Toggle Spoiler',
+                extended_valid_elements: '#i[class],#em[class]',
                 target_list: false
             });
         });
--- a/resources/views/js/_modal_wysiwyg.blade.php
+++ b/resources/views/js/_modal_wysiwyg.blade.php
@@ -8,9 +8,11 @@ plugins: [
 'insertdatetime media table paste code help wordcount'
 ],
 toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | removeformat | code',
+extended_valid_elements: '#i[class],#em[class]',
 content_css: [
 '//www.tiny.cloud/css/codepen.min.css',
 '{{ asset('css/app.css') }}',
-'{{ asset('css/lorekeeper.css') }}'
+'{{ asset('css/lorekeeper.css') }}',
+'{{ asset('css/all.min.css') }}', // fontawesome
 ]
 });
--- a/resources/views/js/_modal_wysiwyg.blade.php
+++ b/resources/views/js/_modal_wysiwyg.blade.php
@@ -8,11 +8,11 @@ plugins: [
 'insertdatetime media table paste code help wordcount'
 ],
 toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | removeformat | code',
-extended_valid_elements: '#i[class],#em[class]',
 content_css: [
 '//www.tiny.cloud/css/codepen.min.css',
 '{{ asset('css/app.css') }}',
 '{{ asset('css/lorekeeper.css') }}',
 '{{ asset('css/all.min.css') }}', // fontawesome
-]
+],
+extended_valid_elements: '#i[class],#em[class]',
 });
--- a/resources/views/layouts/app.blade.php
+++ b/resources/views/layouts/app.blade.php
@@ -170,9 +170,11 @@
                     toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                     content_css: [
                         '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
+                        '{{ asset('css/lorekeeper.css') }}',
+                        '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
+                    extended_valid_elements: '#i[class],#em[class]', // # <- sets autopadding with &nbsp;
                     target_list: false
                 });
                 bsCustomFileInput.init();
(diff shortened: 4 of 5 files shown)


resources/views/layouts/app.blade.php
```diff
@@ -170,9 +170,11 @@
                     toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                     content_css: [
                         '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
+                        '{{ asset('css/lorekeeper.css') }}',
+                        '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
+                    extended_valid_elements: '#i[class],#em[class]', // # <- sets autopadding with &nbsp;
```
perappu, line 177: I don't have the opportunity to test this myself at the moment, so!

Did you verify that this change doesn't mess things up if someone uses these tags as "intended" like `<i>example text</i>` or `<em>example text</em>`?

resources/views/comments/comments.blade.php
```diff
@@ -98,9 +98,11 @@
                 toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                 content_css: [
                     '{{ asset('css/app.css') }}',
-                    '{{ asset('css/lorekeeper.css') }}'
+                    '{{ asset('css/lorekeeper.css') }}',
+                    '{{ asset('css/all.min.css') }}', // fontawesome
                 ],
                 spoiler_caption: 'Toggle Spoiler',
+                extended_valid_elements: '#i[class],#em[class]',
```
perappu, line 105: just making a note for my own reference that my other comment would apply here too

perappu (commented):
This is a nice little bugfix, thank you again for being a new contributor! ❤️ Just one question in the comments, and then once that's resolved I'll be happy to put an approval on this.

resources/views/layouts/app.blade.php
strabii: hi!! thank you for reviewing!!

afaik it shouldn't interfere with normal use of the tags. I've attached two recordings one with the # behavior added and one without. tinyMCE does crunch the code already when switching from code mode to liveeditor but a lot of the consequences of that happening probably isn't really relevant to the code addition 🤔

Though I was initially a bit hesitant with setting the autopadding on by default, since some of the weirder padding behavior it spawns with might bother coders (it pushes up the fontawesome but manually writing nbsp doesnt?), but to the average non-coder having your tags deleted might be more confusing unwanted? ALSO yeah it was good to doublecheck since whoops it looks like the autopadding will also autopad empty `<i></i>` that might also be unwanted behavior 🥴 (I attempted to swap the # inside the bracket (`i[#class]`) to see if it can only target class attribute additions but this broke it (everything gets deleted) so this is not a possible alternative fix orz.)

version where # is added:
https://github.com/user-attachments/assets/b52a41a1-2613-44b1-811e-a4acbdecc395

version where it does not autopad:
https://github.com/user-attachments/assets/cf91005d-df51-4231-8925-02ecf1987ff9

SpeedyD:
I'm not sure if this is a good idea.. despite this looks like it could be a great fix.

The reason for that is that there has been a PR for the develop branch that _consolidated all the TinyMCE_ instances into a single file, and.. that would actually be a much better and easier, way to handle this.. (See also PR #1210.)

Especially since it would only be required _once_ to be everywhere on Lorekeeper sites, instead of repeating it 5 times, which incidentally, isn't even all the times TinyMCE is used on Lorekeeper. (There's well over 20 instances.)

So effectively, my suggestion is to move this to the develop branch.

itinerare:
No, I explicitly requested for this to target the release branch.

SpeedyD:
> No, I explicitly requested for this to target the release branch.

🤔

Well, if you requested/directed this for release, I'm assuming the Monday merge will keep an eye on this distinct difference.. Then I have no qualms, I suppose.

itinerare:
Changes being awkward to merge back into develop just isn't a compelling reason not to target the release branch for applicable fixes... there's no call to hold something that'll provide QoL back just to save a few minutes handling merge conflicts.

resources/views/layouts/app.blade.php
perappu: I think I'm okay with it keeping empty `<i></i>` tags... it's certainly not the weirdest thing TinyMCE does, lol.

My concern was mainly with whether or not it would replace any "normal" contents with a `&nbsp;`. If it doesn't do that, then I'm fine with this!

perappu:
Yeah, I have faith in Merc to be able to copy the fix over to any necessary `tinymce.init()`s in develop.

Though that does beg the question, @strabii would you be comfortable copying over this fix to other `tinymce.init()` instances in the code? There's a, uh, lot of them, but I think it'd be good to be thorough with this.

resources/views/layouts/app.blade.php
itinerare: Yeah, I figured that keeping empty `<i>` tags is harmless enough if undesired, and enables some desired behavior that currently simply doesn't work.

strabii:
@perappu I think I can do, so but could you clarify where or how? by where I mean specificially is this still in release or am I going to do this in other branches (like main or dev)? (I have less experience with dev though and I'm not sure what to expect from the comments bringing up the tinymce changes on dev) :0

and by how I mean my searching for `tinymce.int(` only found 5 files (6 instances) of tinymce.init() being initialized in release, so is there something different to search to find the rest of them?
image

resources/views/js/_modal_wysiwyg.blade.php
```diff
@@ -8,9 +8,11 @@
...
 ],
 toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | removeformat | code',
+extended_valid_elements: '#i[class],#em[class]',
 content_css: [
 '//www.tiny.cloud/css/codepen.min.css',
 '{{ asset('css/app.css') }}',
-'{{ asset('css/lorekeeper.css') }}'
+'{{ asset('css/lorekeeper.css') }}',
+'{{ asset('css/all.min.css') }}', // fontawesome
 ]
```
SpeedyD: Nitpick: You have the extended_valid_elements BELOW the assets everywhere else, but ABOVE the assets here?

resources/views/character/_image_js.blade.php
```diff
@@ -64,7 +66,8 @@
                     toolbar: 'undo redo | formatselect | bold italic backcolor | alignleft aligncenter alignright alignjustify | bullist numlist outdent indent | link image | spoiler-add spoiler-remove | removeformat | code',
                     content_css: [
                         '{{ asset('css/app.css') }}',
-                        '{{ asset('css/lorekeeper.css') }}'
+                        '{{ asset('css/lorekeeper.css') }}',
+                        '{{ asset('css/all.min.css') }}', // fontawesome
                     ],
                     spoiler_caption: 'Toggle Spoiler',
```
SpeedyD, line 72: No extended_valid_elements here?

SpeedyD:
> Though that does beg the question, @strabii would you be comfortable copying over this fix to other `tinymce.init()` instances in the code? There's a, uh, lot of them, but I think it'd be good to be thorough with this.

@perappu: One of the goals of the consolidation project was to ensure it ONLY loaded on pages that actually used it, as opposed to app.blade.php and let it load everywhere.. which, as you can tell from strabii's screenshot, is in the deadlast search result.

There's 20+ instances, but only 6 inits overall.

SpeedyD (changes requested):
On the other hand, I did see some things I'd like to see fixed either way. :)

perappu:
> I think I can do, so but could you clarify where or how? by where I mean specificially is this still in release or am I going to do this in other branches (like main or dev)? (I have less experience with dev though and I'm not sure what to expect from the comments bringing up the tinymce changes on dev) :0

Yeah your screenshot has all of them! I misunderstood what Speedy said and forgot how often tinymce is actually initialized lol. That's my only ask, having this fix on all the tinymce.init()s. Which I think you do now that I'm looking closer, so disregard me LOL

Don't get too turned around, this is still a v3 PR and it'll just get pulled forward into dev and handled by Mercury when it gets merged.

resources/views/character/_image_js.blade.php
strabii: oops missed it here! added it back in

resources/views/js/_modal_wysiwyg.blade.php
strabii: ah it was because nothing was put underneath content_css for the js modal so I thought it might've been intentional it should remain as the last element. if I should still move extended elements down under it though just say the word!

SpeedyD: 100% just a nitpick, doesn't really matter. Technically all 6 inits should be the exact same but they are, uh, not. Not a big deal, though- and perhaps in this case better. (It'll get fixed once the develop branch becomes main or release.)

So don't worry, just close this if you feel it unnecessary.

SpeedyD (approved):

resources/views/js/_modal_wysiwyg.blade.php
strabii: I think it's cool! standardizing always makes sense, so I moved it down :D

SpeedyD (approved):

perappu (approved):

PR #1478, opened 2026-05-12
fix: avatar upload validation rules

strabii:
UMMM while working on my own lk the other day I noticed there's a complete lack of avatar validation rules. So despite users being limited in their file uploads elsewhere (masterlist/galleries), they're able to upload files of any size and technically pick any type of file for avatars. I'm not sure but the latter feels like it could pose a possible security risk?

**current behavior:**
- uploading insanely big images on live servers will crash it with a "413 Content Too Large" and the failure to upload an avatar causes the user's entire profile page (and presumably wherever the avatar shows) unrenderable.
- on local testing, insanely big images can get through in their full size, and assuming on a live server if the file avoids the content too large error somehow, and this would cause massive eating of filehosting space and very quickly. This would also be very unoptimal for users browsing their own pages if their browser is reloading a huge image every time they pass by it (I know it would cause a lot of data drain for me...).
- users can upload any type of file type, albeit the image processing system will reject it, who knows if something malicious can get through it...

**fixes:**
- set a maximum size on user's avatar uploads (default 1mb, though idk if most people would want something larger?)
- set image types accepted for the avatar upload (jpeg, jpg, gif, png, webp)


app/Models/User/User.php
```diff
@@ -83,6 +83,15 @@ class User extends Authenticatable implements MustVerifyEmail {
      */
     public $timestamps = true;
 
+    /**
+     * Validation rules for avatar creation.
+     *
+     * @var array
+     */
+    public static $createRules = [
+        'avatar'      => 'required|mimes:jpeg,jpg,gif,png,webp|max:1024',
```
perappu, line 92:
```suggestion
        'avatar'      => 'nullable|mimes:jpeg,jpg,gif,png,webp|max:1024',
```

Users aren't required to upload a brand new avatar every single time the model is updated, so this should be a "nullable" rather than a "required".

perappu (changes requested):
Thank you for PRing! I know it can be kinda scary doing it for the first time, but we're always happy to see new contributors :D

Not sure if this is a concern for you but I know it can make some people nervous, so please don't take the request for changes personally, the review process is here to help make your changes as perfect as possible ❤️

app/Http/Controllers/Users/AccountController.php
```diff
@@ -86,6 +86,7 @@ public function postProfile(Request $request) {
      * @return \Illuminate\Http\RedirectResponse
      */
     public function postAvatar(Request $request, UserService $service) {
+        $request->validate(User::$createRules);
```
perappu:
```suggestion
        $request->validate(User::$updateRules);
```

Maybe kind of pedantic, but you're updating the model, not creating a whole new user, so this should be an `$updateRules`.

app/Models/User/User.php
```diff
@@ -83,6 +83,15 @@ class User extends Authenticatable implements MustVerifyEmail {
      */
     public $timestamps = true;
 
+    /**
+     * Validation rules for avatar creation.
+     *
+     * @var array
+     */
+    public static $createRules = [
```
perappu:
```suggestion
    public static $updateRules = [
```

Just a change in line with my other comment

app/Models/User/User.php
```diff
@@ -83,6 +83,15 @@ class User extends Authenticatable implements MustVerifyEmail {
      */
     public $timestamps = true;
 
+    /**
+     * Validation rules for avatar creation.
```
perappu:
```suggestion
     * Validation rules for updating.
```

Just updating the comment in line w/ other recommendations

perappu (changes requested):

app/Http/Controllers/Users/AccountController.php
perappu: Alternatively when I think about it more we could make it explicitly `$avatarUpdateRules`, since it's _just_ the update rules for the avatar and not the whole user. User models are weird... what's your thoughts @itinerare ?

itinerare: Yeah, with how broken-up handling for updates to users is, it wouldn't be a bad idea to be more explicit about what validation rules are for... Though in that case, the rules *should* use `required` (and arguably being able to do this for different operations is a decent reason to have more specific validation rules).

strabii: all the suggested changes make sense!

I was referencing design update uploads so that's why the required rule carried over haha. I'll batch and add the suggested current edits for now but I'm guessing more discussion is to be had on what we want the final update rules/variables to use? 🤔

perappu: Yeah I like the idea of breaking up user validation with `$avatarUpdateRules` when I think about it more. I'd say go ahead and change your variables/comments to reference that instead!

And like Merc said, you'll want to bring back `required` since we're making it specifically for avatars.

app/Http/Controllers/Users/AccountController.php
```diff
@@ -86,6 +86,7 @@ public function postProfile(Request $request) {
      * @return \Illuminate\Http\RedirectResponse
      */
     public function postAvatar(Request $request, UserService $service) {
+        $request->validate(User::$updateRules);
```
strabii:
```suggestion
        $request->validate(User::$avatarUpdateRules);
```

app/Models/User/User.php
```diff
@@ -83,6 +83,15 @@ class User extends Authenticatable implements MustVerifyEmail {
      */
     public $timestamps = true;
 
+    /**
+     * Validation rules for updating.
+     *
+     * @var array
+     */
+    public static $updateRules = [
+        'avatar'      => 'nullable|mimes:jpeg,jpg,gif,png,webp|max:1024',
```
strabii:
```suggestion
        'avatar'      => 'required|mimes:jpeg,jpg,gif,png,webp|max:1024',
```

app/Models/User/User.php
```diff
@@ -83,6 +83,15 @@ class User extends Authenticatable implements MustVerifyEmail {
      */
     public $timestamps = true;
 
+    /**
+     * Validation rules for updating.
+     *
+     * @var array
+     */
+    public static $updateRules = [
```
strabii:
```suggestion
    public static $avatarUpdateRules = [
```

strabii (commented):
changed variable to $avatarUpdateRules and added "required" back to validation

app/Http/Controllers/Users/AccountController.php
strabii: cool!! the variable should be updated and readded required back! c:

itinerare (approved):

SpeedyD (approved):

perappu (approved):

PR #1479, opened 2026-05-13
refactor: optimize isStaff and hasPower checks

perappu:
since rank and its powers are always eager loaded on a user, an infuriatingly tiny solution to hopefully most of our rank problems


itinerare (approved):
oh my *god*

ScuffedNewt (approved):

MarskyMessier (approved):

PR #1482, opened 2026-05-20
fix: prev/next character buttons not wrapping on mobile

perappu:
fixes this:
vivaldi_egMUKidaYk

into this:
image

note: when this gets merged forward, text-wrap should be included in dev's 4.6 bootstrap already (but I don't think leaving it in will hurt)

### DIFF
--- a/public/css/lorekeeper.css
+++ b/public/css/lorekeeper.css
@@ -456,6 +456,9 @@ main > .row {
     height: 2rem;
     padding-right: .5rem;
 }
+.text-wrap {
+    text-wrap: auto;
+}
 
 /* NOTIFICATIONS *********************************************************************************/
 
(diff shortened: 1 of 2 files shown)


itinerare (approved):

SpeedyD:
~~I'd like to approve this, but I'm going to have to wonder why you removed the `float-left` and `float-right` classes.~~

Never mind. Apparently it was never needed. Text-right is enough, as is the clean columns. Why didn't I do that properly before..

SpeedyD (approved):

perappu:
> ~I'd like to approve this, but I'm going to have to wonder why you removed the `float-left` and `float-right` classes.~
> 
> Never mind. Apparently it was never needed. Text-right is enough, as is the clean columns. Why didn't I do that properly before..

Yeah LOL it's completely redundant, glad you figured it out. Pretty much everything else in that section has float on it so makes sense you'd stick it there too

SpeedyD:
> Yeah LOL it's completely redundant, glad you figured it out. Pretty much everything else in that section has float on it so makes sense you'd stick it there too

Yeah, I definitely just wanted to stay consistent, only to find out now how unnecessary it is XD

PR #1485, opened 2026-05-24
feat(images): let there be gifs🧙

ScuffedNewt:


SpeedyD (changes requested):
You added the new column to FeatureCategories, but didn't edit the model.

While I'm on the subject, you did not add this to News posts and SitePages, which I'm sure can also benefit from this.

ScuffedNewt:
> You added the new column to FeatureCategories, but didn't edit the model.
> 
> While I'm on the subject, you did not add this to News posts and SitePages, which I'm sure can also benefit from this.

Oopsies, I'll go back and add it to feature categories 
I don't plan on adding it to news or site pages, someone else can do that if they wish

SpeedyD:
> I don't plan on adding it to news or site pages, someone else can do that if they wish

...Incoming PR soon, then.

app/Services/PromptService.php
```diff
@@ -206,7 +209,7 @@ public function createPrompt($data, $user) {
 
             $prompt = Prompt::create(Arr::only($data, [
                 'prompt_category_id', 'name', 'summary', 'description', 'parsed_description', 'is_active', 'start_at', 'end_at', 'hide_before_start', 'hide_after_end', 'has_image', 'prefix', 'hide_submissions', 'staff_only', 'hash',
-                'limit', 'limit_period', 'limit_character',
+                'limit', 'limit_period', 'limit_character', 'image_extension',
             ]));
```
SpeedyD, line 213: This and UpdatePrompt (L90) are the only times in this entire PR that 'image_extension' is added to the Arr list. Can you explain why? 🤔 I'm probably overlooking something..

app/Services/FeatureService.php
SpeedyD:
```suggestion
                $data['image_extension'] = $data['image']->getClientOriginalExtension();
                $data['hash'] = randomString(10);
```
Another for createFeatureCategory()

app/Services/PromptService.php
ScuffedNewt: the other updates use $data directly in the update / create

app/Services/CurrencyService.php
```diff
@@ -206,6 +208,7 @@ public function createCurrency($data, $user) {
             $icon = $image = null;
```
ScuffedNewt: clarity

database/migrations/2026_05_24_094925_add_more_image_types.php
```diff
@@ -0,0 +1,116 @@
...
+            });
+            Schema::table('currency_categories', function (Blueprint $table) {
+                $table->dropColumn('image_extension');
+            });
+            //
+            Schema::table('shops', function (Blueprint $table) {
+                $table->dropColumn('image_extension');
+            });
+        } catch (Exception $e) {
+            //
```
perappu: I'll give a more thorough review later, but just want to immediately point out that an empty catch block is capital b Bad.

if your goal is to avoid an error with dropColumn if the column doesn't exist, it would be muuuuch better to check for the existence of the column on each table with an if statement. Just wrap each Schema::table in an individual `if(Schema::hasColumn('tablename', 'image_extension'))` or something

ScuffedNewt: this was so i could run rollback on local, it has nothing to do with attempting to catch

SpeedyD: I've added `dropIfExists` and removed the try-catch in a PR I'm making for this - that's why all my previous suggestions disappeared ^^;

perappu: Okay cool that's fine! I figured it was to help with local development, I just didn't want it going into prod like that

app/Services/PromptService.php
SpeedyD: Ah, thanks for pointing that out. Now I'm curious why it's not happening here, but that's a question for another time. :)

SpeedyD (commented):
As I've added to this PR, I don't think I can officially approve it - though of course I do :P

I'd like to ask @perappu to look this over again when they got the time :)

SpeedyD (approved):
~~Meh, screw it, might as well~~

itinerare:
My only question is why the image validation rules are being removed, rather than added to....

perappu:
> My only question is why the image validation rules are being removed, rather than added to....

Yeah catching stuff like this is why I wanted to look closer at the PR before approving lol

while these are admin side so the chance of it mattering is super slim, it can be a security flaw to not do file type checking. Just a standard allowing png/jpg/jpeg/webp/gif should be fine I would think

ScuffedNewt:
Tbh my thoughts process was more just: let people upload things like avif, twitter's awful jpg copy etc, but happy to rescope it down to more traditional formats

itinerare:
I'd much rather just have validation rules covering most expected use cases... it's reasonable to expect people with more unusual wants to add to them if needed, I think.

SpeedyD:
🫡

PR #1493, opened 2026-06-05
fix: patch for CRLF injection vulnerability in email inputs

perappu:
Patch for laravel/framework security advisory https://github.com/laravel/framework/security/advisories/GHSA-5vg9-5847-vvmq

Adds middleware that runs `filter_var($val, FILTER_SANITIZE_EMAIL)` against any form request field labeled "email".

Notably:
- The malicious string `victim@example.com\r\nBcc: attacker@evil.com` is already blocked by frontend validation (had to remove that when testing this!).
- I saw an article indicating that if you use the native validate() with your form data then you're safe, so technically we may already be fine? It did seem to be getting blocked by just the normal 'email' validation rule when I was testing this (and as far as I can tell, we never accept an email without validating it first). Still, sanitization != validation -- this will sanitize it before it even hits the validation rule.

**Also, important note:** Because I ran into this... people should run `composer self-update` when they upgrade, because I noticed I had an older version of composer that didn't recognize the newer `policy.advisories.ignore-id` config.

### DIFF
--- a/app/Http/Kernel.php
+++ b/app/Http/Kernel.php
@@ -20,6 +20,7 @@ class Kernel extends HttpKernel
         \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
         \App\Http\Middleware\TrustProxies::class,
         \App\Http\Middleware\ParsePostRequestFields::class,
+        \App\Http\Middleware\SanitizeEmails::class,
     ];
 
     /**
--- a/app/Http/Middleware/SanitizeEmails.php
+++ b/app/Http/Middleware/SanitizeEmails.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace App\Http\Middleware;
+
+use Closure;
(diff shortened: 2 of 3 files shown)


ScuffedNewt (approved):

itinerare (approved):

itinerare:
I _was_ wondering if the `email` validation rule alone sufficed, but better safe than sorry with issues like this.

perappu:
> I _was_ wondering if the `email` validation rule alone sufficed, but better safe than sorry with issues like this.

Yeah I wasn't 100% sure because of the way the security advisory was worded haha, I thought maybe it was an issue with the email validation rule itself? but I guess it must be some validation internal to the symfony mailer or something instead...

itinerare:
I guess it's one of those things where it *should* reasonably be covered out of the box, but is resolvable at the application layer regardless...

itinerare:
(Holding for release on Monday as per usual, incidentally.)

PR #1495, opened 2026-06-10
Fixes(Ranks): Sorting and Power Changes

SpeedyD:
I noticed how secondary admin can still not be sorted, and it kinda made me realize that admin ranks cannot be sorted at all.. Despite that some sites may have it differently. Also, the whole.. all of the ranks is a bit much?

Here's the gist:
- Secondary "Admin" power ranks can now be sorted, but it cannot be dead last.
  - As the default user rank is last, I figured it's good to prevent admin ranks to be sorted there.
- Neither main Admin rank nor Default user rank can be sorted or edited.
  - Default user rank can also no longer have powers.
- "Admin" power ranks now have an abbreviated list in the ranks list.
  - With an added help that explains it has all powers.
- The rank sorted last now has an extra mention it's the Default User Rank.
  - With an added help explaining it's the rank that will be given automatically to users.
- The Default User rank can not be deleted anymore.
  - This will prevent admin from deleting all ranks except admin and giving all new users the admin rank by default.

We should probably allow admin to _choose_ which rank should be the default user rank, but for now I figured this was an easy and quick way to fix this without adding more to the database. :)

### DIFF
--- a/app/Models/Rank/Rank.php
+++ b/app/Models/Rank/Rank.php
@@ -85,6 +85,39 @@ class Rank extends Model {
         return $this->attributes['is_admin'] || $this->attributes['is_secondary_admin'];
     }
 
+    /**
+     * Check if the rank is a secondary admin rank.
+     *
+     * @return bool
+     */
+    public function getIsSecondaryAdminAttribute() {
+        return $this->attributes['is_secondary_admin'];
+    }
+
+    /**
+     * Check if the rank is the default user rank.
+     *
+     * @return bool
+     */
+    public function getIsDefaultRankAttribute() {
+        if ($this->sort == 0) {
+            return true;
+        }
+        return false;
+    }
+
+    /**
+     * Check if the rank can be sorted.
+     *
+     * @return bool
+     */
+    public function getIsSortableAttribute() {
+        if ($this->attributes['is_admin'] || $this->isDefaultRank) {
+            return false;
+        }
+        return true;
+    }
+
     /**********************************************************************************************
 
         OTHER FUNCTIONS
@@ -104,23 +137,29 @@ class Rank extends Model {
         }
         if ($this->hasPower('edit_ranks')) {
             if ($this->isAdmin) {
-                // editing a false admin rank
-                if ($rank->powers()->where('power', 'admin')->exists()) {
-                    if ($this->attributes['is_admin']) {
-                        return 3; // must remove admin power to edit more granularly
-                    } else {
-                        return 4; // false admin rank, cannot edit
-                    }
+                if ($rank->attributes['is_admin']) {
+                    return 2; // is admin, cannot edit
                 }
 
                 if ($rank->id != $this->id) {
-                    return 1;
-                } // can edit everything
+                    if ($this->sort > $rank->sort) {
+                        if ($rank->IsSecondaryAdmin) {
+                            return 3; // editing a false admin rank, must remove admin power to edit more granularly
+                        }
+                        else {
+                            return 1; // can edit everything, as admin
+                        }
+                    }
+                    else {
+                        return 4; // cannot edit higher rank
+                    }
+                }
                 else {
-                    return 2;
-                } // limited edit: cannot edit sort order/powers
+                    return 4; // cannot edit own rank
+                }
+
             } elseif ($this->sort > $rank->sort) {
-                return 1;
+                return 1; // can edit everything, rank is higher
             }
         }
 
--- a/app/Models/Rank/Rank.php
+++ b/app/Models/Rank/Rank.php
@@ -103,6 +103,7 @@ class Rank extends Model {
         if ($this->sort == 0) {
             return true;
         }
+
         return false;
     }
 
@@ -115,6 +116,7 @@ class Rank extends Model {
         if ($this->attributes['is_admin'] || $this->isDefaultRank) {
             return false;
         }
+
         return true;
     }
 
@@ -145,19 +147,15 @@ class Rank extends Model {
                     if ($this->sort > $rank->sort) {
                         if ($rank->IsSecondaryAdmin) {
                             return 3; // editing a false admin rank, must remove admin power to edit more granularly
-                        }
-                        else {
+                        } else {
                             return 1; // can edit everything, as admin
                         }
-                    }
-                    else {
+                    } else {
                         return 4; // cannot edit higher rank
                     }
-                }
-                else {
+                } else {
                     return 4; // cannot edit own rank
                 }
-
             } elseif ($this->sort > $rank->sort) {
                 return 1; // can edit everything, rank is higher
             }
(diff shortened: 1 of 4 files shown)


perappu:
I'm not sure I really understand the purpose of this PR.... there are many core systems and extensions that are reliant on admin rank being sorted to the top & default user rank being sorted to the bottom. I think it could have very large repercussions if we changed this behavior. I think it's also good UX to enforce "the person with the most powers is at the top, the person with the least is at the bottom."

Some of these changes are good (like preventing users from deleting the default rank, additional clarification on the default rank behavior) but I'm not sure about altering sorting.

SpeedyD:
Honestly? The problem is that the sorting focuses on 'isAdmin'.. which is also answered with a true on secondary admins, making them impossible to sort.

I made a new user for secondary admins and it was just frozen on sort position 1. No matter what I did. And some edits wouldn't go through for some reason. (Translation: It's second from the bottom.. And somehow, the rank above it also had sort position 1.)

I would like to however know..
> there are many core systems and extensions that are reliant on admin rank being sorted to the top & default user rank being sorted to the bottom.

...which, precisely? Because I think a reliance like that may actually be quite a problematic method.

Which makes me still think it might be a good idea to make an 'is default user rank' checkbox somewhere, though that will obviously require more than just a little bit of code like this.. and a migration. Probably a different PR.

perappu:
Here are several places I know off hand that are in core:
https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Actions/Fortify/CreateNewUser.php#L25
https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Console/Commands/SetupAdminUser.php#L66
https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Console/Commands/SetIsAdmin.php#L27

"Sort" serves more than just a cosmetic purpose when it comes to ranks. It's how we detect the correct admin/user rank, regardless of its actual ID. I actually did a PR not that long ago to replace a hardcoded "detect admin rank as ID 1" on both core and multiple extensions, because assuming specific ID number can be very unreliable (under certain circumstances, MariaDB/MySQL can create non-sequential IDs, and it is generally considered best practice to not rely on IDs being sequential).

If you wanted to have the role IDs be more dynamic/manually set, you would need to make it a site setting.... which seems like it could break really, really easily. What if a user changes the site setting, for example, to a moderator rank? Or sets the admin/default rank to 0? Site settings do not have any data validation by design. Doing it based by sort ensures that there is always a valid value for highest/lowest rank.

I just really don't think it's a necessary change, especially given the downsides. The system we have right now for detecting highest/lowest rank works well, so I see no need to replace/potentially break what's perfectly functional.

I'm not opposed to making secondary admin sortable, I just don't think making _all_ admin ranks sortable is the solution.

Since this is a very structural design thing, I'd definitely like @itinerare's weigh-in on it.

SpeedyD:
> https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Actions/Fortify/CreateNewUser.php#L25

This specific one actually, I was thinking of tackling. Why would this require sort or even.. a site setting? It would make more sense to do this through User Ranks. Add a 'default_user_rank' column to the current lowest one, then when updating in User Ranks, set all to false and the changed one to true. Then we could just `Rank::where('default_user_rank', '1')->first()->id();`

> https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Console/Commands/SetupAdminUser.php#L66

This happens during initial setup, and.. only if ranks already exist? A backup, just in case.. Sounds like we could literally just go `Rank::where('is_admin', '1')->first();`, and the sort problem is gone.

> https://github.com/lk-arpg/lorekeeper/blob/6734bce027c3b6e72037bc65383fd53f607f134e/app/Console/Commands/SetIsAdmin.php#L27

And this one is, if I recall correctly, literally only there to change TO the new system, using 'is_admin'. Which means you'd run it once- when you swap from release to develop, and then likely never again. I'd suggest adding a check of `$adminRank = Rank::where('is_admin', '1')->first(); if ($adminRank == null) { `  or something similar, to prevent sort being used if there already is an admin rank.

> "Sort" serves more than just a cosmetic purpose when it comes to ranks.

Which, in my opinion, it really shouldn't.

> It's how we detect the correct admin/user rank, regardless of its actual ID. I actually did a PR not that long ago to replace a hardcoded "detect admin rank as ID 1" on both core and multiple extensions, because assuming specific ID number can be very unreliable (under certain circumstances, MariaDB/MySQL can create non-sequential IDs, and it is generally considered best practice to not rely on IDs being sequential).

Yeah. Hence why I suggest checking for 'is_admin'. There's only ONE rank that has that- and that's the admin account. Any other "admin" accounts have 'is_secondary_admin'.

> If you wanted to have the role IDs be more dynamic/manually set, you would need to make it a site setting.... which seems like it could break really, really easily. What if a user changes the site setting, for example, to a moderator rank? Or sets the admin/default rank to 0? Site settings do not have any data validation by design. Doing it based by sort ensures that there is always a valid value for highest/lowest rank.

..Except I don't want it to be a site setting. I actually do not see any specific reason for it being a site setting. Explained above. Hell, if ONLY because I DO want data validation do I think it's better to NOT make it a site setting.

> I just really don't think it's a necessary change, especially given the downsides. The system we have right now for detecting highest/lowest rank works well, so I see no need to replace/potentially break what's perfectly functional.

Except that 'sorting' isn't exactly intuitive for keeping admin and default user ranks intact. Honestly, the more I think about it, the more I feel like I SHOULD continue and make above suggestions a new PR.


But, that said, I would also like @itinerare to weigh in on this.

itinerare:
I'm also of the mind that the current system works pretty well...? My grievances with using sort have been more along the lines of it occasionally making some queries a bit jankier than they need to be, but I'm pretty sure that mostly applies to the admin rank. That said, I do feel that that rank *shouldn't* be sortable.

That said, I wouldn't mind secondary admin ranks being sortable, and being unable to delete the default user rank is... probably something that should've been implemented a while ago. That plus improving the messaging around how the system works would probably go a ways on its own, if that's something that people are getting confused by...?

... It might also be an idea to disallow assigning powers to the default user rank for safety, but that should probably be a separate PR.

SpeedyD:
...Fine, then, I see I'm outnumbered in my opinion. Then I'll do it in reverse, if the first and last have to remain as they are, then let's keep them there permanently.

Main Admin rank can no longer be sorted.
Default user rank can no longer be sorted or given powers.

Also, redid the whole 'editable' section because _apparently_ secondary admins could edit main admins? Somehow? I didn't fully get the logic so I just scrapped it and started anew.

I did have to finagle the numbering slightly because the default user rank isn't sortable anymore, but it worked out.

Opinions now, @perappu, @itinerare?

PR #1505, opened 2026-06-20
fix: suppress irrelevant/duplicate security advisories

perappu:
Here we go again...

- [PKSA-3r5d-mb8f-1qw9](https://packagist.org/security-advisories/PKSA-3r5d-mb8f-1qw9): Literally just the vulnerability we already patched in #1493 but with different affected versions (still us)
- [PKSA-m5cs-t1y6-qpcs](https://github.com/advisories/GHSA-crmm-hgp2-wgrp): As far as I can tell this only applies to temporary signed urls from the Storage driver. I think we literally only use the Storage driver like a tiny bit in develop, so it's completely irrelevant to us. If we make a shift to the Storage driver fully in the future then it might be a problem? But at that point we'll probably be at a patched version of Laravel.

Also, just to note, the composer.lock will need to be updated for the merge into v3 and above.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,9 @@
             "advisories": {
                 "ignore-id": {
                     "PKSA-mdq4-51ck-6kdq": "Hotfixed using custom sanitization middleware.",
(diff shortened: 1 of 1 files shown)


itinerare (approved):
Holding for merge on Monday as per usual.

PR #1512, opened 2026-07-16
feat(prompts): change character prompt limits to be based on attached characters

perappu:
Currently, prompt limits that are set "per character" are basically just "have you submitted this prompt more times than the number of characters you own". This change actually checks how many times a prompt has been submitted with the specific attached character(s).

Also some minor formatting fixes/a small bug with setting a prompt to be character based at all.

(In case you're wondering what's up with the if statement in `admin/submissions/submission.blade.php`, it's because if you don't shove the add_help directly next to the word "Submission" then it adds a linebreak and looks bad, lol. Same with some of the tweaks in `_prompt.blade.php`.)

### DIFF
--- a/app/Http/Controllers/Admin/SubmissionController.php
+++ b/app/Http/Controllers/Admin/SubmissionController.php
@@ -67,10 +67,11 @@ class SubmissionController extends Controller {
         $prompt = $submission->prompt;
 
         if ($prompt->limit_character) {
-            $limit = $prompt->limit * Character::visible()->where('is_myo_slot', 0)->where('user_id', $submission->user_id)->count();
+            $count = $prompt->getCount($submission->user, $submission->characters->pluck('character'));
         } else {
-            $limit = $prompt->limit;
+            $count = $prompt->getCount($submission->user);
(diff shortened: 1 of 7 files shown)


SpeedyD:
I can't believe I'm going to be saying this, I'm really hating this..

But could you add a way to toggle it to the previous version? I happen to know of a site that actually _LIKES_ the current  method.

..no matter how inane this method actually is.

itinerare:
Hmm, to my way of thinking, the existing logic is almost... a bug? So I think it makes more sense to go ahead with this change and just ask that, if you really like the existing logic for whatever reason, it's your prerogative to override that decision at the site level.

SpeedyD:
Yeah, that's fair enough.. and I honestly agree.

I think just having brought it up means I've done my part. I'll inform the site that this update is incoming.. and then I'm pulling my hands from this.. Aside from one last thing.

SpeedyD (approved):
..cause as I said, I agree with the change.

MarskyMessier (approved):

Issue #1491, opened 2026-05-27
Character trade being auto approved under certain conditions


MarskyMessier:
The current system with trades has an issue that can happen in the following conditions : 
- `open_transfer_queue` setting is set to 1 (true)
- `open_trades_queue` setting is set to 0 (false)
This normally means that trades are auto approved, but that character transfers need staff approval. However, if the settings are set like above, character transfer will be sent to the staff queue, but any trades containing a character will slip thru and be auto approved. I believe this is because of how the logic function in https://github.com/lk-arpg/lorekeeper/blob/270005c974062db90ef6caaecf747421c4593fa8/app/Services/TradeManager.php#L398 which checks for either of the setting to be false.

MarskyMessier:
Working on this, @itinerare do you remember what the `isset($trade->data['recipient']['characters']))` is for at the end? As I don't think trade recipient should be able to bypass the moderator approval either, however I assume it's there for a reason.

itinerare:
I wasn't the last person to meddle with that, I don't think, so I'd probably have to crack open the full thing to make a meaningful assessment...

MarskyMessier:
Dang, I'll do more testing then and see if it breaks anything removing it.

ScuffedNewt:
FYI you can see who last touched a line using blame
https://github.com/lk-arpg/lorekeeper/blame/270005c974062db90ef6caaecf747421c4593fa8/app/Services/TradeManager.php#L398

SpeedyD:
> FYI you can see who last touched a line using blame https://github.com/lk-arpg/lorekeeper/blame/270005c974062db90ef6caaecf747421c4593fa8/app/Services/TradeManager.php#L398

Not to be that guy, @ScuffedNewt, but uhh.. blame says it's _you_. Can YOU explain it? :o

ScuffedNewt:
Yes, commented on the PR which is why I mentioned here about the blame! @SpeedyD

SpeedyD:
..Did not think about that, sorry. ^^;

PR #1513, opened 2026-07-27
fix: trade logic now catches for characters

MarskyMessier:
In reference to #1491 , this should fix the issue of trade filter not being enough.
Now it should:
- auto-complete trade if trade setting is 0 and there's no character in either sender or recipient slot
- auto-complete trade if transfer setting is 0 and there's a character in the trade

It does leave the scenario of trades moderated but transfer auto approved wonky, but I feel allowing this (where filtering thru trade themselves might be trickier) is better than the reverse, especially as trades don't only contain characters but all allowable items.

I've also modified the admin panel to display trades depending on its setting or transfer, meaning even if trades are auto-modded, in the case of transfer being manual it won't hide from Admins.

### DIFF
--- a/app/Services/TradeManager.php
+++ b/app/Services/TradeManager.php
@@ -395,7 +395,7 @@ class TradeManager extends Service {
             }
 
             if ($trade->is_sender_trade_confirmed && $trade->is_recipient_trade_confirmed) {
-                if ((!Settings::get('open_trades_queue') || !(Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
+                if ( (!Settings::get('open_trades_queue')) && (!isset($trade->data['sender']['characters']) && !isset($trade->data['recipient']['characters'])) || (!Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters'])) )  {
                     // Distribute the trade attachments
                     $this->creditAttachments($trade);
 
--- a/app/Http/Controllers/Admin/HomeController.php
+++ b/app/Http/Controllers/Admin/HomeController.php
@@ -25,7 +25,7 @@ class HomeController extends Controller {
      */
     public function getIndex() {
         $openTransfersQueue = Settings::get('open_transfers_queue');
-        $openTradesQueue = Settings::get('open_trades_queue');
+        $openTradesQueue = Settings::get('open_trades_queue') || $openTransfersQueue;
         $galleryRequireApproval = Settings::get('gallery_submissions_require_approval');
         $galleryCurrencyAwards = Settings::get('gallery_submissions_reward_currency');
 


MarskyMessier:
Noting I might've made a mistake with syntax, currently working on fixing it.

app/Services/TradeManager.php
```diff
@@ -395,7 +395,7 @@ public function confirmTrade($data, $user) {
             }
 
             if ($trade->is_sender_trade_confirmed && $trade->is_recipient_trade_confirmed) {
-                if ((!Settings::get('open_trades_queue') || !(Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters'])))) {
+                if ( (!Settings::get('open_trades_queue')) && (!isset($trade->data['sender']['characters']) && !isset($trade->data['recipient']['characters'])) || (!Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters'])) )  {
```
ScuffedNewt, line 398: It's been a bit since I touched this logic but having some doubts about this change

The proposal has two logic branches
```php
(!Settings::get('open_trades_queue')) &&
(!isset($trade->data['sender']['characters']) && !isset($trade->data['recipient']['characters']))
```

and

```php
(!Settings::get('open_transfers_queue')) && (isset($trade->data['sender']['characters'])
```

The second branch doesn't check if the recipient is sending characters, it's important to note that trades are two-way and that "sender" is just the person who created the trade
edit: this means people could work around approvals by just getting the other person to create the trade

what I think is intended with this PR:
```php
$hasCharacters =
    isset($trade->data['sender']['characters'])
    || isset($trade->data['recipient']['characters']);

if ((!Settings::get('open_trades_queue') && !$hasCharacters) ||
    (!Settings::get('open_transfers_queue') && $hasCharacters)
) {
    //
}
```

This puts the trade in the approval queue when either:
- Trades require approval & no characters attached
- Trades don't require approval, but the transfers queue is closed and either party added characters

HOWEVER, considering the trades setting specifically states `"0: Trades do not need mod approval, 1: Trades must be approved by a mod."` I would argue that this should override the transfer queue being open - since trades can involve characters  + items, etc

setup for that:
```php
$hasCharacters = isset($trade->data['sender']['characters']) || isset($trade->data['recipient']['characters']);
$requiresApproval = Settings::get('open_trades_queue') || ( $hasCharacters && Settings::get('open_transfers_queue') );

if (!$requiresApproval) { // } ...
```

Feel free to correct me if I misinterpreted the intention or changes, long day at the meat factory

MarskyMessier: The first branch targets trades specifically, the second is for transfer. Transfer cannot have a recipient sending a character, it's a one-way trade, hence why there's no need to check for a recipient sending a character (they cannot, with how transfers work by defacto). I did verify and made sure any trades with character will be checked, this is what the first branch does: it checks for the trade setting and also if there's any characters in either recipient or sender, both conditions + trade check have to return false to have it auto approved basically (meaning trades aren't moderated and there's no character or blank slots involved). 

Btw for the trade/transfer it also overrides it, given it checks for both trade & content of the trade, so if either is false in the first branch it gets grabbed. This means if transfer queue is opened, but trade is not, no matter if there's characters in a trade, it *will* be checked because trades are moderated. Hopefully this makes sense, it's... something to explain and was something to sort up :'D

I was also honestly thinking about perhaps refactoring how things are handled or named, especially as "open transfer" and "open trades" aren't super descriptive on their own, and might be a bit confusing because item transfer & trades  =/= character tranfers - however that might be for another PR given this is out of the scope here.

ScuffedNewt: this is the `TradeManager` though, not the character manager? so transfers don't use this function

MarskyMessier: Whoops right - I do think I got mixed up because of the transfer settings being present and it has been a while since I last really dwelved back in LK <<, either way the first branch should be correct. I think your proposition would be better tho, it's easier to read too in case someone needs to read that back at any point. Feel free to request a change for it !

Issue #1510, opened 2026-07-10
Memory size error on gallery index with many gallery views


MarskyMessier:
Had an issue on a site with a very large gallery index where it'd occasionally throw a memory alloc error despite being up to date with the gallery files. I was not able to reproduce the error reliably on a local instance.
This should in theory be resolved by storing image width ahead of time during submission instead of doing that during the thumbnail query used in the index ?

The error log:
`production.ERROR: Allowed memory size of 134217728 bytes exhausted (tried to allocate 135168 bytes) {"userId":1463,"exception":"[object] (Symfony\\Component\\ErrorHandler\\Error\\FatalError(code: 0): Allowed memory size of 134217728 bytes exhausted (tried to allocate 135168 bytes) at /var/www/SITENAME/www/vendor/laravel/framework/src/Illuminate/Database/Eloquent/Builder.php:411)`

liwoyadan:
The thing that sticks out to me here is: 
- You're at PHP's default out-of-the-box memory limit of 128mb (134217728 bytes)
- The allocation that caused this error was only a bit over ~130*kb* (135168 bytes), implying the rest of the memory was being eaten up elsewhere
- And your stacktrace doesn't point to an image handler - it's on `Eloquent\Builder`, which implies the memory being chewed up to how much of the database is being fetched at the given time than computing image widths

A quick bandaid for this is bumping your PHP's `memory_limit` up to 256M instead (if you're on a host that'll allow it) because 128M is low for a site of any substantial size, but this doesn't inherently address the root cause of what's actually making the memory get chewed so much on the gallery. (FWIW local PHP `memory_limit` is often different than production's value because you're on a local environment, which is likely why you haven't been able to recreate the issue there + on a large prod site, you have multiple other users hitting routes and querying the DB simultaneously.)

I have a speculation on what's actually making this occur on galleries specifically but haven't had many opportunities to test it 🤔 whenever I crawl onto a PC today I'll see about digging up some code chunks and sending them over to you to see if it manages to remedy the live site you have this issue on too. 🫡

itinerare:
Yeah, storing thumbnail image width rather than wrangling it at time of page load was my guess as to the fix/culprit, but I also haven't had a decent opportunity to dig into it....

PR #1514, opened 2026-08-01
fix(galleries): speed up gallery load time

liwoyadan:
Resolves issue #1510

First pass reduced 63k+ model load to ~200 models loaded on gallery index. `$with` and `$withCount` on models hydrates the eager load on *every single model retrieved* which is why large galleries had issues with load times. And then subsequent investigation and adjustments followed: it turned out the `->through()` query was building a heavy join in the database, which could be bypassed by simply making a single query builder that fetched child submissions by plucking children gallery IDs. Eager loads were kept where actually needed, and additional scopes were added - one scope consolidates them all while they all can be called individually for convenience if anyone wants to change it up.

Also cleaned up some redundancy: `->visible()` already checks for acceptance. `submissions()` relation already sorts in descending order. Made the sidebar queries match what they're meant to display, and replaced `->count()` with `->exists()` for efficiency.

At first I refactored the `Image::make()` into a native PHP function, which benchmarked much faster, but it was pointed out how the max is 200px anyways so I removed the whole `@php`  and let flex do its thing.

Some numbers from adding this specific eager load before the display data scope, after all the other stuff:
`->with('collaborators', 'participants')`
> **GALLERY INDEX**
> Before the eager load: 263 queries, 211 models, 1.37s
> After the eager load: 196 queries, 207 models, 976ms
> 
> **GALLERY PAGE**
> Before: 140 queries, 126 models, 945ms
> After: 88 queries, 115 models, 815ms
> 
> **CHILD SUBMISSION ONLY GALLERIES**
> Before: 161 queries, 141 models, 1.36s
> After: 121 queries, 139 models, 980ms
> 
> **ALL RECENT**
> Before: 132 queries, 111 models, 1.21s
> After: 92 queries, 111 models, 813ms

The one aspect to know for the future: even at ~200 models, a gallery index still had a 9 second load time. Upon further investigation, it was *only* the index and the individual gallery page where only child submissions were shown. Lesson learned is that a giant join query itself makes site load chug, which was subsequently resolved via a simpler query to gallery submissions directly in-model. (And it doesn't help that MySQL/MariaDB provisions 128M by default, but access to touching that varies by host provider and is only a bandaid - doesn't target the root issue.)

Oh also children submission-only galleries did not paginate. I added pagination for it, but wasn't entirely sure if the no-pagination was intentional, so it's free to just strip out.

PRing to v3 because I think enough sites are on v3/will be moving to v3 that have hefty enough galleries to benefit from this. I think this is as far as it can be refactored without changing DB schema itself, and from what I can tell the tables already have appropriate indices...

Thank you to Mark for helping test this and report back data to me so I could finish pinpointing and finetuning what exactly was causing the pages to load so slow. :)

### DIFF
--- a/resources/views/galleries/gallery.blade.php
+++ b/resources/views/galleries/gallery.blade.php
@@ -70,11 +70,15 @@
 
         {!! $submissions->render() !!}
     @elseif($childSubmissions->count())
+        {!! $childSubmissions->render() !!}
+
         <div class="d-flex align-content-around flex-wrap mb-2">
-            @foreach ($childSubmissions->orderBy('created_at', 'DESC')->take(20)->get() as $submission)
+            @foreach ($childSubmissions as $submission)
                 @include('galleries._thumb', ['submission' => $submission, 'gallery' => false])
             @endforeach
         </div>
+
+        {!! $childSubmissions->render() !!}
     @else
         <p>No submissions found!</p>
     @endif
--- a/resources/views/galleries/gallery.blade.php
+++ b/resources/views/galleries/gallery.blade.php
@@ -71,7 +71,7 @@
         {!! $submissions->render() !!}
     @elseif($childSubmissions->count())
         <div class="d-flex align-content-around flex-wrap mb-2">
-            @foreach ($childSubmissions->orderBy('created_at', 'DESC')->get()->take(20) as $submission)
+            @foreach ($childSubmissions->orderBy('created_at', 'DESC')->take(20)->get() as $submission)
                 @include('galleries._thumb', ['submission' => $submission, 'gallery' => false])
             @endforeach
         </div>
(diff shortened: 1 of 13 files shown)


itinerare:
Thank you for digging into this, especially so thoroughly!

resources/views/galleries/gallery.blade.php
```diff
@@ -70,11 +70,15 @@
 
         {!! $submissions->render() !!}
     @elseif($childSubmissions->count())
+        {!! $childSubmissions->render() !!}
```
itinerare, line 73: If I recall correctly, it was semi-intentional that these didn't paginate, but I don't have strong feelings about it either way.

itinerare:
> At first I refactored the Image::make() into a native PHP function, which benchmarked much faster, but it was pointed out how the max is 200px anyways so I removed the whole `@php` and let flex do its thing.

Yeah... yeah this is the actual smarter solution. |D

itinerare (approved):

SpeedyD (approved):

PR #1518, opened 2026-08-14
fix: unable to remove images on features

perappu:
Fixes missing secondary argument that resulted in images not actually getting removed from features when the box is checked.


SpeedyD (approved):
How did we miss this

liwoyadan (approved):

PR #1519, opened 2026-08-15
fix(traits): clarify the deletion modal

liwoyadan:
I've had enough clients baffled and ask me why they can't delete a trait when they've removed it from every single existing character on the site. I realized: the deletion modal isn't clear about this. It counts traits in design updates too. So this is just a very small wording change to mention that design updates are also counted.


itinerare (approved):

AW0005 (approved):

perappu (approved):
please dear god yes I've dealt with this confusion like three times in the last two weeks

PR #1521, opened 2026-08-17
fix: quick raffle is_active validation + winner optimization

strabii:
really simple fixes on v3:

- added validation to adding raffle tickets to check if a raffle is active. I noticed mostly everywhere that lets you add raffle tickets as rewards will never display inactive raffles, so I don't think it makes sense if users can still join an inactive raffle. It's absolutely possible if say, somehow, a raffle was made active and the ticket was added as a prompt reward, and then the raffle was set inactive or something of the sort- an user could still enter the inactive raffle.
- I think I optimized the code for removing a winner from the raffle pool when in the rolling loop.... it uses the `whereNotIn` query using the value stored in $winner before the next loop starts it over again. nulls in $winner->alias or $winner->user_id do not impact the function. please please please the 3000 ticket foreach statement was so funny to me I keep losing sleep over it

### DIFF
--- a/app/Services/RaffleManager.php
+++ b/app/Services/RaffleManager.php
@@ -27,10 +27,6 @@ class RaffleManager extends Service {
      * @return int
      */
     public function addTickets($raffle, $data) {
-        if (!$raffle->is_active) {
-            throw new \Exception('Raffle is not currently active!');
-        }
-
         $count = 0;
         foreach ($data['user_id'] as $key=> $id) {
             if ($user = User::where('id', $id)->first()) {
@@ -57,13 +53,11 @@ class RaffleManager extends Service {
      * @return int
      */
     public function addTicket($user, $raffle, $count = 1) {
-        if (!$raffle->is_active) {
-            throw new \Exception('Raffle is not currently active!');
-        }
-
         if (!$user) {
             return 0;
         } elseif (!$raffle) {
+            return 0;
+        } elseif (!$raffle->is_active) {
             return 0;
         } elseif ($count == 0) {
             return 0;
--- a/app/Services/RaffleManager.php
+++ b/app/Services/RaffleManager.php
@@ -27,6 +27,10 @@ class RaffleManager extends Service {
      * @return int
      */
     public function addTickets($raffle, $data) {
+        if (!$raffle->is_active) {
+            throw new \Exception('Raffle is not currently active!');
+        }
+
         $count = 0;
         foreach ($data['user_id'] as $key=> $id) {
             if ($user = User::where('id', $id)->first()) {
@@ -53,6 +57,10 @@ class RaffleManager extends Service {
      * @return int
      */
     public function addTicket($user, $raffle, $count = 1) {
+        if (!$raffle->is_active) {
+            throw new \Exception('Raffle is not currently active!');
+        }
+
         if (!$user) {
             return 0;
         } elseif (!$raffle) {
@@ -197,13 +205,8 @@ class RaffleManager extends Service {
 
             $ticketCount--;
 
-            // remove tickets for the same user...I'm unsure how this is going to hold up with 3000 tickets,
-            foreach ($ticketPool as $key=> $ticket) {
-                if (($ticket->user_id != null && $ticket->user_id == $winner->user_id) || ($ticket->user_id == null && $ticket->alias == $winner->alias)) {
-                    $ticketPool->forget($key);
-                }
-            }
-            $ticketPool = $ticketPool->values();
+            // remove tickets for the same user
+            $ticketPool = $ticketPool->whereNotIn('user_id', $winner->user_id)->whereNotIn('alias', $winner->alias)->values();
             $ticketCount = $ticketPool->count();
         }
 


ScuffedNewt (approved):

SpeedyD (approved):

app/Services/RaffleManager.php
```diff
@@ -53,6 +57,10 @@ public function addTickets($raffle, $data) {
      * @return int
      */
     public function addTicket($user, $raffle, $count = 1) {
+        if (!$raffle->is_active) {
+            throw new \Exception('Raffle is not currently active!');
```
itinerare: Rather than use the usual if -> throw exception setup, this should use the method of the checks below. It should also be after the check for whether the raffle exists (`elseif (!$raffle) {...`), or it's liable to cause an error if there's not a raffle to check the status of.

app/Services/RaffleManager.php
```diff
@@ -27,6 +27,10 @@ class RaffleManager extends Service {
      * @return int
      */
     public function addTickets($raffle, $data) {
+        if (!$raffle->is_active) {
+            throw new \Exception('Raffle is not currently active!');
```
itinerare: This probably just doesn't need to be there as long as `addTicket()` has proper validation; *this* function doesn't do anything except delegate to that one.

app/Services/RaffleManager.php
```diff
@@ -197,13 +205,8 @@ private function rollWinners($raffle) {
...
 
-            // remove tickets for the same user...I'm unsure how this is going to hold up with 3000 tickets,
-            foreach ($ticketPool as $key=> $ticket) {
-                if (($ticket->user_id != null && $ticket->user_id == $winner->user_id) || ($ticket->user_id == null && $ticket->alias == $winner->alias)) {
-                    $ticketPool->forget($key);
-                }
-            }
-            $ticketPool = $ticketPool->values();
+            // remove tickets for the same user
+            $ticketPool = $ticketPool->whereNotIn('user_id', $winner->user_id)->whereNotIn('alias', $winner->alias)->values();
```
itinerare, line 203: Unsure of Laravel's behavior here if only one of user ID/alias are set (though I like the idea of this optimization in theory)....

itinerare (changes requested):

app/Services/RaffleManager.php
strabii: oh yes this is a good thing to test, and good news! I didn't really see any abnormal behaviors with rolling :D

If assigning a raffle ticket to an on-site user, the alias won't process so the alias will always be null afaik. And vice versa an alias will never have an on-site user ID. So only one user/alias only exists per raffle ticket at a time it looks like.

I did some test rolls with a single user instance still:
A raffle with a single on-site user: Rolled fine
A raffle with a single off-site alias: Rolled fine
A raffle with 6 tickets from a single on-site user: Also rolled fine (rolling for 2 winners still only rolled 1 person since theres only 1 eligible person)
A raffle group with 2 raffles rolling 1 winner each and both raffles only had the same user: Rolled user on raffleA, removed user on raffleB (bc not eligible anymore) so raffleB was empty lol

app/Services/RaffleManager.php
strabii: thumbs up! I was thinking the data processing just shouldn't even start to hit the foreach loop entirely tbh 🤔

app/Services/RaffleManager.php
itinerare: Yeah, there should only ever be one of user ID or alias set for a ticket; as long as a `null` input causes that part of the query to be effectively inert, all should be well.

app/Services/RaffleManager.php
itinerare: Makes sense in theory (and in the context of other areas of the project), this area's just kinda odd. Best to keep to its existing conventions until some theoretical future point (out of the scope of this PR; it'd need to land on unstable at minimum) where it gets refreshed and brought more in line with the rest of the project.

strabii: I've removed this part here so the validation only happens inside AddTicket() !

app/Services/RaffleManager.php
strabii: finally github is back online ToT edited this to match the rest of the rafflechecks so it returns 0 like the rest instead of throw error

app/Services/RaffleManager.php
strabii: yep nulls haven't caused any errors for me so should be well indeed!

itinerare: Did a little testing of my own and tinkering with other possible methods, and curiously enough this is a quirk of `whereNotIn()`; with `where(..., '!=',...`, there is undesired behavior, but not so with `whereNotIn()`. So this looks good, as far as I can tell!

itinerare (approved):

app/Services/RaffleManager.php
strabii: hooray!! no more 3000 ticket foreach loops! XD

SpeedyD (approved):

PR #1526, opened 2026-08-25
feat: add form adapter

itinerare:
I had the realization that I could simply cut the Gordian knot that was the prospect of maintaining the adapter package (`alban/laravel-collective-spatie-html-parser`) with minor tweaks... by simply incorporating the adapter, which is largely contained in a single file, into the project itself. This allows the removal of the `laravelcollective/html` package without any actual disruption... at least in theory.

As far as I can tell, the only meaningful change required from the original package was to the default behavior of labels; most of the labels in the project (unfortunately) don't actually note their corresponding form field/only provide one input, and the adapter does *not* assume it to be the label text, so out of the box labels simply vanished. Easy enough to fix, though.

...That being said, while I gave this some cursory testing, I would *really* like for folks to give it more practical testing.


perappu (approved):

ScuffedNewt (approved):

app/Providers/FormsProvider.php
```diff
@@ -0,0 +1,26 @@
...
+
+use App\Helpers\FormAdapter;
+use Illuminate\Support\ServiceProvider;
+
+class FormsProvider extends ServiceProvider {
+    /**
+     * Register bindings in the container.
+     */
+    public function register() {
+        $this->registerFormAdpater();
```
SpeedyD: ..Adpater?

app/Providers/FormsProvider.php
```diff
@@ -0,0 +1,26 @@
...
+
+        $this->app->alias('form', FormAdapter::class);
+    }
+
+    /**
+     * Register the HTML adapter instance.
+     */
+    protected function registerFormAdpater() {
```
SpeedyD: ...oh wow, nvm the function is also misspelled, so it'll work just fine

SpeedyD (approved):
Found a misspelling, but uh. I guess it functions?

PR #1533, opened 2026-08-27
feat(items): auto select stack in modal

liwoyadan:
So we can collectively stop having users confused on why nothing is happening because they don't realize an item stack needs to be selected, hahaha.

As per moif's recommendation, this is an *opt-in* user setting - so users are aware they have chosen to enable this for their own/their characters' inventories and won't accidentally lose something. `php artisan migrate` as usual.

When opted in, there is a banner - when no stack is selected - saying that because nothing is selected, the first stack will automatically be selected and used. The banner goes away if a stack is intentionally selected. It still hits whatever checks in the back for the action selected (i.e. if the first stack is accountbound and the user tries to transfer, it'll still throw an exception).

### DIFF
--- a/resources/views/character/_inventory_stack.blade.php
+++ b/resources/views/character/_inventory_stack.blade.php
@@ -145,9 +145,10 @@
         var $rowId = "#itemRow" + $checkbox.value
         $($rowId).find('.quantity-select').prop('name', $checkbox.checked ? 'quantities[]' : '')
     }
-    
+
     @if (!$readOnly && $user && $user->settings->stack_auto_selected)
         var $autoSelectNotice = $('#autoSelectNotice');
+
         function updateAutoSelectNotice() {
             $autoSelectNotice.toggle($('.item-check:checked').length == 0);
         }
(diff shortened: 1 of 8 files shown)


resources/views/character/_inventory_stack.blade.php
```diff
@@ -140,4 +145,26 @@ function updateQuantities($checkbox) {
         var $rowId = "#itemRow" + $checkbox.value
         $($rowId).find('.quantity-select').prop('name', $checkbox.checked ? 'quantities[]' : '')
     }
+
+    @if (!$readOnly && $user && $user->settings->stack_auto_selected)
```
ScuffedNewt, line 149: could we make this a reusable widget for use in extensions?

itinerare: Given the fussiness of PR editing, would you be amenable to plucking out the JS in a subsequent PR?

ScuffedNewt: Yes!

ScuffedNewt (approved):

itinerare (approved):

PR #1552, opened 2026-09-11
fix: site images not scaling

perappu:
at some point site images began to look like this:
image

this makes them Not Do That (in the most lorekeeperest way I could think of)


itinerare (approved):
Yeah that's sufficiently lorekeepery

MarskyMessier (approved):
big image begone

PR #1557, opened 2026-09-17
fix(prompts): clarify submission count limits on prompt submission forms

perappu:
a bunch of stuff to do with inability to save a prompt/submission or inaccurate displaying of submission count info

- Fixes drafts counting against limits
- Migration to fix limit_period being an enum (which was preventing use of the time periods defined in the config)
- Relabels the "default prompt rewards" section to "prompt details" and displays the name of the selected prompt, separates out default rewards from submission count limits (same blade just two different headers)
- Moves said "prompt details" section out of the rewards card
- A bunch of ajax stuff that will show the submission count for a specific character when one is selected (I don't think it was listed anywhere previously?? so the user kinda just had to like. magically know.  might be good to display it on prompt entry as well but I don't have it in me to figure that out right now)

Preview of how it looks:
image
image

tbh I do not know if this is a fix or a feat or if this should truly be all one commit but it felt like everything was reliant on everything else, so


ScuffedNewt (approved):

liwoyadan (approved):

itinerare:
I think technically it's maybe a feat but in spirit it's a fix, it's all good. Sometimes you gotta just improve something to fix it

PR #1561, opened 2026-09-20
remove redundant alt text

LostInProgres:
Been thinking about this a lot lately, and ultimately came to the conclusion that I might as well just- cough this up for others to judge, lol. Admittedly not 100% sure about this but. Hmm. It's here now!

These are all alt texts I believe should be blank due to being redundant. Afaik these images are decorative, in part because the name is already the very next thing that (would be) read. i.e. if you have an item with the name text "potion" and the alt text "potion" the screen reader will just read out "potion potion" which isn't very helpful.

I do want to eventually add proper alt text on items, but even with proper alt text this should probably only be displayed on the encyclopaedia, as in all these cases the user doesn’t really miss out on any context with the image not being there.

IDK. What else to put? Uhhh have cat tax! You deserve it!
image (7)


ScuffedNewt:
dont really have any opinion on removal or not, but do we know if empty alt text has any issues with it or?

perappu:
I hate to say this, but I'm staunchly against this. Screenreaders are very reliant on alt text. Many people who use them would prefer to have "redundant" alt text than simple have their screenreader ambiguously read out basically "there is an image here". From my understanding, it would not just read out "potion potion" -- it would read out "potion", and then that the next thing is an image with the description "potion". (though it depends on the screenreader)

You either want to have alt text, or in the case of something purely decorative, a tag that tells a screenreader to ignore it. I would also argue that these images do not serve as purely decoration -- they are important illustrations depicting the the text around them. Removing alt text is a step backwards towards accessibility, and I'd actually argue we need more of it.

Perhaps a better solution would be something like, updating the alt text to read out "an image of [item name]" instead of removing it? Or, even better, allowing users to define the alt text on the create/edit item page. I know you mentioned doing something like that, but I think making it empty in the meantime is a subpar compromise.

That said, I will of course let @itinerare and possibly @liwoyadan (if you have the mental energy, since you're more knowledgeable on this topic than me) chime in before a decision is made either way.

> dont really have any opinion on removal or not, but do we know if empty alt text has any issues with it or?

also to answer this, from my understanding, blank alt text usually just causes it to skip the image entirely

LostInProgres:
> I hate to say this, but I'm staunchly against this. Screenreaders are very reliant on alt text. Many people who use them would prefer to have "redundant" alt text than simple have their screenreader ambiguously read out basically "there is an image here". From my understanding, it would not just read out "potion potion" -- it would read out "potion", and then that the next thing is an image with the description "potion". (though it depends on the screenreader)

From my testing with multiple screen readers, "potion potion" is exactly what gets read out a lot of the time. It will occasionally read out "Potion graphic, potion x1" or similar instead. Regardless though, in all the instances I am suggesting changing it to a blank alt, the item name is read out directly after the image.

> You either want to have alt text, or in the case of something purely decorative, a tag that tells a screenreader to ignore it. I would also argue that these images do not serve as purely decoration -- they are important illustrations depicting the the text around them. Removing alt text is a step backwards towards accessibility, and I'd actually argue we need more of it.
>
> Perhaps a better solution would be something like, updating the alt text to read out "an image of [item name]" instead of removing it? Or, even better, allowing users to define the alt text on the create/edit item page. I know you mentioned doing something like that, but I think making it empty in the meantime is a subpar compromise.

While I see the point being made, from my research best practice for alt text specifically recommends to avoid redundant information. In this case, refer to example 2 (and in some cases 3) on https://webaim.org/techniques/alttext/ -- the context for the image that would require alt text is already inline with the text right below/besides the image, duplicating it provides unnecessary information that provides no new information when using a screen reader.

Alt text containing "an image of" are recommended against. I also feel like on many inventory-esque pages like these, it is important to keep the page's function in mind. While custom alt text is nice and _definitely_ something I plan to add (in the process of it- trying to figure out the logistics) I do think it's worth it to keep in mind that inventories ultimately might not be the place for detailed descriptions like these as it would make the page incredibly verbose. Regardless, that feels like a bridge to cross when custom alt text _is_ added.

perappu:
Those are some fair enough points! (forgot that "image of a potion" is a terrible suggestion) I'm glad to hear you actually tested this with some screen readers.

I don't know if these qualify as purely decorative, though. And I'm not sure I want to do this _before_ we add better support for defining alt text... Let me run it by my blind friend!

Anarchie:
i hope it's okay to comment, i'm just subscribed so i can get more used to laravel and use the unstable, y'know? sorry to intrude haha

but i use a screenreader sometimes due to being low vision with severe migraine auras + dayblindness from photophobia. i don't always use it but what i mean is, do you need extra input?

for what it's worth, using the code to ignore the image if there's descriptive text right next to it is usually fine, in the case of items and the like.

ScuffedNewt:
you're not intruding, this is an open forum! everyone's input is appreciated and wanted - in fact we always are looking for more people to contribute.
If you have any strong opinion or recommendation it is always good to hear!

perappu:
I have spoken with said friend! ([here's a link to their organization](https://steatornis.no/en)) They also ran it by another blind person for a second opinion and a developer they know who focuses on accessibility.

Here were their thoughts:
- Have screen readers skip the item select table icons (but they noted that having the icons there in general is good for other kinds of accessibility)
- The inventory page is a big gray area that could go either way. The images are not straightforwardly decorative as they provide context and flavor for the game.
- However, alt text on the inventory can be removed if we **keep** the alt text in the inventory stack modal, and that's where it would be appropriate for more lengthy, descriptive alt text to live.

itinerare:
That makes sense to me... a lot of pages are very "function first" and have images for *visual* readability/decoration, but extra detail there is liable to bog down the page/make navigation more difficult than it needs to be, esp in games with many items (which, I think any approach to this needs to keep in mind that sometimes games have pet site quantities of items/stuff in folks' inventories). That issue doesn't exist for pages/views dedicated to just one item/similar objects, though.

Anarchie:
oh, i'm really glad you got ahold of your friends, that's worded really well! C: huge thanks to them.

i was taking a moment and trying to think of a way to put that singular image pages like the inventory or other encyclopedic endpoints really are not a bother at all and are useful to have more context on, while multiple images in a row can be a lot of noise, and if described in text already, then absolutely, it's the kind of thing that's being mentioned.

confirming that on the sites i was checking i was getting "[item name], Image, Link, [item name], Link", and so on. the speed i have mine at is slow enough for that to not be overwhelming because of me fussing with settings to help with a focus on sensory sensitivity, but others may be getting that faster and more overwhelmingly, yeah. :o

LostInProgres:
putting this in draft rq because @ch3rvb was kind enough to write me the code to make custom alt text~

Wanted to have people look at this before going trough and adding it to other areas, to save myself some trouble in case edits are needed. Goal is to just essentially put it on all the create/edit pages that allow you to upload an image.

the field to add alt text is in a widget, and the data is stored in a single table. The module also has a key so that it can be used on things that have more than one image (like the Icon and World images on currencies)

That said; just wanted to confirm- it seems like we're agreeing to leave the alt text as alt="" on pages with (potential of) a lot of repetition, mostly inventories, shops, trade listings. and the like. Then leave the full alt text for pages where the item is more "isolated", like the encyclopaedia and the modals in inventory stacks/shop stock.

routes/lorekeeper/admin.php
```diff
@@ -507,3 +507,8 @@
 Route::group(['prefix' => 'rewards', 'middleware' => 'power:manage_data'], function () {
     Route::post('/', 'RewardController@postPopulateRewards');
 });
+
+// ALT TEXT
+Route::group(['prefix' => 'alt-text', 'middleware' => 'power:manage_data'], function () {
```
ScuffedNewt: i dont like that this is its own controller... the whole implementation feels a bit smelly to me... can we reference how rewards do it on prompt model?

i dont have any strong thoughts on how to go about this otherwise but i feel like we are repeating this widget pattern overzealously

perappu:
To be honest I'm with Newt, it is a known fact that I love traits and want to shove 'em in everything ever, but this feels kinda overkill.

I was honestly thinking all we needed was a new "alt_text" string on items that goes as a plain text input on the create/edit items page. Super simple, just a single field that can get referenced.

I can see a trait structure like this for a i8n feature (it was one I've speculated on in the past), but I think making a whole trait _just_ for alt text is a lot. It's like the equivalent of making a trait to put a description field on something.

